PR #14, opened 2022-09-23
Add Docker Compose spec for deploying backend

bisgardo:
## Purpose

Enable local deployment of backend with a TimescaleDB instance. Frontend would be a nice addition, but then #16 should be merged first. Besides, backend was all I needed for now.

## Changes

Add Docker Compose spec with backend and TimescaleDB instances.

## Checklist

      hard-to-understand areas.


docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
...
+# The backend may be run as an existing (official) image or built from local source.
+# See the readme for usage.
+
+version: "3"
+services:
+  backend:
+    image: ${CCDSCAN_BACKEND_IMAGE-concordium/ccdscan:test}
+    build: ./backend
+    ports:
+    - "5000:5000"
```
soerenbf, line 11: Maybe we should choose another host port. 5000 is always in use on macOS, (something added in macOS Monterey it seems).

I tried changing the port to 8001, and that made the build work (or at least so it seems, haven't tested by calling any endpoints 🤷 )

I assume this is something that cascades to other places in the application that all reference this port.

bisgardo: Wow thanks Apple. Yes, this port is referenced a lot in code and docs so would prefer not to introduce inconsistencies or cascading changes.

How about I parameterize the port with `CCDSCAN_BACKEND_PORT` that defaults to 5000. The readme will then suggest mac users to set it explicitly like `CCDSCAN_BACKEND_PORT=5001 docker-compose up`?

soerenbf: Sure, let's do that 😀 is the application also available on port 5001?

soerenbf (approved):

docker-compose.yaml
bisgardo: Done. Not sure what you mean by application. This is only the backend and yes it'll be available on port 5000 by default and otherwise the value that you set.

I'll look into adding the frontend later.

soerenbf: > Done. Not sure what you mean by application. This is only the backend and yes it'll be available on the host on port 5000 by default and otherwise the value that you set.

Yes.. was just a brainfart... I meant the backend, but it doesn't make any sense as it runs inside the container 🙃

PR #16, opened 2022-09-26
Add workaround to NITRO_PRESET not taking precedence

bisgardo:
## Purpose

The Nitro preset is defined to `firebase` as that is what we use in production. If you build your own private instance, you probably want to be able to run it in a local node.js instance instead of deploying to Firebase.

The variable `NITRO_PRESET` should override this but didn't until [Nitro v0.5.0](https://github.com/unjs/nitro/releases/tag/v0.5.0) ([commit](https://github.com/unjs/nitro/commit/92d711fe936fda0ff877c23d8a0d73ed4ea4adc4)). It's not easy to figure out which version of Nitro we use in this project, but the variable doesn't work, and judging by timing, it's clearly an earlier version than this.

Currently the workaround used [here](https://github.com/bisgardo/concordium-docker/pull/53/files) is to completely overwrite `nuxt.config.ts` which is not a great solution.

## Changes

Read  `NITRO_PRESET` directly in the config file and let the value take precedence there.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -54,7 +54,10 @@ export default defineNuxtConfig({
 			ENVIRONMENT === 'testnet',
 	},
 	nitro: {
-		preset: process.env.NITRO_PRESET || 'firebase', // workaround until whatever depends on Nitro has been upgraded to include  'https://github.com/unjs/nitro/commit/92d711fe936fda0ff877c23d8a0d73ed4ea4adc4'
+		// Workaround: Until whatever depends on Nitro has been upgraded to include 'https://github.com/unjs/nitro/commit/92d711fe936fda0ff877c23d8a0d73ed4ea4adc4', we do it manually.
+		// Workaround on workaround: Have to use empty string instead of "node-server" as otherwise we get error "Cannot resolve preset: node-server"
+		// (see 'https://forum.cleavr.io/t/cannot-resolve-node-server-preset/686').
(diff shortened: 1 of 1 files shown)


frontend/nuxt.config.ts
```diff
@@ -54,7 +54,10 @@ export default defineNuxtConfig({
 			ENVIRONMENT === 'testnet',
 	},
 	nitro: {
-		preset: 'firebase',
+		// Workaround: Until whatever depends on Nitro has been upgraded to include 'https://github.com/unjs/nitro/commit/92d711fe936fda0ff877c23d8a0d73ed4ea4adc4', we do it manually.
+		// Workaround on workaround: Have to use empty string instead of "node-server" as otherwise we get error "Cannot resolve preset: node-server"
+		// (see 'https://forum.cleavr.io/t/cannot-resolve-node-server-preset/686').
+		preset: process.env.NITRO_PRESET === 'node-server' ? '' : 'firebase',
```
soerenbf, line 60: Can't build, tried with `NITRO_PRESET=node-server yarn build`, but got the error described in the workaround on workaround comment 😁

bisgardo: I must have added the empty string thing after testing because that's what breaks it. The problem is that preset `node-server` was renamed from `server`, and the empty prefix *in the config file* would make it fall back to that value. So just using `NITRO_PRESET=server` is the correct solution for the nuxt/nitro version that we use.

PR #23, opened 2022-10-31
Use localhost for development

bisgardo:
## Purpose

By default we should expect development to happen against a local Node and TimescaleDB. At least, the current values don't point at anything that's up anymore.

## Changes

Make default DB connection string and Node gRPC address point to services on localhost.


soerenbf (approved):

soerenbf (commented):
In the backend README the setup part for timescaleDB states the old password. Should probably also update that

bisgardo:
> In the backend README the setup part for timescaleDB states the old password. Should probably also update that

Done (including another occurrence in a test class).

PR #39, opened 2022-12-17
Nodes Sorting

parv0888:
## Purpose
Sorting of Nodes on various available fields

## Changes
Sorting of Nodes on various available fields

## Checklist

      hard-to-understand areas.

## CLA acceptance

- [x] I accept the above linked CLA.


parv0888:
[Screencast from 17-12-22 07:00:29 AM IST.webm](https://user-images.githubusercontent.com/1246947/208216763-4b5cf309-46fe-404f-9227-5e68c37dae68.webm)

backend/Application/Api/GraphQL/Network/ClientVersionComparer.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+            else if (!String.IsNullOrEmpty(x) && !String.IsNullOrEmpty(y))
+            {
+                var xParts = x.Split(".");
+                var yParts = y.Split(".");
+
+                if (xParts.Length != yParts.Length)
+                {
+                    return 0;
+                }
```
abizjak, line 27: I'm not sure why this is sensible. So version `1.0` will be deemed equal to version `3.1.2`?

backend/Application/Api/GraphQL/Network/ClientVersionComparer.cs
```diff
@@ -0,0 +1,48 @@
...
+
+                for (int i = 0; i < xParts.Length; i++)
+                {
+                    var xPart = Int64.Parse(xParts[i]);
+                    var yPart = Int64.Parse(yParts[i]);
```
abizjak, line 32: I am not sure this is safe. What if put a version as `1.2.3-rc1`?

You don't really have any guarantees about the version, and we don't want the backend to raise an exception just because somebody puts a dummy version in the list.

backend/Application/Api/GraphQL/Network/ClientVersionComparer.cs
abizjak: Why not just do a lexicographic sorting, where a prefix of a string is less than it?

parv0888: Hi @abizjak 
We have fixed the semver parsing and removed the code being used for the same. 
We have assumed that the version will be always semver compatible
https://github.dev/Concordium/concordium-node/blob/59f58d14079b980312740daf2f2233c386ed1d90/concordium-node/src/configuration.rs#L30

parv0888: lexicographic sorting puts 10 before 2

backend/Application/Api/GraphQL/Network/ClientVersionComparer.cs
parv0888: we have fixed semver parsing. It should not raise an exception till the version is any valid Semver.

parv0888:
image

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
```diff
@@ -7,12 +7,33 @@ namespace Application.Api.GraphQL.Network;
...
+                : statuses.OrderByDescending(s => Semver.SemVersion.Parse(s.ClientVersion, Semver.SemVersionStyles.Any, MaxSemverLength), Semver.SemVersion.SortOrderComparer),
+            NodeSortField.ConsensusBakerId => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.ConsensusBakerId) : statuses.OrderByDescending(s => s.ConsensusBakerId),
+            NodeSortField.FinalizedBlockHeight => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.FinalizedBlockHeight) : statuses.OrderByDescending(s => s.FinalizedBlockHeight),
+            NodeSortField.NodeName => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.NodeName) : statuses.OrderByDescending(s => s.NodeName),
+            NodeSortField.PeersCount => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.PeersCount) : statuses.OrderByDescending(s => s.PeersCount),
+            NodeSortField.Uptime => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.Uptime) : statuses.OrderByDescending(s => s.Uptime),
+            _ => throw new NotImplementedException()
```
abizjak, line 31: Can this happen? Why not just do nothing in that case instead of raising an exception?

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
```diff
@@ -7,12 +7,33 @@ namespace Application.Api.GraphQL.Network;
...
+            NodeSortField.BlocksReceivedCount => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.BlocksReceivedCount) : statuses.OrderByDescending(s => s.BlocksReceivedCount),
+            NodeSortField.ClientVersion => sortDirection == NodeSortDirection.ASC
+                ? statuses.OrderBy(s => Semver.SemVersion.Parse(s.ClientVersion, Semver.SemVersionStyles.Any, MaxSemverLength), Semver.SemVersion.SortOrderComparer)
+                : statuses.OrderByDescending(s => Semver.SemVersion.Parse(s.ClientVersion, Semver.SemVersionStyles.Any, MaxSemverLength), Semver.SemVersion.SortOrderComparer),
+            NodeSortField.ConsensusBakerId => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.ConsensusBakerId) : statuses.OrderByDescending(s => s.ConsensusBakerId),
+            NodeSortField.FinalizedBlockHeight => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.FinalizedBlockHeight) : statuses.OrderByDescending(s => s.FinalizedBlockHeight),
+            NodeSortField.NodeName => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.NodeName) : statuses.OrderByDescending(s => s.NodeName),
```
abizjak, line 28: Probably the best user experience is to sort ignoring case. So `foo` and `Foo` would be next to each other.

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
```diff
@@ -7,12 +7,33 @@ namespace Application.Api.GraphQL.Network;
...
     {
-        return nodeSummarySnapshot.NodeStatuses;
+        var statuses = nodeSummarySnapshot.NodeStatuses.AsQueryable();
+        statuses = sortField switch
+        {
+            NodeSortField.AveragePing => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.AveragePing) : statuses.OrderByDescending(s => s.AveragePing),
+            NodeSortField.BlocksReceivedCount => sortDirection == NodeSortDirection.ASC ? statuses.OrderBy(s => s.BlocksReceivedCount) : statuses.OrderByDescending(s => s.BlocksReceivedCount),
+            NodeSortField.ClientVersion => sortDirection == NodeSortDirection.ASC
+                ? statuses.OrderBy(s => Semver.SemVersion.Parse(s.ClientVersion, Semver.SemVersionStyles.Any, MaxSemverLength), Semver.SemVersion.SortOrderComparer)
+                : statuses.OrderByDescending(s => Semver.SemVersion.Parse(s.ClientVersion, Semver.SemVersionStyles.Any, MaxSemverLength), Semver.SemVersion.SortOrderComparer),
```
abizjak, line 25: At present there is no **guarantee** that the collector backend only returns valid SemVer for the client version field.

So in principle this is problematic. What will happen in that case? Will the exception be intercepted and the query will fail?

parv0888: The attached link says that is_compatible_version should be semver

parv0888: Also can you help us with what the format of the version can be? So that we can find an appropriate parser for the same

abizjak: At present there is no guarantee about the format. I would recommend the following logic

To compare versions v1, v2 do the following.
- attempt to parse v1, v2 as SemVer
- if successful compare them as SemVer
- if not then
- if v1 is valid SemVer then v1 < v2
- else both are not valid semVer, compare them as strings.

backend/Application/Api/GraphQL/Network/ClientVersionComparer.cs
abizjak: Lexicographic sorting of parts.

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
parv0888: This can happen when a new field NodeSortField is added without it being handeled here for sorting. 
So if that field is selected by the user (or developer who is adding it). A `NotImplementedException` will be better than doing nothing.

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
parv0888: done

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
parv0888: @abizjak  - we have followed the steps that you have mentioned. and added tests
however we do think that the versions can only be a `SemVer`  since Semver type is being used at multiple places including the Handshake
https://github.com/Concordium/concordium-node/blob/0fc12c7b4392477afaa61d64595c947175f70803/concordium-node/src/network/mod.rs#L84

Is there something incorrect with that reasoning?
Will the nodes be able to talk to each other if a node publishes any other format for its version?

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
abizjak: Ok.

backend/Application/Api/GraphQL/Network/NetworkQuery.cs
abizjak: You reasoning is correct assuming you would get the version from the node. But you do not get it from the node. You get the version from the "collector backend" which currently does not seem to do any validation on the client version.

That should also be fixed, but it is better that we are robust in ccdscan as well.

abizjak (approved):

Issue #7, opened 2022-08-04
CCDSCAN calculates baker removal time wrong


smh1001:
**Bug Description**
When deregistering a baker the time is 0 when 3 weeks have gone by, where the baker is actually removed when the payday event passes (10am CEST on mainnet). After that CCDSCAN starts counting the time up agian until the payday actually passes

**Steps to Reproduce**
Deregister a baker sometime except at 10am CEST

**Expected Result**
The baker is removed on the first payday that passes after 3 weeks.

**Actual Result**
CCDSCAN apparently thinks the baker is removed after exactly 3 weeks.

parv0888:
Raised PR
https://github.com/Concordium/concordium-scan/pull/43

mh-concordium:
Tested on 0.1.0/1.0.1. Time is still wrong. Looking at the Stagenet, time of the effective change is before the actual payday while it should be on the payday.

mh-concordium:
Here is an example of baker 4146. Hovering over the '15 days' shows exact time and that time doesn't equal to payday time.

![image](https://user-images.githubusercontent.com/98821384/226432904-f2cd4831-6c8e-47b8-a171-f8d4f7611b9a.png)

abizjak:
I think it is at least very unexpected for users to show the "effective" time in the hover. Nothing really happens at that time, and the baker is not removed.

I think the correct thing to do in the tooltip would be to show when the baker is actually removed, which is the payday after the effective time.

PR #43, opened 2022-12-25
Incorrect Baker Removal Time

parv0888:
## Purpose
Fixes https://github.com/Concordium/concordium-scan/issues/7

## Changes
Showing Max(effectiveRemovalTime, PaydayTime) - Now

## Checklist

      hard-to-understand areas.

## CLA acceptance

- [x] I accept the above linked CLA.

### DIFF
--- a/frontend/src/queries/useBakerQuery.ts
+++ b/frontend/src/queries/useBakerQuery.ts
@@ -18,10 +18,17 @@ export type BakerWithAPYFilter = Baker & {
 
 type BakerResponse = {
 	bakerByBakerId: BakerWithAPYFilter
+	paydayStatus: {
+		nextPaydayTime: string
+	}
 }
 
 const BakerQuery = gql<BakerResponse>`
 	query ($bakerId: Long!) {
+		paydayStatus {
+			nextPaydayTime
+		}
+
 		bakerByBakerId(bakerId: $bakerId) {
 			id
 			bakerId
--- a/frontend/src/utils/format.ts
+++ b/frontend/src/utils/format.ts
@@ -104,6 +104,16 @@ export const convertTimestampToRelative = (
 		addSuffix,
 	})
 
+export const maxTimestamp = (timestamp1: string, timestamp2: string) => {
+	const t1 = parseISO(timestamp1)
+	const t2 = parseISO(timestamp2)
+	if (t1 > t2) {
+		return timestamp1
+	}
+
+	return timestamp2
+}
+
 /**
  * Converts microCCD to CCD with fixed decimals
  * @param {number} amount - Value in microCCD
--- a/frontend/src/components/BakerDetails/BakerDetailsContainer.vue
+++ b/frontend/src/components/BakerDetails/BakerDetailsContainer.vue
@@ -11,11 +11,13 @@
 			data.bakerByBakerId.state.pool
 		"
 		:baker="data.bakerByBakerId"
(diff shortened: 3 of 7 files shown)


frontend/src/utils/format.ts
```diff
@@ -104,6 +104,26 @@ export const convertTimestampToRelative = (
...
+	const nextDayPayDayTime = new Date(
+		Date.UTC(
+			nextDay.getUTCFullYear(),
+			nextDay.getUTCMonth(),
+			nextDay.getUTCDate(),
+			paydayTime.getUTCHours(),
+			paydayTime.getUTCMinutes(),
+			paydayTime.getUTCSeconds(),
+			paydayTime.getUTCMilliseconds()
+		)
```
abizjak, line 121: I'm not sure why this is correct. Suppose that `time` is on day X at 9:59. And paydays are each day at 10.

Then would this not return day `X+1` at 10? Why is this correct for the purposes of computing baker removal time?

I think you need something like
```
if time is before paydayTime on the same day then use paydayTime
otherwise use paydayTime + 1 day
```

abizjak: Perhaps I misunderstood the context.

parv0888: @abizjak - thanks its been fixed

frontend/src/utils/format.ts
```diff
@@ -104,6 +113,37 @@ export const convertTimestampToRelative = (
...
+ * Calculates effective time of the event.
+ * When event effectively happens on the next payday after the event
+ * @param {string} timestamp Time of the event
+ * @param {string} nextPaydayTime Time of the next payday. This should be take from the most recent block ideally.
+ * @returns {string} Effective time of the event. Time when event will actually take place.
+ */
+export const tillNextPayday = (timestamp: string, nextPaydayTime: string) => {
+	const time = parseISO(timestamp)
+	const paydayTime = parseISO(nextPaydayTime)
+	const nextDay = addDays(time, 1)
```
abizjak, line 126: Ok, so this is going to work at the moment, but it is not entirely correct.

In order for it to be correct you'd have to get the paydayDuration from the chain. paydayDuration is not fixed, and is a chain parameter. At the moment it is 1 day, and will likely be 1 day for the foreseeable future.

Is that very hard to do?

parv0888: Should not be @abizjak 
Can you help us with a property thats returned by json having the information for payday duration and not next payday time? is `rewardPeriodLength` is good example?

abizjak: Yes, it is rewardPeriodLength. I think that is in "epochs" so you need to convert it to seconds.

parv0888: fixed

frontend/src/queries/useBakerQuery.ts
```diff
@@ -18,10 +18,27 @@ export type BakerWithAPYFilter = Baker & {
 
 type BakerResponse = {
 	bakerByBakerId: BakerWithAPYFilter
+	paydayStatus: {
+		nextPaydayTime: string
+	}
+	latestChainParameters: {
+		rewardPeriodLength?: number
```
parv0888, line 25: Get epoch duration from getConsensusStatus

parv0888: done

abizjak (approved):

PR #83, opened 2023-06-21
Fix capital bound division by zero issue

schwartz-concordium:
## Purpose

A update to pool parameter has caused CCD scan to crash. This is due to a division by zero exception.

## Changes

A conditional check is added which avoids division by zero when capital bopund is set to 1.


shjortConcordium (approved):
LGTM

PR #88, opened 2023-08-01
Update to Protocol v6

schwartz-concordium:
## Purpose

Update to align with new consensus protocol and hence changes done in gRPC API
https://github.com/Concordium/concordium-grpc-api/commit/13e23942735a427b188b6a27909044ff6493f50a and SDK https://github.com/Concordium/concordium-net-sdk/pull/58.

## Changes

- Chain Parameters
   - Added Chain Parameter v2 used by protocol 6.
   - Election Difficulty
       - Not present as default in chain parameters
   - Gas Finalization Proof made nullable since only present in Chain Parameter v1 and v0.
   - Recreated materialized view which calculates pool compound rates and altered columns to be given as doubles.

## Checklist

      hard-to-understand areas.


backend/DatabaseScripts/SqlScripts/0041_Recreate_materialized_view_graphql_pool_mean_apys_due_to_double_cast.sql
```diff
@@ -0,0 +1,73 @@
...
+used by Entity Framework when mapping.
+Changes below cast calculated values to PostgreSQL 'double' values such that they can be mapped.
+
+See link for PostgreSQL numeric datatypes
+https://www.postgresql.org/docs/current/datatype-numeric.html
+*/
+
+DROP MATERIALIZED VIEW IF EXISTS graphql_pool_mean_apys;
+
+create materialized view graphql_pool_mean_apys as
```
ali-concordium, line 13: I'm not sure how well this view will scale. It might not be important either.

The reason is we have a lot of sub-queries, so if data_30_days grows in size, the sub-queries will become more expensive. It's hard to determine without knowing the size of datasets we'll be working with.

Is the view used in any of the automated tests? If not, it might be a good idea to add a test, to ensure it will if any of the dependent tables/columns are modified.

ali-concordium (approved):

backend/DatabaseScripts/SqlScripts/0041_Recreate_materialized_view_graphql_pool_mean_apys_due_to_double_cast.sql
schwartz-concordium: Talk with @ali-concordium and will merge without integration test. Test has been done manually against environment.

PR #89, opened 2023-08-10
Bug fixes for CCD Scan Protocol 6 update regression test

schwartz-concordium:
## Purpose

Bug fixes found at regression test related to protocol 6 updates.

## Changes

* https://concordium.atlassian.net/browse/CBW-1238
  https://concordium.atlassian.net/browse/CBW-1237
  https://concordium.atlassian.net/browse/CBW-1229
   * Exception was thrown when asking for passive delegation is chain parameters wasn’t version 1. In protocol 6 we have chain parameter version 2 which had the exception thrown. Fix this.
   * Due to currently a short durations of payday cycle APY’s explode. In the DB columns was numeric which couldn’t be parsed to C# doubles.
A related fix for a view has already been done for migration 0041, and same fix, changing columns to type double in DB, has now been done for affected table.
* Fx file permissions on `keep_alive.sh` needed for DB restore which by mistake changed in former PR.

## Checklist

      hard-to-understand areas.


orhoj (approved):

backend/Application/Api/GraphQL/ChainParameters.cs
```diff
@@ -89,6 +89,37 @@ public override int GetHashCode()
         }
     }
 
+    /// <summary>
+    /// Present from protocol version 4 and above and hence from chain parameters 1 and above.
+    /// </summary>
+    internal static bool TryGetPassiveCommissions(
```
ali-concordium, line 95: I think it's worth being consistent in which flavor of switch cases we use. Elsewhere I've seen switch expressions.
Also it seems we're missing a test for this method.

ali-concordium (approved):

backend/Application/Api/GraphQL/ChainParameters.cs
schwartz-concordium: Good point added test :-)

PR #90, opened 2023-08-14
[CBW-1244] Add two transaction events when account is created

schwartz-concordium:
## Purpose

* https://concordium.atlassian.net/browse/CBW-1244
   Add two transaction events when account is created to align with how it is before protocol update.

## Checklist

      hard-to-understand areas.


backend/Tests/Api/GraphQL/Import/TransactionsWriterTest.cs
```diff
@@ -216,16 +216,32 @@ public async Task TransactionEvents_Transferred()
     public async Task TransactionEvents_AccountCreated()
```
shjortConcordium, line 216: I guess the name should be changed? :thinking: 
(Same for the other test)

shjortConcordium (approved):

Issue #78, opened 2023-05-08
CCDScan: Change "Finalization Time"


sdkonerira:
Update how "Finalization time" is calculated and displayed in CCD scan using the following formula:

- take the block slot time for the last finalized block (i.e., highest finalized block we have in the network)
- take the reported finalization time for this block for all the nodes that have it and take the median or average
- display the difference

dg-concordium:
You should put this in the CCDScan repo, not documentation

schwartz-concordium:
One should look at the last finalized block hash on block info. Say one block has finalized block hash A and then the next different finalized block hash i B. 
Then any block between A and B should have finalization time set to block slot time B - block slot time A.

This PR updates the logic to follow above mentioned
 https://github.com/Concordium/concordium-scan/pull/91

schwartz-concordium:
Done in backend 1.8.12 and frontend 1.5.34

PR #91, opened 2023-08-15
Change calculation of finalization time to use latest block hash info from block info

schwartz-concordium:
## Purpose

Update calculation of finalization time to work with protocol 6.
* https://concordium.atlassian.net/browse/CBW-1231

## Changes

Currently finalization summaries are queried from the block but one can just look at the latest finalized block hash on the block info returned from get block info node query.

Changes are needed since finalization summaries isn't present in protocol 6. Using latest finalization block hash in
block info is a robust calculation in both old (before protocol 6) and new consensus protocol. 

## Checklist

      hard-to-understand areas.


td202 (approved):
Looks good.

schwartz-concordium:
Before merge, add back finalization time chart which was removed in PR
https://github.com/Concordium/concordium-scan/pull/92

schwartz-concordium:
https://github.com/Concordium/concordium-scan/issues/78

schwartz-concordium:
If this is merged as is, then all blocks, which doesn’t have a finalization time set, will have finalization time set to the first different finalized block hash seen (which is around now) and the latest finalized block hash (which is before protocol 6).

Because of this a lot of block will get finalization time set to several months.

I migration job should be added, such that all blocks which hasn’t set their time should have them correctly set. This should be done by iterating from blocks after protocol 6.

schwartz-concordium:
Close in favour of https://github.com/Concordium/concordium-scan/pull/181 which has migration job

PR #96, opened 2023-08-25
Adding observability

schwartz-concordium:
## Purpose

Adding observability to new smart contract functionality, such that performance can be observed.
https://concordium.atlassian.net/browse/CBW-1249
## Changes

* Added Prometheus metrics which give durations of imports, count of relevant processed transaction events and retry counts.
* Added health checks, where application goes into degraded state if some job stops running.

## Checklist

      hard-to-understand areas.


backend/Application/Aggregates/Contract/ContractAggregate.cs
```diff
@@ -73,7 +86,11 @@ internal async Task NodeImportJob(IContractNodeClient client, CancellationToken
         }
     }
 
-    internal async Task NodeImport(
+    /// <summary>
+    /// For a given height fetched state from node and store relevant events.
```
ali-concordium, line 90: Rename "fetched" to "fetch"

backend/Application/Aggregates/Contract/Jobs/ContractDatabaseImportJob.cs
```diff
@@ -74,6 +85,27 @@ public string GetUniqueIdentifier()
...
+                {
+                    ContractMetrics.SetReadHeight(latest.BlockHeight, ImportSource.DatabaseImport);
+                }
+
+                await Task.Delay(_contractAggregateOptions.MetricDelay, token);
+            }
+        }
+        catch (TaskCanceledException)
+        {
+        }
```
ali-concordium, line 106: We should at the very least have a code comment explaining why it's empty, despite it seeming obvious from the context.

backend/Application/Aggregates/Contract/Observability/ContractMetrics.cs
```diff
@@ -0,0 +1,94 @@
+using System.Diagnostics;
+using Application.Aggregates.Contract.Types;
+using Prometheus;
+
+namespace Application.Aggregates.Contract.Observability;
+
+internal static class ContractMetrics
```
ali-concordium, line 7: If possible, make this class generic and place it in a class library, so it's not tied to Contracts and so it can be used elsewhere.

ali-concordium (approved):

backend/Application/Aggregates/Contract/Observability/ContractMetrics.cs
schwartz-concordium: I think having the metric code for smart contract “close” to other contract code adheres to done elsewhere for contract implementation.

The Duration metric could be made generic but then I would need to make the property ReadDuration non private. I will keep it for now and refactor when more aggregates are added.

PR #108, opened 2023-09-26
Fix overlapping Search Results

schwartz-concordium:
## Purpose

Due to introduction of new tab ‘Contract’ the search result columns was overlapping. This PR fixes this.
Bug
- https://concordium.atlassian.net/browse/CBW-1364

## Changes

- Change Search Result View from going to full screen at screen size LG (1024) instead of XL (1280).
- Change navigation tabs from going to burger style at screen size less than XL (1280) instead of first at LG (1024)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/frontend/src/components/Search/Search.vue
+++ b/frontend/src/components/Search/Search.vue
@@ -371,7 +371,7 @@ const getTooltipPosition = (index: number) => {
 /**
  * Class used for search results.
  *
- * When the windows is less then `Breakpoint.LG` the search results has full page width.
+ * When the screen size is less then `Breakpoint.LG` the search results has full page width.
  */
 const getSearchClass = computed(() => {
 	if (breakpoint.value >= Breakpoint.XXL) {
--- a/frontend/src/components/Search/Search.vue
+++ b/frontend/src/components/Search/Search.vue
@@ -45,7 +45,7 @@
 						<div
 							v-for="(block, index) in data.search.blocks.nodes"
 							:key="block.blockHash"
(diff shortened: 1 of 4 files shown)


frontend/src/components/Search/Search.vue
```diff
@@ -259,9 +264,11 @@ import ContractLink from '~/components/molecules/ContractLink.vue'
 import { useDateNow } from '~/composables/useDateNow'
 import type { Position } from '~/composables/useTooltip'
 import NodeLink from '~/components/molecules/NodeLink.vue'
+import { Breakpoint } from '~~/src/composables/useBreakpoint'
```
gitguillebetis, line 267: is it needed?

frontend/src/components/Search/Search.vue
```diff
@@ -259,9 +264,11 @@ import ContractLink from '~/components/molecules/ContractLink.vue'
...
 
 const { NOW } = useDateNow()
 const drawer = useDrawer()
+const { breakpoint } = useBreakpoint()
```
gitguillebetis, line 271: is it needed?

schwartz-concordium: no I have removed it - nice catch :-)

frontend/src/components/Search/Search.vue
schwartz-concordium: no I have removed it - nice catch :-)

gitguillebetis (approved):

PR #111, opened 2023-10-04
Add tool tips to new contract pages

schwartz-concordium:
## Purpose

This PR adds tooltips to new contract- and module pages.

## Changes
- Add tooltips to new contract- and module pages
- On contract initialized and update events only show version when present.
- Add parent id to log enrichment for tracing

## Checklist

      hard-to-understand areas.


dg-concordium (approved):
Doris and I made a few small changes and additions.

dg-concordium:
Also, there were a couple of links I added to the contracts documentation but I was unsure how to format the links to work.

frontend/src/components/Contracts/ContractDetailsContent.vue
```diff
@@ -14,13 +14,19 @@
...
+					</template>
 					<template #default>
 						<ModuleLink :module-reference="contract.moduleReference" />
 					</template>
 				</DetailsCard>
 				<DetailsCard>
-					<template #title>Creator</template>
+					<template #title>
+						Creator
+						<InfoTooltip text="Account address of contract creator"/>
```
abizjak, line 28:
```suggestion
						<InfoTooltip text="Account address of the contract instance creator."/>
```

frontend/src/components/Contracts/ContractDetailsEvents.vue
```diff
@@ -115,6 +105,31 @@ type Props = {
...
 defineProps<Props>()
 
+function getResumedLabel(resumed: boolean) : string {
+	return resumed ? "Sucessfully resumed" : "Failed"
+}
+
+function getEventTooltip(eventType: string) {
+	switch (eventType) {
+        case 'ContractInitialized':
+            return "Contract has been initialized on-chain";
```
abizjak, line 115:
```suggestion
            return "Contract instance has been initialized on-chain";
```

frontend/src/components/Contracts/ContractDetailsEvents.vue
```diff
@@ -115,6 +105,31 @@ type Props = {
...
+        case 'ContractUpdated':
+            return "Contract has been updated based on receive function";
```
abizjak, line 117:
```suggestion
            return "Contract has been updated";
```

frontend/src/components/Contracts/ContractDetailsEvents.vue
```diff
@@ -115,6 +105,31 @@ type Props = {
...
+        case 'ContractModuleDeployed':
+            return "Contract module has been deployed on-chain";
+        case 'ContractCall':
+            return "Contract has read from or written to an entrypoint on another contract";
```
abizjak, line 121:
```suggestion
            return "Contract has invoked an entrypoint of another contract";
```

frontend/src/components/Contracts/ContractDetailsEvents.vue
```diff
@@ -115,6 +105,31 @@ type Props = {
...
+        case 'ContractUpgraded':
+            return "Contract module has been upgraded";
+        case 'ContractResumed':
+            return "Contract continued execution after interrupt";
+        case 'Transferred':
+            return "CCD balance reduced on contract by transfer to another contract or account";
```
abizjak, line 127:
```suggestion
            return "Contract instance has transferred an amount of CCD balance to an account.";
```

frontend/src/components/Contracts/Events/ContractInitialized.vue
```diff
@@ -6,28 +6,34 @@
 		</div>
 	</div>
 	<div>
-		<div>Init Name:</div>
+		<div>Init name:
```
abizjak, line 9: This should be contract name instead.

frontend/src/components/Contracts/Events/ContractInitialized.vue
```diff
@@ -6,28 +6,34 @@
...
+			<InfoTooltip text="Initial entrypoint of the contract."/>
```
abizjak, line 10: And this tooltip should say "name of the contract"

frontend/src/components/Details/LogsHEX.vue
```diff
@@ -1,6 +1,8 @@
 <template>
 	<div class="w-full">
-		<div>Logs (HEX):</div>
+		<div>Logs (HEX):
+			<InfoTooltip text="Logs from contract in activity. You can copy the HEX number and decode the log if the contract uses a schema."/>
```
abizjak, line 4:
```suggestion
			<InfoTooltip text="Logs produced by contract execution. You can copy the hex string and decode the log if the contract uses a schema."/>
```

frontend/src/components/Details/MessageHEX.vue
```diff
@@ -1,6 +1,8 @@
 <template>
 	<div class="w-full">
-		<div>Message (HEX):</div>
+		<div>Message (HEX):
+			<InfoTooltip text="Input parameters given to activity entrypoint. TODO (maybe something with HEX)"/>
```
abizjak, line 4: Yes, the same as for logs. But please don't use "HEX" in all-caps. It's not an acronym.

frontend/src/components/Module/ModuleDetailsContractLinkEvents.vue
```diff
@@ -4,7 +4,9 @@
 			<TableTh>Transaction</TableTh>
 			<TableTh>Contract Address</TableTh>
 			<TableTh>Age</TableTh>
-			<TableTh>Action</TableTh>
+			<TableTh>Action
+				<InfoTooltip text="Action which defines if a contract was linked (added) or unlinked (removed) to the module execution code."/>
```
abizjak, line 8: I don't understand this concept. An instance has a code pointer (to the module) at all times. That can be changed by the upgrade call. I don't understand which events this is describing.

frontend/src/components/Module/ModuleDetailsRejectEvents.vue
```diff
@@ -26,49 +26,23 @@
...
-					<div>Init Name:</div>
-					<div>{{ moduleRejectEvent.rejectedEvent.initName }}</div>
-					<div>Module Reference:</div>
-					<div>
-						<ModuleLink
-							:module-reference="moduleRejectEvent.rejectedEvent.moduleRef"
-						/>
+				<div v-if="moduleRejectEvent.rejectedEvent.__typename === 'InvalidInitMethod'">
+					<div>Init name:
+						<InfoTooltip text="Initial entrypoint of the contract."/>
```
abizjak, line 34: Use contract name for this. There is no such thing as "initial entrypoint". Conceptually this is a constructor and no instance exists until this succeeds. It's conceputally different from an entrypoint.

frontend/src/components/Module/ModuleDetailsRejectEvents.vue
```diff
@@ -26,49 +26,23 @@
...
-					<div>Receive Name:</div>
-					<div>{{ moduleRejectEvent.rejectedEvent.receiveName }}</div>
-					<div>Module Reference:</div>
-					<div>
-						<ModuleLink
-							:module-reference="moduleRejectEvent.rejectedEvent.moduleRef"
-						/>
+				<div v-if="moduleRejectEvent.rejectedEvent.__typename === 'InvalidReceiveMethod'">
+					<div>Receive name:
+						<InfoTooltip text="Entrypoint of the activity of the contract."/>
```
abizjak, line 40: I don't understand where this "activity" terminology is coming from. I would say just "entrypoint of the contract instance".

frontend/src/pages/contracts/index.vue
```diff
@@ -6,9 +6,15 @@
...
-					<TableTh align="right">Creator</TableTh>
+					<TableTh align="right">
+						Creator
+						<InfoTooltip text="Account address of contract creator"/>
+					</TableTh>
 					<TableTh align="right">Initial Transaction</TableTh>
-					<TableTh align="right">Module</TableTh>
+					<TableTh align="right">
+						Module
+						<InfoTooltip text="Container which holds execution code for one-or-more contracts. The below references holds the current execution code of the contract."/>
```
abizjak, line 16: This string repeats in a few places. It'd be better to define a constant for it.

frontend/src/components/Module/ModuleDetailsContractLinkEvents.vue
schwartz-concordium: When a contracts upgrades by pointing to module A instead of B two events will be created. One with action removed on module B and one with action added on module A.

abizjak: But why? This is an atomic action and only one event is generated on the chain. Why do we need to display two and introduce new concepts?

schwartz-concordium: Atomic operations can generate multiple events and the data model of the CCD scan shouldn't be a 1-1 representation of the chains data model. The CCD scan’s purpose is to make a projection of the chains state such that the state can be interpreted easily by the users.

It makes sense to show to the users when a contract pointer was removed from one module. If the current table and events wasn’t present the user would need to click a lot around to get the same information.

If you have more questions to the UI and its components you are welcome to come over :-) I don’t think continuing the discussion here will be efficient.

PR #117, opened 2023-10-16
Module Source Import

schwartz-concordium:
## Purpose

Update module references to include module source and schema. 

The logic for extracting the schema has been taken from this code:
https://github.com/Concordium/concordium-dapp-libraries/blob/ddf82b960274d5d6a1f574f4f4586ae2465d26de/samples/contractupdate/src/useContractSchemaRpc.ts#L13

## Changes

- Contract node import job to enrich modules with source and schema
- Added job which updates existing

## Checklist

      hard-to-understand areas.


backend/Application/Aggregates/Contract/Resilience/Policies.cs
```diff
@@ -0,0 +1,76 @@
+using System.Threading.Tasks;
+using Npgsql;
+using Polly;
+
+namespace Application.Aggregates.Contract.Resilience;
+
+internal static class Policies
+{
+    /// <summary>
+    /// Create a async retry policy which retries on all transient database errors.
```
shjortConcordium, line 10:
```suggestion
    /// Create an async retry policy, which retries on all transient database errors.
```
:mag_right:

backend/Application/Aggregates/Contract/Resilience/Policies.cs
```diff
@@ -0,0 +1,76 @@
...
+                    });
+        }
+
+        return policy;
+    }    
+    
+    /// <summary>
+    /// Create a async retry policy which retries on all transient database errors.
+    /// </summary>
+    internal static AsyncPolicy<T> GetTransientPolicy<T>(ILogger logger, int retryCount, TimeSpan delay)
```
shjortConcordium, line 46: Why do you need to copy the entire function here for a "Generic version"? :thinking: 
( / Or why do you need the non-generic)

backend/Application/Aggregates/Contract/Jobs/ParallelBatchBlockHeightJob.cs
```diff
@@ -41,11 +41,14 @@ ContractHealthCheck healthCheck
...
-    
+
+    /// <inheritdoc/>
+    public bool ShouldNodeImportAwait() => _statelessJob.ShouldNodeImportAwait();
+
     public async Task StartImport(CancellationToken token)
     {
         using var _ = TraceContext.StartActivity(GetUniqueIdentifier());
-        using var __ = LogContext.PushProperty(nameof(ParallelBatchBlockHeightJob<TStatelessJob>), GetUniqueIdentifier());
+        using var __ = LogContext.PushProperty("Job", GetUniqueIdentifier());
```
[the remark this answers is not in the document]
schwartz-concordium: The above is a bug. This makes a key - value index to the logs, and in above the key was dynamic across jobs. We don’t want that.

backend/Application/Aggregates/Contract/Resilience/Policies.cs
schwartz-concordium: This is because of the API’s used in the framework. It has two difference builders dependent if there is a return type or not.

backend/Application/Aggregates/Contract/Resilience/Policies.cs
```diff
@@ -0,0 +1,76 @@
...
+                        logger.Error(ex, $"Triggering retry policy with {currentRetryCount} due to exception");
+                        return Task.CompletedTask;
+                    });
+        }
+
+        return policy;
+    }    
+    
+    /// <summary>
+    /// Create a async retry policy which retries on all transient database errors.
```
shjortConcordium, line 44:
```suggestion
    /// Create an async retry policy, which retries on all transient database errors.
```
I guess its the same here, sorry :thinking:

backend/Application/Aggregates/Contract/Resilience/Policies.cs
shjortConcordium: Okay, it just seems weird that you have to implement the (seemingly) same function twice :shrug:

shjortConcordium (approved):
LGTM

PR #119, opened 2023-10-20
Display module formatted module schema

schwartz-concordium:
## Purpose

This PR add FFI from C# to rust such that schema, events and contract receive parameters can be deserialised.

GraphQL endpoints are added for module such that a formatted schema, if present in module, can be given to the UI.

The UI is updated to show the formatted schema.

Note FFI and module schema extraction will be migrated to the .NET SDK when contract views are in place. It is first implemented partially in CCD Scan to polish and “battle” test the API’s. 

https://github.com/Concordium/concordium-scan/assets/132270889/abec44cc-c3b5-406c-85a2-29629f73bce0

https://concordium.atlassian.net/browse/CBW-1253
https://concordium.atlassian.net/browse/CBW-1256

## Changes

* Add FFI which deserialise schema, events and receive parameters
* Update GraphQL endpoints for module
* Update module UI to show formatted schema when present.

## Checklist

      hard-to-understand areas.


backend/rust-bindings/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+                *result = CString::new(e.to_string()).unwrap().into_raw()    
+            }
+            return false
+        }
+    };
+    unsafe {
+        *result = CString::new(display.to_string()).unwrap().into_raw()
+    }
+
+    true
```
shjortConcordium, line 63: I've created a PR to "simplify" this a bit.

backend/Application/Aggregates/Contract/Entities/ModuleReferenceEvent.cs
```diff
@@ -184,6 +187,21 @@ public class ModuleReferenceEventQuery
     [ExtendObjectType(typeof(ModuleReferenceEvent))]
     public sealed class ModuleReferenceEventExtensions
     {
+        /// <summary>
+        /// Returns module schema in a human interpretable form. Once present if the schema is embedded into the
+        /// Wasm module.
+        /// </summary>
+        public string? GetDisplaySchema([Parent] ModuleReferenceEvent module)
```
shjortConcordium, line 194: This function doesn't seem to be used, and I don't understand (and can't find) what the `[Parent]` does? :pleading_face:

Makefile
```diff
@@ -0,0 +1,2 @@
+rb:
```
bisgardo, line 1:
```suggestion
build-rust-bindings:
```

no need for  people having to guess what this means :)

.gitmodules
bisgardo, line 1: I think we should look into getting the relevant parts on crates.io at some point. Not asking for this to be changed, just voicing my dislike that it's necessary.

backend/Application/Aggregates/Contract/Entities/Contract.cs
```diff
@@ -20,6 +20,7 @@ public sealed class Contract : BaseIdentification
     public ulong ContractAddressIndex { get; init; }
     public ulong ContractAddressSubIndex { get; init; }
     public string ContractAddress { get; init; } = null!;
+    [GraphQLIgnore]
```
bisgardo, line 23: What does this do exactly and why is it part of this PR?

shjortConcordium (commented):
LGTM, have a few comments.
Though I'm not sure about why the [GraphQLIgnore] annotations are being added :thinking:

backend/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Unstable features
 - GraphQL endpoints for contracts and modules should currently be seen as unstable, indicating that the query models are likely to undergo changes.
```
bisgardo, line 4: I think "state of the application" comments like this belong in the readme, not changelog which should be an append-only log.

.github/workflows/ci.yml
```diff
@@ -4,15 +4,31 @@ on:
...
-        fetch-depth: 0
+        profile: minimal
+        toolchain: ${{ env.RUST_VERSION }}
+        override: true
+
+    - name: Build Rust Project
+      run: cargo build --manifest-path=./backend/rust-bindings/Cargo.toml --release
+    
+    - name: Test Rust Project
+      run: cargo test --manifest-path=./backend/rust-bindings/Cargo.toml --release
```
shjortConcordium, line 31:
```suggestion
    - name: Test Rust Project
      run: cargo test --manifest-path=./backend/rust-bindings/Cargo.toml --release

    - name: Clippy Rust Project
      run: cargo clippy --manifest-path=./backend/rust-bindings/Cargo.toml --color=always --tests

    - name: Format Rust Project
      run: cargo fmt --manifest-path=./backend/rust-bindings/Cargo.toml --check
```
I would also add clippy and format checks for the CI.

backend/Tests/Aggregates/Contract/Types/ModuleSchemaVersionTests.cs
```diff
@@ -0,0 +1,30 @@
...
+using FluentAssertions;
+
+namespace Tests.Aggregates.Contract.Types;
+
+public sealed class ModuleSchemaVersionTests
+{
+    [Theory]
+    [InlineData(ModuleSchemaVersion.Undefined, (byte)0)]
+    [InlineData(ModuleSchemaVersion.V0, (byte)0)]
+    [InlineData(ModuleSchemaVersion.V2, (byte)2)]
```
bisgardo, line 11: Does this turn the test into multiple cases?

backend/Tests/Aggregates/Contract/Types/ModuleSchemaVersionTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        // Assert
+        if (ModuleSchemaVersion.Undefined == version)
+        {
+            ffiOption.is_some.Should().Be(0);
+        }
+        else
+        {
+            ffiOption.is_some.Should().Be(1);
+            ffiOption.t.Should().Be(mapped);
+        }
```
bisgardo, line 28: Not a big problem here, but I think conditions in tests should be avoided. It would be much clearer to me if it was just 3 cases written directly.

backend/Tests/Aggregates/Contract/Types/ModuleSchemaVersionTests.cs
schwartz-concordium: Yes

backend/Application/Aggregates/Contract/Entities/Contract.cs
schwartz-concordium: I’m updating the module schema and to have it consistent I’m also updating other entities.

backend/Application/Aggregates/Contract/Entities/ModuleReferenceEvent.cs
schwartz-concordium: Explained in person

backend/Application/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+    
+    /// <summary>
+    /// Wrap a interop result. If the interop call <see cref="Succeeded"/> the <see cref="Message"/> contains the result.
+    /// If it failed the <see cref="Message"/> will contain the error message. 
+    /// </summary>
+    /// <param name="Message">Message from interop operation.</param>
+    /// <param name="Succeeded">If the call succeeded.</param>
+    internal readonly record struct InteropResult(string? Message, bool Succeeded);
```
bisgardo, line 99: Why don't we use exceptions for errors (or whatever is conventional error handling for C#)?

schwartz-concordium:
> LGTM, have a few comments. Though I'm not sure about why the [GraphQLIgnore] annotations are being added 🤔

I’m added them since they shouldn’t be exposed in the graphql schema. To be consistent I have removed them not needed from all entities.

.gitmodules
schwartz-concordium: I agree; I would much rather depend on some published crates. This is a bigger discussion, I guess. :-)

backend/Application/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,131 @@
...
+    
+    private static void Free(IntPtr ptr)
```
bisgardo, line 101:
```suggestion
    private static void FreeIfNonzero(IntPtr ptr)
```
maybe?

backend/Application/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            Marshal.FreeHGlobal(ptr);
+        }
+    }
+    
+    /// <summary>
+    /// A C# layout which compiled to a C interpretable structure. This is used as an optional parameter.
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct FFIOption
```
bisgardo, line 113: Maybe let name indicate that it's an option of byte?

.github/workflows/ci.yml
```diff
@@ -4,15 +4,38 @@ on:
...
+      uses: actions-rs/toolchain@v1
       with:
-        fetch-depth: 0
+        profile: minimal
+        toolchain: ${{ env.RUST_VERSION }}
+        override: true
+        components: rustfmt, clippy
+
+    - name: Build Rust Project
+      run: cargo build --manifest-path=./backend/rust-bindings/Cargo.toml --release
```
abizjak, line 29: Clippy will also cover this, no need for this step.

.github/workflows/ci.yml
```diff
@@ -4,15 +4,38 @@ on:
...
+    
+    - name: Clippy Rust Project
+      run: cargo clippy --manifest-path=./backend/rust-bindings/Cargo.toml --tests
+
+    - name: Format Rust Project
+      run: cargo fmt --manifest-path=./backend/rust-bindings/Cargo.toml --check
```
abizjak, line 35: You should format first since it's cheaper than clippy.

backend/Application/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,131 @@
...
+    ///
+    /// Receive parameters are those given to a contract entrypoint on a update call.
+    /// </summary>
+    /// <param name="schema">Module schema in hexadecimal</param>
+    /// <param name="contractName">Contract name</param>
+    /// <param name="entrypoint">Entrypoint of contract</param>
+    /// <param name="value">Receive parameters in hexadecimal</param>
+    /// <param name="schemaVersion">Optional schema version if present from module</param> 
+    /// <returns>Receive parameters in a human interpretable form</returns>
+    internal static InteropResult GetReceiveContractParameter(string schema, string contractName, string entrypoint, string value, FFIOption schemaVersion)
```
bisgardo, line 55: I would take `schemaVersion` as a non-FFI type and do the wrapping in here to avoid leaking the low level types. Same in other cases.

.gitmodules
abizjak: As it happens concordium_base is published on crates.io. 

Nevertheless we use submodules for internal dependencies since they make it much less cumbersome to do updates in multiple components at the same time without having to publish every minor thing before it is mature. So please stick to that.

Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[workspace]
+members=["backend/rust-bindings"]
+resolver="2"
```
abizjak, line 3: Could you please not use workspaces where there is not a critical need? Espeically a workspace with one member is just overhead.

backend/Application/Aggregates/Contract/Entities/Contract.cs
schwartz-concordium: Yes it removed the field from the graphql schema. Public properties are included as default and needs to be removed explicitly.

backend/Application/Interop/InteropBinding.cs
schwartz-concordium: Exceptions are expensive and since we either way have a result structure I think it is find to have a functional pattern here.

Cargo.toml
schwartz-concordium: When opening the project from root the rust analyser starts. Is there an overhead having it?

backend/Application/Interop/InteropBinding.cs
schwartz-concordium: I’m only using a byte for boolean because it can’t be directly mapped and a byte for the enum value. I rather want to call it something like `FFIEnumOption` if I should update it, since this maybe represents better usages of it :-)

backend/Application/Interop/InteropBinding.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <param name="schema">Module schema in hexadecimal</param>
+    /// <param name="schemaVersion">Optional schema version if present from module</param> 
+    /// <returns>Module schema in a human interpretable form</returns>
+    internal static InteropResult SchemaDisplay(string schema, FFIOption schemaVersion)
+    {
+        var result = IntPtr.Zero;
+        try
+        {
+            var schemaDisplay = schema_display(schema, schemaVersion, ref result);
+            var ptrToStringAnsi = Marshal.PtrToStringAnsi(result);
```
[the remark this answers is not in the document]
schwartz-concordium: Yes a little nit - but everything for some nice code :-)

backend/Application/Interop/InteropBinding.cs
bisgardo: > Exceptions are expensive and since we either way have a result structure I think it is find to have a functional pattern here.

That sounds like a premature optimization to me. Do we avoid exceptions in general in the app? If not, then why is this different?

My understanding is that exceptions being slow only matters if you're using it for normal control flow. And more so in C++ than languages like C#. Either way I prefer following standard conventions until you know that there's an actual problem.

If we want to use result types there's one here: https://dotnet.github.io/dotNext/api/DotNext.Result-1.html.

backend/Application/Interop/InteropBinding.cs
bisgardo: I'm referring to the byte value `t`, not the `is_some` boolean.

Calling it `FFIEnumOption` would only be accurate for enums that happen to turn into a byte (like schema version), so I think that would be misleading.

backend/Application/Interop/InteropBinding.cs
schwartz-concordium: I don’t want to add another dependency just to have a result wrapper :-)

I changed to exception pattern.

README.md
```diff
@@ -44,4 +44,15 @@ See the description of `CCDSCAN_BACKEND_PORT` for an explanation of why Mac user
...
 
-In `./timescaledb-restore` a description is given how to spin up and restore CCD scan database locally from Stage- or Testnet backups.
\ No newline at end of file
+In `./timescaledb-restore` a description is given how to spin up and restore CCD scan database locally from Stage- or Testnet backups.
+
+## Rust-bindings
+
+The CCD scan does interops to compiled code from rust. The `.csproj` file is dependent on compiled binaries which is why one manually needs to run below
+```
+make rb
```
shjortConcordium, line 53:
```suggestion
make build-rust-bindings

```

shjortConcordium (approved):
LGTM.

bisgardo (approved):
Looks very good. I appreciate the screen capture!

Cargo.toml
abizjak: There is a mental overhead. Now there is yet another configuration file that controls behaviour.

I am not sure what you mean by the rust-analyzer. At least in emacs you can add arbitrary folders to the workspace so this is not neeed.

PR #121, opened 2023-10-23
Refactor rust-bindings to reduce glue code

shjortConcordium:
## Purpose

Refactor of interop glue code. (Easier to show it)

I'm not sure it needs to use a closure, but it was mainly to isolate the handling of the results, and writing the to result variable.

### DIFF
--- a/backend/rust-bindings/src/lib.rs
+++ b/backend/rust-bindings/src/lib.rs
@@ -25,6 +25,25 @@ impl FFIOption {
     }
 }
 
+fn handle_result<F, T>(result: *mut *mut c_char, f: F) -> bool where F: FnOnce() -> Result<T>, T: ToString {
+    match f() {
+        Ok(output) => {
+            unsafe {
+                *result = CString::new(output.to_string()).unwrap().into_raw()
+            }
+
+            true
+        },
+        Err(e) => {
+            unsafe {
+                *result = CString::new(e.to_string()).unwrap().into_raw()
+            }
+
+            false
+        }
+    }
+}
+
 /// Get module schema in a human interpretable form.
 /// 
 /// # Arguments
@@ -41,26 +60,11 @@ pub extern "C" fn schema_display(
     schema: *const c_char,
     schema_version: FFIOption,
     result: *mut *mut c_char) -> bool {
-    
-    let schema_hex = match get_str_from_pointer(schema, result) {
-        Ok(out) => out,
-        Err(_) => return false,
-    };
 
-    let display = match schema_display_aux(schema_hex, schema_version.into_option()) {
-        Ok(display) => display,
-        Err(e) => {
-            unsafe {
-                *result = CString::new(e.to_string()).unwrap().into_raw()    
-            }
-            return false
-        }
-    };
-    unsafe {
-        *result = CString::new(display.to_string()).unwrap().into_raw()
-    }
-
-    true
+    handle_result(result, || {
+        let schema_hex = get_str_from_pointer(schema)?;
+        schema_display_aux(schema_hex, schema_version.into_option())
+    })
 }
 
 #[no_mangle]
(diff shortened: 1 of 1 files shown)


backend/rust-bindings/src/lib.rs
```diff
@@ -25,6 +25,25 @@ impl FFIOption {
     }
 }
 
+fn handle_result<F, T>(result: *mut *mut c_char, f: F) -> bool where F: FnOnce() -> Result<T>, T: ToString {
```
bisgardo, line 28:
```suggestion
fn assign_result<T: ToString>(target: *mut *mut c_char, src: Result<T>) -> bool {
```

I don't see the need for taking a function; it could just take the `Result<T>`? In it's current form, the function doesn't just "handle" the result, it also creates it.

backend/rust-bindings/src/lib.rs
```diff
@@ -213,17 +178,9 @@ fn deserialize_type_value(
...
-    let str_slice: &str = match c_str.to_str() {
-        Ok(r) => r,
-        Err(e) => {
-            unsafe {
-                *result = CString::new(e.to_string()).unwrap().into_raw()
-            }
-            return Err(anyhow!("not able to parse pointer to string"))
-        },
-    };
+    let str_slice: &str = c_str.to_str()?;
```
bisgardo, line 183: no need for adding context to error message?

bisgardo (commented):
Great simplification though with a little too many empty lines for my taste ;)

backend/rust-bindings/src/lib.rs
shjortConcordium: I've changed it to only take the result, it doesn't feel as clean, but it avoids passing functions. :shrug:

backend/rust-bindings/src/lib.rs
shjortConcordium: I just wanted to refactor, I've only moved the place the result is written to. :shrug:

backend/rust-bindings/src/lib.rs
```diff
@@ -41,26 +60,11 @@ pub extern "C" fn schema_display(
...
-            }
-            return false
-        }
-    };
-    unsafe {
-        *result = CString::new(display.to_string()).unwrap().into_raw()
-    }
-
-    true
+    assign_result(result, (|| {
```
schwartz-concordium, line 64: IMHO this doesn’t look more clean. Here a closure is defined which is executed at once.

Passing closures around is something you normally see - ex. every time you use `map`.

backend/rust-bindings/src/lib.rs
bisgardo: The error message before was `"not able to parse pointer to string"`. I don't know what the current msg is (might be worse, might be better) but it looks changed to me.

backend/rust-bindings/src/lib.rs
bisgardo: The solution I had in mind was more like
```
let schema_hex = get_str_from_pointer(schema)?;
let schema_res = schema_display_aux(schema_hex, schema_version.into_option());
assign_result(result, schema_res)
```

backend/rust-bindings/src/lib.rs
shjortConcordium:
```
    let schema_hex = match get_str_from_pointer(schema, result) {
        Ok(out) => out,
        Err(_) => return false,
    };
```
This was how it was used. That message was thrown away anyway. :shrug:

bisgardo: Well if you have to be all whole program analysis about it...

backend/rust-bindings/src/lib.rs
shjortConcordium: Well that wouldn't deal with the results from the `get_str_from_pointer`.
You need to have a closure (or move this stuff into separate functions) to avoid dealing with all the results individually, like was done originally.

bisgardo: makes sense

backend/rust-bindings/src/lib.rs
schwartz-concordium: I think it is more clean now since you handle the error in the callee function. I see now that I could have just written the error and returned the same (which will also not be used but I avoid the custom error).

backend/rust-bindings/src/lib.rs
```diff
@@ -28,6 +28,21 @@ impl FFIOption {
     }
 }
 
+fn assign_result<F: FnOnce() -> Result<T>, T: ToString>(target: *mut *mut c_char, f: F) -> bool {
```
bisgardo, line 31: Maybe add docstring like "Compute result using the provided callback, convert it into a C string and assign it to the provided target. The returned bool indicates whether the computation was successful or not." Or whatever may be more accurate.

I don't think the function name exactly conveys the behavior, but to not having this turn into excessive nitting, I will leave it with that.

bisgardo (commented):
I don't feel like I should be the one approving this but it looks good to me.

schwartz-concordium (approved):

PR #122, opened 2023-10-23
Update dockerfile to include rust build

schwartz-concordium:
## Purpose

After PR https://github.com/Concordium/concordium-scan/pull/119 CCD scan depends on interop calls to rust binaries. These needs to be build in the docker image. 

## Changes

* Dockerfile - Updated image to build rust binaries
* Dockerfile - Remove `dotnet build` and `dotnet restore` since it is included in `dotnet publish` and we don’t use the intermediate steps for anything.
`restore` also didn’t succeed restoring everything since files wasn’t copied in correct order *(copy everything done after `dotnet restore`)*. See this run where `DatabaseScripts.csproj ` wasn’t restored since files wasn’t present *(line `Skipping project "/src/DatabaseScripts/DatabaseScripts.csproj" because it was not found.`)*.
https://jenkins.internal.concordium.com/job/ccdscan-backend-docker-image/44/consoleFull
* Dockerfile - Only copy explicit folders needed instead of everything.
* Add parameter to control rust build as part in `dotnet build`(part of `dotnet publish`)

## Checklist

      hard-to-understand areas.


backend/Application/Application.csproj
```diff
@@ -58,7 +59,7 @@
             </None>
     </ItemGroup>
 
-    <Target Name="BuildRustProject" BeforeTargets="Build">
+    <Target Name="BuildRustProject" BeforeTargets="Build" Condition="$(RunRustBuild)">
         <Exec Command="cargo build --manifest-path=../rust-bindings/Cargo.toml --release" />
     </Target>
```
bisgardo, line 64: Discussed offline, included here for reference: It may be that this command rebuilds the target after if already copied the existing version.

This doesn't matter in the docker build because you build the bindings manually before the dotnet project, but it could be a source of confusion during development.

backend/Dockerfile
```diff
@@ -2,16 +2,18 @@ FROM mcr.microsoft.com/dotnet/aspnet:6.0 AS base
 WORKDIR /app
 EXPOSE 5000
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0 AS build
+FROM rust:1.65 AS rust-builder
 WORKDIR /src
-COPY ["Application/Application.csproj", "Application/"]
-RUN dotnet restore "Application/Application.csproj"
```
bisgardo, line 8: Discussed offline, included here for reference: Calling `dotnet restore` before `COPY` improves chances of a cache hit from the docker daemon.

Doesn't matter for our build because we never reuse build VMs and don't have a distributed image cache, but is generally good practice.

bisgardo (approved):
Please update PR title to reflect that only the Docker build file is updated, not the Jenkins workflow.

PR #127, opened 2023-11-06
Optimize updates on contract event deserialization job

schwartz-concordium:
## Purpose

The contract event deserialization job is slow. This PR does several updates to increase performance like changing to Dapper instead of Entity Framework and use in memory processing.

It has been validated using `EXPLAIN ANALYZE` in Postgres that the update statement correctly uses the index from the primary key:
```
Update on graphql_contract_events  (cost=0.43..2.66 rows=0 width=0) (actual time=153.918..153.937 rows=0 loops=1)
  ->  Index Scan using graphql_contract_events_pkey on graphql_contract_events  (cost=0.43..2.66 rows=1 width=46) (actual time=0.920..0.959 rows=1 loops=1)
        Index Cond: ((block_height = 33552) AND (transaction_index = 0) AND (event_index = 0) AND (contract_address_index = 9) AND (contract_address_subindex = 0))
Planning Time: 137.682 ms
Execution Time: 155.104 ms

```
I have also validated that restricting the count of updates doesn’t improve performance.

## Changes

* Changing to Dapper when updating events and avoid entities being tracked by Entity Framework
* Avoid lookups to database for each events by instead fetching all contract initialization events, module events and module linking events into memory.

## Checklist

      hard-to-understand areas.


backend/Application/Aggregates/Contract/Jobs/InitialContractEventDeserializationFieldsCatchUpJob.cs
```diff
@@ -69,19 +72,33 @@ await Policies.GetTransientPolicy(GetUniqueIdentifier(), _logger, _contractAggre
...
                 {
                     try
                     {
-                        await contractEvent.ParseEvent(contractRepository, moduleReadonlyRepository);
+                        var transactionResultEvent = await TryParseEvent(contractEvent, contractRepository, moduleReadonlyRepository);
+                        if (transactionResultEvent == null)
+                        {
+                            continue;
+                        }
+                        eventUpdates.Add(new Update(
```
orhoj, line 94: I would move this into the conditional statement and check for `transactionResultEvent != null`.

backend/Tests/Aggregates/Contract/Jobs/InitialContractEventDeserializationFieldsCatchUpJobTests.cs
```diff
@@ -158,4 +164,401 @@ ContractAddress contractAddress
...
+        // Act
+        var actualModule = await moduleReadonlyRepository.GetModuleReferenceEventAtAsync(contract, blockHeight, transactionIndex, eventIndex);
+        
+        // Assert
+        actualModule.ModuleReference.Should().Be(moduleRef);
+    }
+    
+    [Fact]
+    public async Task
+        GivenModuleEventAtSameTransactionIndexInMemory_WithDifferentEventIndex_WhenGetModuleReferenceEventAtAsync_ThenReturnLatestFromDatabase()
```
shjortConcordium, line 258: :smiling_face_with_tear:

shjortConcordium (approved):
LGTM, but won't this run into issues when stuff can't fit into memory? :thinking:

schwartz-concordium:
> LGTM, but won't this run into issues when stuff can't fit into memory? 🤔

Yes but currently there aren't that many events. These jobs should be seen as migrations and will only be run once.

PR #129, opened 2023-11-07
Add commission column

schwartz-concordium:
## Purpose

Due to tokenomic changes validators would be able to change validation commissions.
https://concordium.atlassian.net/browse/CBW-1464

## Changes

* Add validator commissions to staking overview

## Checklist

      hard-to-understand areas.


frontend/src/components/Staking/BakerPools.vue
```diff
@@ -61,6 +67,22 @@
 						<AccountLink :address="baker.account.address.asString" />
 					</TableTd>
 
+					<TableTd
+						v-if="hasPoolData && breakpoint >= Breakpoint.SM"
+						align="right"
+					>
+						<span
+							v-if="baker.state.__typename === 'ActiveBakerState'"
```
orhoj, line 75: For the APY it also checks `Number.isFinite(baker.state.pool?.apy.bakerApy)`. Why do we need that check for that, but not something similar here?

schwartz-concordium: Looking at the function declaration it seems like it is used to validate if it is a number with decimals. Other than that I can’t see why it is used.

IMHO we don’t need it since we own both front- and backend.

shjortConcordium (approved):

PR #133, opened 2023-11-14
Fix warning message on Staking page

schwartz-concordium:
## Purpose

If you navigate to the Staking page you get a warning message in the browser console. Try on testnet
https://testnet.ccdscan.io/staking

## Changes

Fix warning message by avoiding boolean to be possible undefined.

### DIFF
--- a/frontend/src/components/Staking/BakerPools.vue
+++ b/frontend/src/components/Staking/BakerPools.vue
@@ -322,7 +322,7 @@ const { first, last, after, before, goToPage, resetPagination } =
 
 type Props = {
 	openStatusFilter: BakerPoolOpenStatus | undefined
-	includeRemoved: boolean | undefined
+	includeRemoved: boolean
 	sort: BakerSort
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/Staking/BakerPools.vue
```diff
@@ -322,7 +322,7 @@ const { first, last, after, before, goToPage, resetPagination } =
 
 type Props = {
 	openStatusFilter: BakerPoolOpenStatus | undefined
-	includeRemoved: boolean | undefined
+	includeRemoved: boolean
 	sort: BakerSort
```
shjortConcordium, line 326: I guess this makes the Prop non-optional. :thinking:

shjortConcordium (approved):
I don't experience the error, but I think it's good to use booleans only like this.

frontend/src/components/Staking/BakerPools.vue
schwartz-concordium: Yes

PR #135, opened 2023-11-15
Fix tooltip for pagination offset

schwartz-concordium:
## Purpose

Fix pagination start tooltip message.
Bug: https://concordium.atlassian.net/browse/CBW-1515


shjortConcordium (approved):
The tooltip doesn't appear for 0, though :thinking:

PR #143, opened 2023-11-27
Commission rate change indicators

schwartz-concordium:
## Purpose

Add commission rate change indicators.

https://github.com/Concordium/concordium-scan/assets/132270889/30314377-2525-42d3-a3b7-77e321dac669

https://concordium.atlassian.net/browse/CBW-1485
https://concordium.atlassian.net/browse/CBW-1484

## Changes

* Add commission rate change indicators to overview and details page
* Fix two wrong usage of “Baker"

## Checklist

      hard-to-understand areas.


frontend/src/queries/useBakerListQuery.ts
```diff
@@ -67,6 +67,11 @@ const BakerQuery = gql<BakerListResponse>`
 								finalizationCommission
 								bakingCommission
 							}
+							paydayCommissionRates {
+								transactionCommission
+								finalizationCommission
```
dg-concordium, line 72: Finalization commission goes away. It is 0 and should not be shown.

frontend/src/queries/useBakerQuery.ts
```diff
@@ -80,6 +80,11 @@ const BakerQuery = gql<BakerResponse>`
 							rank
 							total
 						}
+						paydayCommissionRates {
+							transactionCommission
+							finalizationCommission
```
dg-concordium, line 85: Finalization commission goes away. It is 0 and should not be shown.

frontend/src/queries/useBakerListQuery.ts
```diff
@@ -67,6 +67,11 @@ const BakerQuery = gql<BakerListResponse>`
...
+								bakingCommission
```
dg-concordium, line 73: This will be called Block reward commission for the users.

frontend/src/queries/useBakerQuery.ts
```diff
@@ -80,6 +80,11 @@ const BakerQuery = gql<BakerResponse>`
...
+							bakingCommission
```
dg-concordium, line 86: This will be called Block reward commission for the users.

frontend/src/queries/useBakerQuery.ts
```diff
@@ -80,6 +80,11 @@ const BakerQuery = gql<BakerResponse>`
...
+						}
 						commissionRates {
 							transactionCommission
 							finalizationCommission
```
dg-concordium, line 90: Finalization commission goes away. It is 0 and should not be shown.

dg-concordium (approved):
A couple of comments. Not sure if the queries have to use the old commission names (probably do). Just as long as those don't appear in the GUI.  That was probably in a different PR.

shjortConcordium:
Only one issue I found, when commission is 0%:

Screenshot from 2023-11-27 14-15-21

Screenshot from 2023-11-27 14-14-59

shjortConcordium:
And when I change commission to 0%, the indicator/changes doesn't appear.

frontend/src/queries/useBakerListQuery.ts
schwartz-concordium: Removed (was not used)

frontend/src/queries/useBakerQuery.ts
schwartz-concordium: Removed (was not used anymore)

frontend/src/queries/useBakerListQuery.ts
schwartz-concordium: We have not changed existing fields and classes in the code, where it isn’t effecting the rendering. Changing all code would have been quite comprehensive. 

New classes and fields do however follow the new terminology.

frontend/src/queries/useBakerQuery.ts
schwartz-concordium: Rendering updated but existing code fields and classes not updated (as like comment above).

schwartz-concordium:
> And when I change commission to 0%, the indicator/changes doesn't appear.

Had wrong condition check which captured 0. Replaced with check on `undefined` and `null` and validates it correctly renders `0` and the indicator.

PR #151, opened 2023-12-02
Migrate from firebase

schwartz-concordium:
## Purpose

This PR implements the necessary changes required to migrate away from Firebase as we intend to self-host the CCD scan frontend.

## Changes
* Added a `Dockerfile` to `./frontend` that will build the image required for custom hosting.
* Removed Firebase dependencies.
* Eliminated Firebase references in `README.md` and added a section on how to locally run the Docker image.
* Updated the frontend `Jenkinsfile` to build the Docker image similar to the backend. Note: a variable for setting the target environment is required. Additionally, the Docker repository hasn't been added. The flow could be migrated from Jenkin to Github as a Github Action.

## Checklist

      hard-to-understand areas.


frontend/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Run
+FROM base
+
+ENV PORT=$PORT
+ENV HOST=$HOST
+
+COPY --from=build /src/.output /src/.output
+
+CMD ["node", ".output/server/index.mjs"]
```
lassemand, line 30:
```suggestion
COPY --from=build /src/.output .

CMD ["node", "server/index.mjs"]
```

frontend/Jenkinsfile
```diff
@@ -5,40 +5,33 @@
...
-          yarn lint
-
-          # Build application bundle
-          env ENVIRONMENT="${TARGET_NET}" yarn build
-
-          # Package
-          tar cvzf out.tar.gz firebase.json .output
+          docker build \
+            --label git_commit="$GIT_COMMIT" \
+            --build-arg ENVIRONMENT=${TARGET_NET} \
```
lassemand, line 25: Future request in a different PR. 

Could we make a build for all the different environments, and let it be decided upon runtime which one we use instead? So we have a single docker image.

lassemand (approved):
Looks awesome. Good job and thanks for fast response.

frontend/Dockerfile
```diff
@@ -0,0 +1,30 @@
+ARG NODE_VERSION=17.0.0
+
+FROM node:${NODE_VERSION}-slim as base
```
lassemand, line 3: A bit funny the slim can be used to build with. I guess that is node specific. :)

frontend/Jenkinsfile
lassemand: As per discussed in person we will keep it as is.

frontend/Dockerfile
schwartz-concordium: Made test locally and total build time of image was 200 s. using only slim and 180 second using non-slim for build.

Will use only slim for simplicity.

PR #165, opened 2024-01-02
Add node query to account cache lookup

schwartz-concordium:
## Purpose

This PR is preparation for step 2, **Migrate all token logic to contract flow, such that token events can be enriched with deserialised messages.**, which will follow PR https://github.com/Concordium/concordium-scan/pull/161.

It enables the account cache to be updated from the node. In theory since account is imported in the main import flow of the CCD scan and contracts in another import flow, it could be the contract flow needs to process information from an account, which hasn’t yet been imported to the database.

By enabling the account cache to lookup the account index on the node and not only the database, we ensure the account cache can be used in other flows than the one which imports account.

## Changes

- Add to account cache look up of account index on the node if it isn’t present in either memory or database

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/Application/Api/GraphQL/Import/AccountLookup.cs
+++ b/backend/Application/Api/GraphQL/Import/AccountLookup.cs
@@ -86,18 +86,24 @@ public class AccountLookup : IAccountLookup
         {
             try
             {
-                var response = _client.GetAccountInfoAsync(AccountAddress.From(nonExistingAccount), _blockHashInput).Result;
+                var response = _client.GetAccountInfoAsync(AccountAddress.From(nonExistingAccount), _blockHashInput)
+                    .Result;
                 var lookupResult = new LookupResult(nonExistingAccount, (long)response.AccountIndex.Index);
                 AddToCache(lookupResult.Key, lookupResult.Result);
                 result.Add(lookupResult);
             }
-            catch (RpcException e)
+            catch (AggregateException e)
             {
                 // Only catch errors due to account not found.
-                if (e.StatusCode != StatusCode.NotFound)
+                if (e.InnerException is not RpcException rpcException)
                 {
                     throw;
                 }
+                if (rpcException.StatusCode != StatusCode.NotFound)
+                {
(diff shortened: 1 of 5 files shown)


backend/Application/Api/GraphQL/Import/AccountLookup.cs
```diff
@@ -17,16 +21,38 @@ public class AccountLookup : IAccountLookup
...
     {
         _cache = cache;
         _dbSettings = dbSettings;
         _metrics = metrics;
+        _client = client;
         _logger = Log.ForContext(GetType());
     }
 
+    /// <summary>
+    /// The method first attempts to get account index using base address from cache,
```
rh-concordium, line 42: Very clear description . However, I would suggest starting with a one-liner that describes what the method does, before the next lines go into detail with _how_ it does it. E.g., "Gets account id's from [accountBaseAddresses]" (disclaimer: I don't know the correct markdown for C#. 

PS: A single sentence summary is recommended in Dart documentation, but I guess the principle is rather universal. https://dart.dev/effective-dart/documentation

backend/Application/Api/GraphQL/Import/AccountLookup.cs
```diff
@@ -52,10 +78,32 @@ public AccountLookup(IMemoryCache cache, DatabaseSettings dbSettings, IMetrics m
                 AddToCache(account.Key, account.Result);
                 result.Add(account);
             }
-
-            var nonExistingAccounts = notCached.Except(result.Select(x => x.Key));
-            foreach (var nonExistingAccount in nonExistingAccounts)
+        }
+        
+        var nonExistingAccounts = notCached.Except(result.Select(x => x.Key));
```
rh-concordium, line 83: Consider dividing the `GetAccountIdsFromBaseAddresses` method into sub-methods that correspond to each of the steps you described in the summary

rh-concordium (commented):
Just two quick comments from me since I noticed in Slack that Szymon will take a look

PR #174, opened 2024-01-24
Fix text overflow of token address and id and change to hexadecimal lowercase

schwartz-concordium:
## Purpose

Fix comments on CIS2 token view.
https://concordium.atlassian.net/browse/CBW-1653
https://concordium.atlassian.net/browse/CBW-1654

## Changes

- Truncate token address and token id when above 160px. Added tooltips such that the complete value can be seen.
- Change to render hexadecimal lowercase.

## Checklist

      hard-to-understand areas.


frontend/src/components/Tokens/TokenDetailsHeader.vue
```diff
@@ -6,7 +6,7 @@
 			<h3 class="w-full text-sm text-theme-faded">CIS2 Token</h3>
 			<div class="w-full flex items-center justify-items-stretch">
 				<h1 class="inline-block text-2xl numerical">
-					{{ tokenAddress }}
+					{{ tokenAddress.toLowerCase() }}
```
abizjak, line 9: I don't think this is right. Token address is case sensitive.

frontend/src/components/molecules/TokenLink.vue
```diff
@@ -2,12 +2,14 @@
...
 		<TokenIcon class="h-5 inline align-text-top" />
 		<LinkButton class="numerical px-2" @blur="emitBlur" @click="handleOnClick">
-			<div text-class="text-theme-body">
-				{{ props.tokenAddress }}
-			</div>
+			<Tooltip :text="tokenAddressLowerCase" text-class="text-theme-body">
+				<div class="token-address truncate">
+					{{ tokenAddressLowerCase }}
+				</div>
+			</Tooltip>
```
abizjak, line 9: I don't think token address should be lowercase. It is case sensitive (like account address). The token id, which is in base16/hex shoudl be lowercase though.

abizjak (approved):

PR #176, opened 2024-01-29
Fix token tooltip

schwartz-concordium:
## Purpose

Tooltips on token- address and id was pointing to the beginning of the `div` element instead of being positioned at the center.

Issue can be seen on testnet.

## Changes
- Change tooltips for token- address and id to be positioned using position absolute

https://github.com/Concordium/concordium-scan/assets/132270889/46585c70-7264-49c8-9a9d-18fb86f875eb

### DIFF
--- a/frontend/src/components/molecules/TokenId.vue
+++ b/frontend/src/components/molecules/TokenId.vue
@@ -1,6 +1,11 @@
 <template>
 	<div class="inline-block">
-		<Tooltip :text="tokenIdLowerCase">
+		<Tooltip
+			:text="tokenIdLowerCase"
+			x="50%"
+			y="0%"
+			tooltip-position="absolute"
+		>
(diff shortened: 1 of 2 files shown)


soerenbf (commented):
I don't get it.. when I go to testnet.ccdscan.io/tokens the tooltips are already aligned at the center of the text.. Has this already been deployed or what is happening here? 😄

schwartz-concordium:
> I don't get it.. when I go to testnet.ccdscan.io/tokens the tooltips are already aligned at the center of the text.. Has this already been deployed or what is happening here? 😄

Seems only to be an issue on DuckDuckGo browser and not Chrome or Firefox - and this “fix" actually breaks it in Chrome and Firefox.
Will close this PR since this solutions is clearly not correct.

PR #181, opened 2024-01-30
Change calculation of finalization time to use latest block hash info from block info

schwartz-concordium:
## Purpose

Currently finalization summaries are queried from the block but one can just look at the latest finalized block hash on the block info returned from get block info node query.

Changes are needed since finalization summaries isn't present in protocol 6. Using latest finalization block hash in
block info is a robust calculation in both old (before protocol 6) and new consensus protocol.

Issue: https://github.com/Concordium/concordium-scan/issues/78
Tasks
- https://concordium.atlassian.net/browse/CBW-1663
- https://concordium.atlassian.net/browse/CBW-1231

Prior PR
- https://github.com/Concordium/concordium-scan/pull/91

## Changes
- Update the finalization time to be set using the latest finalized block hash.
- There are two migration jobs:
   - One that updates the last finalized block height seen in ImportState, ensuring that the block processed by the import flow has the correct finalization time set.
   - Another that updates all prior finalization times that are currently set to null

## Checklist

      hard-to-understand areas.


backend/Application/Database/MigrationJobs/_03_UpdateFinalizationTimes.cs
```diff
@@ -0,0 +1,161 @@
...
+        await using var context = await _contextFactory.CreateDbContextAsync(token);
+
+        var firstBlockWithNull = await context
+            .Blocks
+            .Where(b => b.BlockStatistics.FinalizationTime == null)
+            .OrderBy(b => b.BlockHeight)
+            .Take(1)
+            .SingleAsync(cancellationToken: token);
+            
+        var priorBlockInfo = await _client.GetBlockInfoAsync(new Absolute((ulong)firstBlockWithNull.BlockHeight - 1), token);
```
td202, line 132: Do we need to worry about the block height being 0?

backend/Application/Database/MigrationJobs/_03_UpdateFinalizationTimes.cs
```diff
@@ -0,0 +1,161 @@
...
+        var priorBlockFinalization = await context.Blocks
+            .SingleAsync(x => x.BlockHash == priorBlockInfo.BlockLastFinalized.ToString(), cancellationToken: token);
```
td202, line 134: I am not sure why we are using this. `firstBlockWithNull` is the first block that isn't recorded as being finalized by another block, so the parent of that is recorded as being finalized by another block. We shouldn't need to consider the last finalized block with respect to that block, which is what we're doing here.

In particular, I think the loop will behave incorrectly on the first iteration with this definition, because the parent of the `firstBlockWithNull` will be changed to being finalized by the finalizer of `firstBlockWithNull`, though its finalization time is already recorded.

td202 (approved):
Looks good. I think there's only really a small issue of a few blocks being marked with the wrong finalization time.

backend/Application/Database/MigrationJobs/_03_UpdateFinalizationTimes.cs
schwartz-concordium: no since we know the data in the databases.

PR #192, opened 2024-02-18
Token amounts can be more than what decimal fits.

abizjak:
## Purpose

Token amounts are up to 256 bits so they don't fit into a decimal.

I believe this should be sufficient since all the actual updates to the amount are via SQL, and there the value that is added is always a BigInteger. So the value stored in the database is always an integer.

But ideally that would also be changed in the column type from "numeric" to "biginteger'.

The "decimal" makes it so that tokens cannot be read from the database and in particular the data migration job fails sometimes.
```
2024-02-17 19:30:10.575 [EROR] [Application.Jobs.JobsBackgroundService`2[[Application.Aggregates.Contract.Jobs.IContractJob, Application, Version=1.8.13.0, Culture=neutral, PublicKeyToken=null],[Application.Aggregates.Contract.Entities.ContractJob, Application, Version=1.8.13.0, Culture=neutral, PublicKeyToken=null]]] [00-e70217de8bab75aa8e02c36b2eb3746f-3725c7d03603a88f-00] _06_AddTokenAddress didn't succeed successfully due to exception.
System.OverflowException: Numeric value does not fit in a System.Decimal
```

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/Application/Aggregates/Contract/Entities/Token.cs
+++ b/backend/Application/Aggregates/Contract/Entities/Token.cs
@@ -1,4 +1,5 @@
 using System.Threading.Tasks;
+using System.Numerics;
 using Application.Api.GraphQL;
 using Application.Api.GraphQL.Accounts;
 using Application.Api.GraphQL.EfCore;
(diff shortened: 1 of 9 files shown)


abizjak:
I don't know which parts of the frontend need to be amended, if any. I don't have it set up to run locally.

schwartz-concordium (approved):

schwartz-concordium:
I will validate the frontend part before merge

PR #195, opened 2024-02-19
Improve token address migration

schwartz-concordium:
## Purpose

Avoid using entity framework in token address migration, since it generates timeouts.

### DIFF
--- a/backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
+++ b/backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
@@ -52,12 +52,12 @@ public sealed class _06_AddTokenAddress : IStatelessJob
             {
                 await using var context = await _contextFactory.CreateDbContextAsync(token);
                 var connection = context.Database.GetDbConnection();
-                var tokens = await context.Tokens
+                var tokens = context.Tokens
                     .AsNoTracking()
                     .Where(t => (int)t.ContractIndex == identifier && t.TokenAddress == null)
-                    .ToListAsync(token);
-                
-                foreach (var cisToken in tokens)
+                    .AsAsyncEnumerable();
+
+                await foreach (var cisToken in tokens)
                 {
                     var tokenAddress = Token.EncodeTokenAddress(cisToken.ContractIndex, cisToken.ContractSubIndex, cisToken.TokenId);
                     await connection.ExecuteAsync($"update graphql_tokens set token_address = '{tokenAddress}' where contract_index = {cisToken.ContractIndex} and contract_sub_index = {cisToken.ContractSubIndex} and token_id = '{cisToken.TokenId}';");
--- a/backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
+++ b/backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
@@ -1,4 +1,3 @@
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Application.Aggregates.Contract.Configurations;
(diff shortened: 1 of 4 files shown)


backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
```diff
@@ -50,18 +51,20 @@ await Policies.GetTransientPolicy(GetUniqueIdentifier(), _logger, _options.Retry
...
             {
                 await using var context = await _contextFactory.CreateDbContextAsync(token);
+                var connection = context.Database.GetDbConnection();
                 var tokens = await context.Tokens
-                    .Where(t => (int)t.ContractIndex == identifier)
+                    .AsNoTracking()
+                    .Where(t => (int)t.ContractIndex == identifier && t.TokenAddress == null)
                     .ToListAsync(token);
+                
                 foreach (var cisToken in tokens.Where(cisToken => cisToken.TokenAddress == null))
```
limemloh, line 60: Is the Where statement needed here? 
From my understanding, it is already checked when fetching above this line

backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
```diff
@@ -50,18 +51,20 @@ await Policies.GetTransientPolicy(GetUniqueIdentifier(), _logger, _options.Retry
...
                 await using var context = await _contextFactory.CreateDbContextAsync(token);
+                var connection = context.Database.GetDbConnection();
                 var tokens = await context.Tokens
-                    .Where(t => (int)t.ContractIndex == identifier)
+                    .AsNoTracking()
+                    .Where(t => (int)t.ContractIndex == identifier && t.TokenAddress == null)
                     .ToListAsync(token);
-                foreach (var cisToken in tokens.Where(cisToken => cisToken.TokenAddress == null))
+                
+                foreach (var cisToken in tokens)
```
limemloh, line 60: Since this is iterating synchronously, won't that mean the entire list of tokens will need to be loaded into memory before it can actually start iterating?
There seems to be an async enumerable, which I assume will stream the rows https://learn.microsoft.com/en-us/dotnet/api/system.data.entity.dbset?view=entity-framework-6.2.0#explicit-interface-implementations

schwartz-concordium: Tried it out but can’t then reuse the database connection when doing the update, since the async call is still active.

Even though there is a lot of tokens I think it is ok once to do the in memory load.

backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
```diff
@@ -50,18 +51,20 @@ await Policies.GetTransientPolicy(GetUniqueIdentifier(), _logger, _options.Retry
             .ExecuteAsync(async () =>
             {
                 await using var context = await _contextFactory.CreateDbContextAsync(token);
+                var connection = context.Database.GetDbConnection();
                 var tokens = await context.Tokens
-                    .Where(t => (int)t.ContractIndex == identifier)
+                    .AsNoTracking()
+                    .Where(t => (int)t.ContractIndex == identifier && t.TokenAddress == null)
```
abizjak, line 57: Why "int" and not "long"?

abizjak (approved):

backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
schwartz-concordium: The identifier created from `Enumerable.Range` is of type `int`.

Issue #194, opened 2024-02-18
/contracts endpoint uses excessive resources


abizjak:
**Bug Description**

When navigating to the `/contracts` endpoint on mainnet the frontend displays an empty list, and the reason for that is that the the database query fails with an out of resource error

The error trace is attached, but it is not much more informative other than showing the query that is made and the out of memory exception.

[contracts-error.txt](https://github.com/Concordium/concordium-scan/files/14324285/contracts-error.txt)

PR #196, opened 2024-02-19
Contract snapshot

schwartz-concordium:
## Purpose

All contract interactions is stored as events and currently no aggregation is stored. Due to this every time balance or current linked module needs to be found, all events needs to be loaded into memory.

Due to some other memory issues, on mainnet the CCD Scan query pod has now got both a CPU- and memory limit. Because of this we are suddently not able to see contract overview page, see issue https://github.com/Concordium/concordium-scan/issues/194.

This is because all events for all contracts needs to be loaded just to deliver this one view.

This PR adds a snapshot of contract state.

## Changes
- Add contract snapshot to import flow
- Add contract snapshot migration
- Update contract graphql queries
- Update frontend

## Checklist

      hard-to-understand areas.


backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
```diff
@@ -0,0 +1,198 @@
...
+    internal static async Task ImportContractSnapshot(
+        IContractRepositoryFactory repositoryFactory,
+        IContractRepository repository, 
+        ImportSource source
+        )
+    {
+        await using var moduleReadonlyRepository = await repositoryFactory.CreateModuleReadonlyRepository();
+            
+        var contractEventsAdded = repository
+            .GetContractEventsAddedInTransaction()
```
limemloh, line 60: Is there some hidden context here? added in which transaction?
How big will this list be?

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
```diff
@@ -0,0 +1,198 @@
...
+            .ToList();
+        
+        var addressesInitialized = contractEventsAdded
+            .Where(ce => ce.Event is ContractInitialized)
+            .Select(ce => (ce.ContractAddressIndex, ce.ContractAddressSubIndex))
+            .ToHashSet();
+
+        var addressesUpdated = contractEventsAdded
+            .Where(ce => !addressesInitialized.Contains((ce.ContractAddressIndex, ce.ContractAddressSubIndex)))
```
limemloh, line 69: Do we only care about update events for contracts that were just initialized? Maybe it is just me, who don't really understands the context this is running in.

limemloh, line 69: Also, this will include the init events as well right?

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
```diff
@@ -0,0 +1,198 @@
...
+            
+            var blockHeight = contractEvents.Last().BlockHeight;
+            
+            var moduleReferenceEvent = GetLatestModuleReferenceContractLinkEvent(address, moduleReadonlyRepository)!;
+
+            var contractSnapshot = new ContractSnapshot(
+                blockHeight,
+                address,
+                contractInitializationEvent.GetName(),
+                moduleReferenceEvent.ModuleReference,
```
limemloh, line 143: The `contractInitializationEvent` should contain a module reference, not sure what `GetLatestModuleReferenceContractLinkEvent` does in this case

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
```diff
@@ -0,0 +1,198 @@
...
+        ContractAddress contractAddress,
+        double initialAmount)
+    {
+        var amount = initialAmount;
+        foreach (var contractEvent in contractEvents)
+        {
+            switch (contractEvent.Event)
+            {
+                case ContractInitialized contractInitialized:
+                    amount += (long)contractInitialized.Amount;
```
limemloh, line 170: **Issue:** In this function we are casting from unsigned (`ulong`) to signed integer (`long`) and then to implicitly from `long` to `double`, all of these conversions can cause data loss.

backend/Application/Aggregates/Contract/Jobs/_06_AddTokenAddress.cs
```diff
@@ -50,18 +51,33 @@ await Policies.GetTransientPolicy(GetUniqueIdentifier(), _logger, _options.Retry
...
                 }
-                await context.SaveChangesAsync(token);
             });
         _logger.Debug($"Completed successfully processing {identifier}");
     }
-
+    
     public bool ShouldNodeImportAwait() => true;
 }
+// UPDATE your_table
```
limemloh, line 71: Was this comment meant to be included?

backend/DatabaseScripts/SqlScripts/0052_create_table_graphql_contract_snapshots.sql
```diff
@@ -0,0 +1,19 @@
+/*
+ Table used for aggregates state of contracts.
+ */
+create table graphql_contract_snapshot (
+    block_height                bigint        not null,
+    contract_address_index      bigint        not null,
+    contract_address_subindex   bigint        not null,
+    contract_name               text          not null,
+    module_reference            text          not null,
+    amount                      numeric       not null,
```
limemloh, line 10: This probably needs to be a `bigint`

limemloh (changes requested):

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
schwartz-concordium: We are currently using entity framework which keeps track of updates entities in the current transaction. In this case there is a transaction for each block height.

The list is already in memory, and will be scoped to data from a single block.

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
schwartz-concordium: I’m not sure I follow your question. This comment is linked to logic, to contract which doesn’t have a initialization event in the scope of contract events given.

Could you try clarify more, what your concern is?

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
schwartz-concordium: The contractInitializationEvent only contains a link to the first module which was linked, but not necessarily the latest one. Hence we need to take the latest module link event.

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
schwartz-concordium: This only filter out contracts. The events are later filtered conditional on the address given.

Hence in those cases, where `addressesUpdated` there will not be any contract initialization events.

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
```diff
@@ -0,0 +1,198 @@
...
+/// Contains aggregated local for <see cref="Contract"/>.
+/// </summary>
+public sealed class ContractSnapshot
+{
+    public ulong BlockHeight { get; init; }
+    public ulong ContractAddressIndex { get; init; }
+    public ulong ContractAddressSubIndex { get; init; }
+    public string ContractName { get; init; }
+    public string ModuleReference { get; init; }
+    public double Amount { get; init; }
```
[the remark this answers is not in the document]
schwartz-concordium: Changed to ulong. We are currently not using microCCD in the CCD scan and don’t want to introduce in this fix.

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
limemloh: Oh, I missed the `!` in the front.

limemloh (approved):

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
limemloh: I misunderstood the above list, so my comments is not relevant afterall.

backend/Application/Aggregates/Contract/Entities/ContractSnapshot.cs
limemloh: I am trying to figure out when `GetLatestModuleReferenceContractLinkEvent` is null, since we don't fall back to the initial module in this case as we do in the update case

schwartz-concordium: In the case where initialization has happened GetLatestModuleReferenceContractLinkEvent should never be null and above will give an exception. We would be failing fast and would be able to debug if this scenario happened before any data corruption occurred.

PR #201, opened 2024-03-11
Export fix

shjortConcordium:
## Purpose

Avoid big accounts to timeout (at 30 seconds) when using the export. 

## Changes

- Add index for account statements on account_id and time.
- Change query to stream data, instead of building a big string.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/Application/Api/Rest/ExportController.cs
+++ b/backend/Application/Api/Rest/ExportController.cs
@@ -45,7 +45,7 @@ public class ExportController : ControllerBase
 
         var query = dbContext.AccountStatementEntries
             .AsNoTracking()
-            .OrderByDescending(x => x.Index)
+            .OrderByDescending(x => x.Timestamp)
             .Where(x => x.AccountId == account.Id)
             ;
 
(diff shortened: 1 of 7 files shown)


backend/Application/Api/Rest/ExportController.cs
```diff
@@ -42,21 +45,24 @@ public async Task<ActionResult> GetStatementExport(string accountAddress, DateTi
 
         var query = dbContext.AccountStatementEntries
             .AsNoTracking()
-            .Where(x => x.AccountId == account.Id);
+            .OrderByDescending(x => x.Timestamp)
+            .Where(x => x.AccountId == account.Id)
+            ;
```
orhoj, line 50: Move to previous line.

backend/Application/Api/Rest/ExportController.cs
```diff
@@ -42,21 +45,24 @@ public async Task<ActionResult> GetStatementExport(string accountAddress, DateTi
 
         var query = dbContext.AccountStatementEntries
             .AsNoTracking()
-            .Where(x => x.AccountId == account.Id);
+            .OrderByDescending(x => x.Timestamp)
```
orhoj, line 48: Why do you have to order this now when it wasn't needed previously?

backend/Application/Api/Rest/ExportController.cs
```diff
@@ -66,37 +72,34 @@ public async Task<ActionResult> GetStatementExport(string accountAddress, DateTi
...
-            x.Amount,
-            x.AccountBalance
-        });
-        var values = await result.ToListAsync();
-        if (values.Count == 0)
-        {
-            return new NoContentResult();
-        }
+        var result = query.Select(x => string.Format("{0}, {1}, {2}, {3}\n", 
+        x.Timestamp.ToString("u"),
```
orhoj, line 76: Fix formatting.

orhoj (changes requested):

backend/Application/Api/Rest/ExportController.cs
shjortConcordium: Because otherwise it orders by index, and then it doesn't use the index I added.

orhoj (approved):

DOBEN (approved):

PR #214, opened 2024-09-23
Support Concordium Protocol Version 7

limemloh:
## Purpose

Prepare CCDScan for Concordium Protocol Version 7.

## Changes

- Fix CI due to `docker-compose` not being available in the ubuntu image anymore (see https://github.com/mariotoffia/FluentDocker/issues/312)
- Added concordium-net-sdk as submodule.
- Introduced some doc comments while learning the codebase.
- Fix bug in validation code, which branched on the node version instead of the protocol version of the block.
- Handle new events for configuring delegation and validators.
- Make stake changes immediate for protocol version 7 blocks and onwards, ignoring pending changes from this point on.

### DIFF
--- a/backend/Application/Api/GraphQL/Import/BakerImportHandler.cs
+++ b/backend/Application/Api/GraphQL/Import/BakerImportHandler.cs
@@ -25,11 +25,14 @@ public class BakerImportHandler
         _logger = Log.ForContext(GetType());
     }
 
+    /// <summary>
+    /// Process block data related to changes for bakers/validators.
+    /// </summary>
     public async Task<BakerUpdateResults> HandleBakerUpdates(BlockDataPayload payload, RewardsSummary rewardsSummary,
         ChainParametersState chainParameters, BlockImportPaydayStatus importPaydayStatus, ImportState importState)
     {
         using var counter = _metrics.MeasureDuration(nameof(BakerImportHandler), nameof(HandleBakerUpdates));
-        
+
         var changeStrategy = BakerChangeStrategyFactory.Create(payload.BlockInfo, chainParameters.Current, importPaydayStatus, _writer,
             payload.AccountInfos.BakersWithNewPendingChanges);;
 
@@ -164,8 +167,8 @@ public class BakerImportHandler
                 or BakerConfigured
                 or BakerKeysUpdated
             );
-        
-        await bakerChangeStrategy.UpdateBakersFromTransactionEvents(txEvents, importState, resultBuilder);
+
+        await bakerChangeStrategy.UpdateBakersFromTransactionEvents(txEvents, importState, resultBuilder, payload.BlockInfo.BlockSlotTime);
 
         // This should happen after the bakers from current block has been added to the database
         if (isFirstBlockAfterPayday)
@@ -224,10 +227,10 @@ public class BakerImportHandler
                         FinalizationCommission = source.PoolInfo.CommissionRates.FinalizationCommission.AsDecimal(),
                         BakingCommission = source.PoolInfo.CommissionRates.BakingCommission.AsDecimal()
                     },
-                    DelegatedStake = source.DelegatedCapital.Value,
+                    DelegatedStake = source.DelegatedCapital!.Value.Value,
                     DelegatorCount = 0,
-                    DelegatedStakeCap = source.DelegatedCapitalCap.Value,
-                    TotalStake = source.BakerEquityCapital.Value + source.DelegatedCapital.Value
+                    DelegatedStakeCap = source.DelegatedCapitalCap!.Value.Value,
+                    TotalStake = source.BakerEquityCapital!.Value.Value + source.DelegatedCapital.Value.Value
                 };
                 pool.ApplyPaydayStatus(source.CurrentPaydayStatus, source.PoolInfo.CommissionRates);
                 
@@ -262,8 +265,8 @@ public class BakerImportHandler
                 {
                     var rates = baker.ActiveState!.Pool!.CommissionRates;
                     rates.FinalizationCommission = AdjustValueToRange(rates.FinalizationCommission, currentFinalizationCommissionRange);
-                    rates.BakingCommission = AdjustValueToRange(rates.BakingCommission, currentBakingCommissionRange);
-                    rates.TransactionCommission = AdjustValueToRange(rates.TransactionCommission, currentTransactionCommissionRange);
+                    rates.BakingCommission = AdjustValueToRange(rates.BakingCommission, currentBakingCommissionRange!);
+                    rates.TransactionCommission = AdjustValueToRange(rates.TransactionCommission, currentTransactionCommissionRange!);
                 },
                 baker => baker.ActiveState!.Pool != null);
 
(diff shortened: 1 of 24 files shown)


backend/README.md
```diff
@@ -1,7 +1,16 @@
...
 
 # Getting Started
+
+This project depends on [concordium-net-sdk](https://github.com/Concordium/concordium-net-sdk) as a Git Submodule, which can be installed using:
+
+```
+git submodule update --init --recursive
+```
+
+Check out the readme of this dependency for how to build it.
```
DOBEN, line 12: Is there a way to automate the build of the dependencies when this backend is built?

backend/Tests/Api/GraphQL/__snapshots__/committed-schema.verified.graphql
```diff
@@ -2117,27 +2117,49 @@ enum AccountStatementEntryType {
...
 enum AccountTransactionType {
+  "Initialize a smart contract instance."
   INITIALIZE_SMART_CONTRACT_INSTANCE
+  "Update a smart contract instance."
   UPDATE_SMART_CONTRACT_INSTANCE
+  "Transfer CCD from an account to another."
   SIMPLE_TRANSFER
+  "Transfer encrypted amount."
   ENCRYPTED_TRANSFER
+  "Same as transfer but with a memo field."
```
DOBEN, line 2130: Just to be consistent with the other comments.
```suggestion
  "Same as transfer, but with a memo field."
```

backend/Tests/Api/GraphQL/__snapshots__/committed-schema.verified.graphql
```diff
@@ -2172,8 +2194,11 @@ enum ContractVersion {
   V1
 }
 
+"Enumeration of the types of credentials."
 enum CredentialDeploymentTransactionType {
+  "Initial credential is a credential that is submitted by the identity\nprovider on behalf of the user. There is only one initial credential\nper identity."
```
DOBEN, line 2199:
```suggestion
  "Initial credential is a credential that is submitted by the identity\nprovider on behalf of the user. There is atmost one initial credential\nper identity."
```

backend/Tests/Api/GraphQL/__snapshots__/committed-schema.verified.graphql
```diff
@@ -2172,8 +2194,11 @@ enum ContractVersion {
...
   INITIAL
+  "A normal credential is one where the identity behind it is only known to\nthe owner of the account, unless the anonymity revocation process was\nfollowed."
```
DOBEN, line 2201:
```suggestion
  "A normal credential is one where the identity behind it is only known to\nthe owner of the account, unless the identity disclosure process has been initiated."
```

DOBEN (approved):

backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs
```diff
@@ -16,45 +16,68 @@ public DelegationImportHandler(AccountWriter writer)
...
+                await _writer.UpdateAccountsWithPendingDelegationChange(firstBlockAfterPayday.PaydayTimestamp,
+                    account => ApplyPendingChange(account, resultBuilder));
+            } else if (payload.BlockInfo.ProtocolVersion == ProtocolVersion.P7) {
+                // Starting from Concordium Protocol Version 7 stake changes are immediate,
+                // meaning no delegators are expected to have pending changes from this point.
+                // Only the first reward day in P7 should this do anything, afterwards it is a
+                // no-op, since no accounts with pending changes are expected.
+                await _writer.UpdateAccountsWithPendingDelegationChange(DateTimeOffset.MaxValue,
+                    account => ApplyPendingChange(account, resultBuilder));
+            }
```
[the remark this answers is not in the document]
limemloh: It is only needed for migrating in the first reward day block of P7.
As stated in the comment, the P7 case is a no-op for every block except one, so for P8 and on nothing should actually happen here.
This is a quick solution, a more efficient implementation would be to set some flag saying the migration is already done.

backend/Application/Api/GraphQL/Import/BakerImportHandler.cs
```diff
@@ -224,10 +227,10 @@ await _writer.UpdateBakers(
                         FinalizationCommission = source.PoolInfo.CommissionRates.FinalizationCommission.AsDecimal(),
                         BakingCommission = source.PoolInfo.CommissionRates.BakingCommission.AsDecimal()
                     },
-                    DelegatedStake = source.DelegatedCapital.Value,
+                    DelegatedStake = source.DelegatedCapital!.Value.Value,
                     DelegatorCount = 0,
-                    DelegatedStakeCap = source.DelegatedCapitalCap.Value,
-                    TotalStake = source.BakerEquityCapital.Value + source.DelegatedCapital.Value
+                    DelegatedStakeCap = source.DelegatedCapitalCap!.Value.Value,
+                    TotalStake = source.BakerEquityCapital!.Value.Value + source.DelegatedCapital.Value.Value
```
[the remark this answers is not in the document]
limemloh: yes, this is due to how C# handles [nullable value types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types) (such as primitives).

backend/Application/Api/GraphQL/Import/Validations/AccountValidator.cs
```diff
@@ -193,16 +193,16 @@ private static string Format(PendingDelegationChange pendingChange)
         };
     }
 
-    private async Task ValidateBakers(List<AccountBaker> nodeAccountBakers, Block block, GraphQlDbContext dbContext, SingleAccountValidationInfo? singleAccountAddress)
+    private async Task ValidateBakers(ProtocolVersion protocolVersion, List<AccountBaker> nodeAccountBakers, Block block, GraphQlDbContext dbContext, SingleAccountValidationInfo? singleAccountAddress)
     {
+        _logger.Information($"Validating bakers in ${block.BlockHash}: ${nodeAccountBakers.Select(a=> a.BakerInfo.BakerId).ToString()}");
```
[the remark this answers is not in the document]
limemloh: Good catch

backend/README.md
limemloh: Actually, it is already automated, the developer just needs to have rust and cargo installed, but don't want to maintain that information here as well as in the .NET SDK.

backend/Tests/Api/GraphQL/__snapshots__/committed-schema.verified.graphql
limemloh: This is actually generated from an enum in .NET SDK, so I'll need to fix this as a PR for that.

backend/Tests/Api/GraphQL/__snapshots__/committed-schema.verified.graphql
limemloh: Same as above, these are generated from types in .NET SDK so will be fixed in a different PR.

backend/Tests/Api/GraphQL/__snapshots__/committed-schema.verified.graphql
limemloh: Same as above, these are generated from types in .NET SDK so will be fixed in a different PR.

DOBEN (approved):

PR #216, opened 2024-09-24
WIP Rust backend

limemloh:
## Purpose

Work in progress, everything might still change.


backend-rust/.env
```diff
@@ -0,0 +1,2 @@
+# Postgres
+DATABASE_URL=postgres://postgres:example@localhost/ccd-scan
```
Victor-N-Suadicani, line 2: We should use [the "offline" database checking that sqlx supports](https://github.com/launchbadge/sqlx/blob/main/sqlx-cli/README.md#enable-building-in-offline-mode-with-query) so that you don't need a database to build.

backend-rust/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "concordium-scan"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1"
```
Victor-N-Suadicani, line 9: I think we should specify full versions for dependencies, [see some argument for why here](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277).

backend-rust/README.md
```diff
@@ -0,0 +1,59 @@
...
+- PostgreSQL server 16
+
+## Setup for development
+
+To develop this service the following tools are required, besides the dependencies listed above:
+
+- [Rust and cargo](https://rustup.rs/)
+- [sqlx-cli](https://crates.io/crates/sqlx-cli)
+
+Then set the environment variable `DATABASE_URL` pointing to the location of the SQL database, this can be done by creating a `.env` file within this directory.
```
Victor-N-Suadicani, line 25: This could be avoided using sqlx's offline feature, but it will still be necessary when adding or editing queries.

backend-rust/rustfmt.toml
```diff
@@ -0,0 +1,15 @@
+edition = "2021"
```
Victor-N-Suadicani, line 1: I know we use this non-standard formatting in other projects and you have been unwilling to revert to the standard in those existing projects - but could we at least use the standard formatting for new projects?

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -0,0 +1,93 @@
...
+    info,
+};
+
+// TODO add env for remaining args.
+#[derive(Parser)]
+struct Cli {
+    /// The URL used for the database, something of the form
+    /// "postgres://postgres:example@localhost/ccd-scan"
+    #[arg(long, env = "DATABASE_URL")]
+    database_url: String,
```
Victor-N-Suadicani, line 23: Keep in mind that command-line arguments are not private (can be seen by other processes) so only using an env var for this might be preferable as this URL could contain a password.

backend-rust/src/bin/ccdscan-indexer.rs
```diff
@@ -0,0 +1,87 @@
...
+    info,
+};
+
+// TODO add env for remaining args.
+#[derive(Parser)]
+struct Cli {
+    /// The URL used for the database, something of the form
+    /// "postgres://postgres:example@localhost/ccd-scan"
+    #[arg(long, env = "DATABASE_URL")]
+    database_url: String,
```
Victor-N-Suadicani, line 25: Similarly here.

backend-rust/src/graphql_api.rs
```diff
@@ -0,0 +1,4222 @@
+//! TODO
```
Victor-N-Suadicani, line 1: This file is way too long. It should probably be split into smaller modules. (is it generated?)

backend-rust/src/indexer.rs
```diff
@@ -0,0 +1,744 @@
+//! TODO:
```
Victor-N-Suadicani, line 1: This file also defines many separate types. It'd be easier to get an overview if it was split up a bit more.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.80"
+components = [ "rustfmt" ]
```
Victor-N-Suadicani, line 3: I think you should rather just [set the minimum supported Rust version in the Cargo.toml](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) and remove this file.

backend-rust/README.md
limemloh: Sure, but contributors should not rely on that

backend-rust/Cargo.toml
limemloh: If we know the minimum supported version of these then sure, but I don't see the argument when include the `Cargo.lock` file.
This seems more relevant for libraries.

rust-toolchain.toml
limemloh: I disagree, if this was a library for others to consume then sure, but since this is an application I think it is good to pin the version like any other dependency.

backend-rust/src/graphql_api.rs
limemloh: I don't think the line count alone should dictate this, there needs to be something gained from it.
In any case, I would prefer such things were postponed until the end of the first working solution, since we don't really have the full picture yet.

backend-rust/src/indexer.rs
limemloh: I think scattering types across files is harder to get an overview of in general, especially if they are tied into some context across files.
But if there are some clear separation of them, it could make sense, but as mentioned above, it should probably be done in the end when changes are less likely

backend-rust/rustfmt.toml
limemloh: IMO this should just use the formatting like every other project we have (which is actually not this config, I'll fix that).
It is not because I prefer the config, and I was not the one who defined it either, but it would be consistent.
Also, if we change it for this project, people would have to change editor configs between this and other projects.

backend-rust/Cargo.toml
Victor-N-Suadicani: If you add dependencies later on, that could change the versions in the lock file, as far as I know. It is relevant for libraries but it is good practice in general.

backend-rust/rustfmt.toml
Victor-N-Suadicani: With that approach, the "cost" of changing to the standard format will only increase and it will only get more and more cumbersome to change the format to the standard. I think it would be better to do a gradual approach where we slowly move towards the standard, and part of that would be new projects using the standard.

backend-rust/src/graphql_api.rs
Victor-N-Suadicani: At this scale, I disagree. I think there is very rarely a reason to have a file of several thousand lines and the existence of such a file would have to be justified, rather than having to justify splitting it up.

backend-rust/src/indexer.rs
Victor-N-Suadicani: If they are tied together, they can still be related to each other via submodules. It doesn't have to be separate modules as siblings, it could be a parent-child relationship.

backend-rust/Cargo.toml
limemloh: I don't see why we should include a patch version unless we know this patch version is the minimum supported.
If we add a new dependency we want it to resolve the shared dependencies and I don't think cargo will bump unrelated dependencies.
In case it resolves to a version we don't support, we can then add a patch version with a comment to why this is the case.

backend-rust/src/graphql_api.rs
limemloh: Let's agree to disagree then.

backend-rust/src/indexer.rs
limemloh: If it makes sense, sure.

backend-rust/rustfmt.toml
limemloh: IMO It is already too cumbersome compared to the value gained from doing it.
Also, it needs to be a team decision to change it, and not something we just randomly do because of preference.

rust-toolchain.toml
limemloh: Never versions introduce new clippy warnings, and we don't want the CI to randomly start failing checks because a new stable release of rust.
We can remove this file, but in the end we will pin a rust version in the build job which is the rust version we care about.

backend-rust/Cargo.toml
Victor-N-Suadicani: >I don't see why we should include a patch version unless we know this patch version is the minimum supported.

But right now, if you say `anyhow = "1"` (just as an example), it seems like the application should work and build with anyhow version `1.0.0`, but that may actually not be the case, because we haven't tested that because cargo chose a later version and we've only tested with that version.

rust-toolchain.toml
Victor-N-Suadicani: Pinning in the build job is fine and makes sense - but if someone wants to build this with a newer compiler, I think they should be able to (even if they get some clippy warnings).

backend-rust/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+chrono = { version = "0.4", features = ["serde"] }
+clap = { version = "4.5", features = ["derive", "env"] }
+concordium-rust-sdk = { path = "./concordium-rust-sdk" }
+derive_more = "0.99"
+dotenv = "0.15"
+futures = "0.3"
+hex = "0.4"
+serde = "1.0"
+serde_json = "1.0"
+sqlx = { version = "0.7", features = ["runtime-tokio", "postgres", "chrono"] }
```
soerenbf, line 23: Reading random posts on the internet warns me that `tokio_postgres` is "faster" than `sqlx`.. We also use `tokio_postgres` everywhere else AFAIK. I'm not opposed to `sqlx` as such, I'm just wondering if it's the most optimal library to achieve the goal. Is this something you've considere?

backend-rust/rustfmt.toml
limemloh: I did not write random preference. I was referring to random changes, we need to align people before doing something like this.
"This situation" is not a problem. You just prefer something else, which is fine. It can be discussed whether we want to spend the time it takes to change this in the team, but should not be something we just start doing before it has been discussed.

> The whole point of sticking to the usual standard that is used in all other open source projects

Seems that a lot of open source projects have some rustfmt setup: servo and rust-analyzer even the rust compiler and rustfmt itself have one.

backend-rust/Cargo.toml
limemloh: Yes, the reason for using `sqlx` is the compile-time checked queries, since it saves us from having test cases catching bugs in the queries. It does come with a cost, especially compile time is up due to this, so it might not be worth it.
I have seen claims that `tokio_postgres` is faster and even 2x faster, but benchmarks can be misleading, so it is something we should compare ourselves. Right now the main concern is to reduce dev time.

rust-toolchain.toml
limemloh: You can always override using `cargo +stable`

backend-rust/Cargo.toml
Victor-N-Suadicani: I've personally used sqlx in the past and I've found it to be very nice. Never had any performance issues with it, even for quite heavy services, but that's just anecdotal obviously.

But the _interface_ of the crate is very nice and much simpler than tokio_postgres.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -0,0 +1,224 @@
...
+    index
+        BIGINT
+        NOT NULL,
+    -- Absolute height of the block containing the transaction.
+    block
+        BIGINT
+        REFERENCES blocks(height)
+        NOT NULL,
+    -- Transaction hash encoded using HEX.
+    hash
```
DOBEN, line 120: Might not be included in this initial PR. However, adding indexes beyond the primary key index would help to increase the performance in common queries.

E.g. an index for querying transactions/bocks based on the transaction_hash/block_hash.

DOBEN (commented):
Well done. I big piece of work and something to start with.

backend-rust/migrations/0001_initialize.up.sql
limemloh: Yes, I intend to do that, but I would like to have most of the implementation ready, before we start optimizing.
It would be good with some realistic numbers to understand the impact, although I agree that these cases will obviously benefit from it.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -0,0 +1,224 @@
...
+    -- Index of the account.
+    index
+        BIGINT
+        PRIMARY KEY
+        NOT NULL,
+    -- Account address bytes encoded using base58check.
+    address
+        CHAR(50)
+        UNIQUE
+        NOT NULL,
```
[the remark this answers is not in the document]
limemloh: Good point.
I initially wanted to put the byte representation in, but since the API provides search for an account address by some prefix, I ended up putting the base58check representation in.

backend-rust/Cargo.toml
limemloh: I imagine that most of the response time is spent by the database processing the query, so hopefully it does not matter too much and the compile time checked queries are speeding up development by removing the need for testing (to some degree at least). I don't expect a massive improvement by replacing .NET by Rust for the same reason.
We need to run with actual data to know for sure.

limemloh: Closing this for now, but feel free to take it up again

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -0,0 +1,224 @@
...
+    -- The account used for sending of the transaction.
+    -- NULL for chain update and account creation transactions.
+    -- Foreign key constraint added later, since account table is not defined yet.
+    sender
+        BIGINT,
+    -- The type of transaction.
+    type
+        transaction_type
+        NOT NULL,
+    -- NULL if the transaction type is not account or the account transaction have no effect on chain.
```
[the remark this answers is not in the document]
limemloh: Thinking about it, I am actually not so sure if this is true.
I think even rejected transactions have a transaction type.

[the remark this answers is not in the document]
soerenbf: That depends on why they were rejected, right? In the rust SDK the transaction type is optional for the transaction effects

backend-rust/src/indexer.rs
```diff
@@ -0,0 +1,779 @@
...
+    /// `false`
+    async fn on_failure(
+        &mut self,
+        endpoint: v2::Endpoint,
+        successive_failures: u64,
+        err: TraverseError,
+    ) -> bool {
+        info!("Failed preprocessing {} times in row: {}", successive_failures, err);
+        self.preprocessing_failures.get_or_create(&NodeMetricLabels::new(&endpoint)).inc();
+        true
```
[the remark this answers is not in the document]
limemloh: Right, good catch.
I temporarily put `true` here for catching issues when developing, but it should not be that in the general case.
The logic for rotating nodes when something fails should already be part of `TraverseConfig::traverse`.

backend-rust/src/indexer.rs
```diff
@@ -0,0 +1,779 @@
...
+            self.baker_id,
+            self.total_amount,
+            self.total_staked
+        )
+        .execute(tx.as_mut())
+            .await?;
+
+        // Check if this block knows of a new finalized block.
+        // If so, mark the blocks since last time as finalized by this block.
+        if self.block_last_finalized != context.last_finalized_hash {
```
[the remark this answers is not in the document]
limemloh: yes, in the old consensus it was very likely, but it is also possible in the new consensus, since finalization requires two executive blocks (i.e. no timeout certificates in between).
The indexer only processes blocks that are finalized, so no rollback is needed, but we need to compute the finalization time and this is the best that I could come up with.

backend-rust/src/indexer.rs
```diff
@@ -0,0 +1,779 @@
...
+        for item in self.prepared_block_items.iter() {
+            item.save(context, tx).await?;
+        }
+        Ok(())
+    }
+}
+
+struct PreparedBlockItem {
+    block_index:      i64,
+    tx_hash:          String,
```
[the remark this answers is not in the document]
limemloh: Since we have to convert it into a string before we can insert, I would like to do it as part of preprocessing.
If we are to make this more type safe then we should make newtype wrappers for every type, which could make sense as long as sqlx still can do compile-time validation of queries.

backend-rust/migrations/0001_initialize.up.sql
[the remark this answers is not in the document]
limemloh: Yes, you are right. If the account transaction failed to be deserialized.

https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/types/struct.AccountTransactionDetails.html#method.transaction_type

.gitignore
```diff
@@ -13,3 +13,5 @@ pgbackrest.conf
 target/
 
 .vscode/
+
```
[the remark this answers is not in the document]
limemloh: Makes sense, I'll add it

backend-rust/src/bin/ccdscan-api.rs
limemloh: Good point, it is convenient through, so I'll add a warning in the documentation for now.

backend-rust/src/graphql_api.rs
```diff
@@ -0,0 +1,4341 @@
...
+        }
+        if last.is_some() {
+            if let Some(edge) = connection.edges.last() {
+                connection.has_previous_page = edge.node.height != 0;
+            }
+        } else {
+            if let Some(edge) = connection.edges.first() {
+                connection.has_previous_page = edge.node.height != 0;
+            }
+        }
```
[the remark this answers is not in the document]
limemloh: I have added a limit which can be configured, global limit was probably a simpler, but I'll keep it as is.

.gitignore
```diff
@@ -13,3 +13,5 @@ pgbackrest.conf
...
+.env
```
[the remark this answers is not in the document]
limemloh: Yes we should not forget to do that, but I believe when we get closer to a release that we will split it into two crates, so we can track version and changes separately, so I'll not add it just yet.

DOBEN (approved):
Well done.  Since this PR already includes updates to the front-end that partially request the data from the new Rust backend database, it might be best to keep the PR open as a feature branch (not adding code to this PR so). We can merge the coming PRs into this PR until we're ready to roll out the new backend.

backend-rust/src/indexer.rs
```diff
@@ -0,0 +1,1342 @@
+#![allow(unused_variables)] // TODO Remove before first release
+#![allow(dead_code)] // TODO Remove before first release
+
```
Victor-N-Suadicani, line 3: I want to point out another thing with regards to these very large files. Since everything here is in the same module, everything is public to everything, so there is no encapsulation possible. Generally speaking, encapsulation is good and helps other good principles like separation of concerns and single responsibility.

Even if these principles are followed right now despite the single-module structure, the code is brittle to problems with these principles in the future. Separating the concerns would better protect against "code rot" or "spaghettification" as this code is maintained in the long term.

Splitting into multiple modules is therefore not just a question of style, but also of code cleanliness and maintainability.

limemloh: On a high level, this all sounds good, and I would not be against splitting up the file. I don't think the number of lines alone should be the argument for doing it, though, and it should be driven by an actual separation of concern.
I would prefer to have a big chunk of the implementation ready, instead of attempting to abstract ahead, then we can have actual concrete suggestions and avoid talking abstract principles.

Victor-N-Suadicani: For some concrete examples, it could make sense to separate each query into a seperate module with its associated types, like the `BlockMetrics` and `BlockMetricsBuckets` types and the `block_metrics` method. Probably you could similarly split other queries out to modules. I think it would be most sensible/easiest to do this gradually (i.e. now, not as a "big bang" change later) as we build the new solution.

limemloh: Let's move this discussion offline instead

PR #217, opened 2024-09-25
Compute APY using the expected payday duration instead of last duration

limemloh:
## Purpose

Fix issue where backend is crashing when protocol update is effective within the same second as the next payday.
This results in a payday duration rounded to `0` seconds, which is an issue for the current computation of APY.

This PR changes the APY computation to be based on the expected payday length instead of the last measured payday length.

Also updated the changelog and prepared for release 1.9.1


DOBEN (approved):
lgtm

PR #269, opened 2024-10-15
Add contract and module events

DOBEN:
## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+            | Event::ContractResumed(_)
+            | Event::ContractUpgraded(_)
+            | Event::ContractUpdated(_) = event
+            {
+                Ok(())
+            } else {
+                Err(ApiError::InternalError(format!(
+                    "Not Transferred, ContractInterrupted, ContractResumed, ContractUpgraded, or \
+                     ContractUpdated event; Wrong event enum tag: {:?}",
+                    mem::discriminant(&event)
```
Victor-N-Suadicani, line 1446: Same here, use the `Debug` representation.

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+                block_slot_time: row.block_slot_time,
+            };
+            contract_events.push(initial_event);
+            total_events_count += 1;
+        }
+
+        Ok(ContractEventsCollectionSegment {
+            page_info:   CollectionSegmentInfo {
+                has_next_page,
+                has_previous_page: skip > 0,
```
Victor-N-Suadicani, line 1532: Unrelated comment but isn't it weird this is included? The frontend could surely figure that out by itself?

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+            },
+            items:       Some(contract_events),
```
Victor-N-Suadicani, line 1534: Seems like this field shouldn't be an `Option` as it is never not `Some` as far as I can tell.

backend-rust/src/graphql_api.rs
```diff
@@ -3346,6 +3559,56 @@ impl SearchResult {
     }
 }
 
+fn decode_value_with_schema(
+    opt_schema: &Option<Type>,
+    schema_name: &str,
+    value: &Vec<u8>,
```
Victor-N-Suadicani, line 3565:
```suggestion
    value: &[u8],
```

backend-rust/src/graphql_api.rs
```diff
@@ -3346,6 +3559,56 @@ impl SearchResult {
     }
 }
 
+fn decode_value_with_schema(
+    opt_schema: &Option<Type>,
```
Victor-N-Suadicani, line 3563:
```suggestion
    opt_schema: Option<&Type>,
```
[See e.g. here.](https://www.reddit.com/r/rust/comments/wqsxk2/is_it_better_to_pass_optiont_or_optiont/ikokviw/)

backend-rust/src/graphql_api.rs
```diff
@@ -3346,6 +3559,56 @@ impl SearchResult {
...
+    schema_name: &str,
+    value: &Vec<u8>,
+    value_name: &str,
+) -> String {
+    match opt_schema {
```
Victor-N-Suadicani, line 3568: Make this `match` into
```rust
let Some(schema) = opt_schema else {
    return format!(
        "Failed to deserialize {} with {} schema: {:?}",
        value_name,
        schema_name,
        e.display(true)
    )
};
```
Then you avoid indenting the whole rest of the body and the return is not so far from the relevant condition.

Alternatively if this function doesn't work for `None` anyway, maybe just don't take an option but take a `Type` directly? Let the caller decide how to handle it if it is missing.

backend-rust/src/graphql_api.rs
```diff
@@ -3346,6 +3559,56 @@ impl SearchResult {
...
+        Some(schema) => {
+            let mut cursor = Cursor::new(&value);
+            match schema.to_json(&mut cursor) {
+                Ok(v) => {
+                    to_string(&v).unwrap_or_else(|e| {
```
Victor-N-Suadicani, line 3573: I would not import `serde_json::to_string` like this. It's not clear what kind of string you're turning it into. Use the fully qualified name, then it's obvious you're turning it into JSON.
```suggestion
                    serde_json::to_string(&v).unwrap_or_else(|e| {
```

backend-rust/src/graphql_api.rs
```diff
@@ -3346,6 +3559,56 @@ impl SearchResult {
...
+                        "Failed to deserialize {} with {} schema: {:?}",
+                        value_name,
+                        schema_name,
+                        e.display(true)
+                    )
+                }
+            }
+        }
+        // Note: Shall we use something better than this string if no schema is
+        // available for decoding.
```
Victor-N-Suadicani, line 3605: Maybe we could rather just return a `null` or some kind of missing value to the frontend and the frontend can display something more intelligent than just a string error message. I.e. it could display something graphical. But we're not looking to change the frontend yet so it's maybe okay to stick with this for now?

backend-rust/src/graphql_api.rs
```diff
@@ -4319,20 +4583,94 @@ impl BakerStakeIncreased {
...
+        });
+
+        // Nice-to-have: pagination info but not used at front-end currently.
+
+        Ok(connection)
+    }
+
+    async fn events<'a>(
+        &self,
+        ctx: &Context<'a>,
```
Victor-N-Suadicani, line 4632:
```suggestion
    async fn events(
        &self,
        ctx: &Context<'_>,
```

backend-rust/src/graphql_api.rs
```diff
@@ -4319,20 +4583,94 @@ impl BakerStakeIncreased {
...
+    ) -> ApiResult<connection::Connection<String, String>> {
+        let pool = get_pool(ctx)?;
+
+        let row = sqlx::query!(
+            r#"
+SELECT
```
Victor-N-Suadicani, line 4638: As in the other places:
```suggestion
        let row = sqlx::query!("
            SELECT
```

backend-rust/src/graphql_api.rs
```diff
@@ -4319,20 +4583,94 @@ impl BakerStakeIncreased {
...
+  contracts.module_reference as module_reference,
+  name as contract_name,
+  schema as display_schema
+FROM contracts
+JOIN smart_contract_modules ON smart_contract_modules.module_reference=contracts.module_reference
+WHERE index=$1 AND sub_index=$2
```
Victor-N-Suadicani, line 4644:
```suggestion
JOIN smart_contract_modules ON smart_contract_modules.module_reference = contracts.module_reference
WHERE index = $1 AND sub_index = $2
```

backend-rust/src/graphql_api.rs
```diff
@@ -4351,6 +4689,26 @@ impl From<concordium_rust_sdk::types::smart_contracts::WasmVersion> for Contract
     }
 }
 
+#[derive(Debug, thiserror::Error, Clone)]
+#[error("Invalid contract version: {value}")]
+pub struct InvalidContractVersionError {
+    value: i32,
+}
```
Victor-N-Suadicani, line 4696:
```suggestion
#[error("Invalid contract version: {.0}")]
pub struct InvalidContractVersionError(i32);
```

backend-rust/src/graphql_api.rs
```diff
@@ -4543,22 +4901,116 @@ pub struct ContractResumed {
...
+    // All logged events by the smart contract during the transaction execution.
+    contract_logs_raw: Vec<Vec<u8>>,
+    input_parameter:   Vec<u8>,
+}
+
+#[ComplexObject]
+impl ContractUpdated {
+    async fn message_as_hex(&self) -> ApiResult<String> { Ok(hex::encode(&self.input_parameter)) }
+
+    async fn message<'a>(&self, ctx: &Context<'a>) -> ApiResult<String> {
```
Victor-N-Suadicani, line 4920:
```suggestion
    async fn message(&self, ctx: &Context<'_>) -> ApiResult<String> {
```

backend-rust/src/graphql_api.rs
```diff
@@ -4543,22 +4901,116 @@ pub struct ContractResumed {
...
+
+        let row = sqlx::query!(
+            r#"
+SELECT
+  contracts.module_reference as module_reference,
+  name as contract_name,
+  schema as display_schema
+FROM contracts
+JOIN smart_contract_modules ON smart_contract_modules.module_reference=contracts.module_reference
+WHERE index=$1 AND sub_index=$2
```
Victor-N-Suadicani, line 4931: As in the other comments

backend-rust/src/graphql_api.rs
```diff
@@ -4543,22 +4901,116 @@ pub struct ContractResumed {
...
+        Ok(connection)
+    }
+
+    async fn events<'a>(
+        &self,
+        ctx: &Context<'a>,
+    ) -> ApiResult<connection::Connection<String, String>> {
+        let pool = get_pool(ctx)?;
+
+        let row = sqlx::query!(
```
Victor-N-Suadicani, line 4976: Here too

backend-rust/src/indexer.rs
```diff
@@ -973,13 +972,24 @@ VALUES
...
+/// status of the event (if the event was in a successful or rejected
+/// transaction).
+struct PreparedEventData {
+    /// The prepared event. Note: this is optional temporarily until we can
+    /// handle all events.
+    event:   Option<PreparedEvent>,
+    /// The status of the event (if the event belongs to a successful or
+    /// rejected transaction).
+    success: bool,
+}
```
Victor-N-Suadicani, line 991: This data structure seems very strange to me.

Why not use `Result<PreparedEvent, E>` with some appropriate `E` instead of this separate `success` field? Can you have `event: Some(e)` at the same time as `success: false`? What about `event: None` and `success: true`? Maybe the `success` field should be a field on the prepared event instead of in this struct?

backend-rust/src/indexer.rs
```diff
@@ -1142,23 +1167,57 @@ impl PreparedEvent {
...
                     event.save(tx).await?;
                 }
                 Ok(())
             }
-            PreparedEvent::ModuleDeployed(event) => event.save(tx).await,
-            PreparedEvent::ContractInitialized(event) => event.save(tx).await,
+            PreparedEvent::ModuleDeployed(event) =>
+            // Only save the module into the `modules` table if the transaction was
+            // successful.
+            {
```
Victor-N-Suadicani, line 1196:
```suggestion
            PreparedEvent::ModuleDeployed(event) => {
                // Only save the module into the `modules` table if the transaction was
                // successful.
```

backend-rust/src/indexer.rs
```diff
@@ -1142,23 +1167,57 @@ impl PreparedEvent {
...
+                if self.success {
+                    event.save(tx).await
+                } else {
+                    Ok(())
+                }
+            }
+            PreparedEvent::ContractInitialized(event) =>
+            // Only save the contract into the `contracts` table if the transaction was
+            // successful.
+            {
```
Victor-N-Suadicani, line 1206:
```suggestion
            PreparedEvent::ContractInitialized(event) => {
                // Only save the contract into the `contracts` table if the transaction was
                // successful.
```

Victor-N-Suadicani (approved):
Various stylistic comments and suggestions, nothing blocking.

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+                                JOIN transactions
+                                ON contract_events.block_height = transactions.block_height
+                                AND contract_events.transaction_index = transactions.index
+                                JOIN accounts
+                                ON transactions.sender = accounts.index
+                                JOIN blocks ON contract_events.block_height = blocks.height
+                                WHERE contract_events.contract_index = $1
+                                AND contract_events.contract_sub_index <= $2
+                                LIMIT $3
+                                OFFSET $4
```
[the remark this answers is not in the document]
Victor-N-Suadicani: Yes, that would be a way to solve the inefficiency. Considering the sequential way we insert stuff in the indexer, this is honestly probably the best solution, since it's quite easy to add that index column.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -309,11 +310,47 @@ CREATE TABLE contracts(
...
 
     -- Make the contract index and subindex the primary key.
     PRIMARY KEY (index, sub_index)
 );
 
+-- Every successful event associated to a contract.
+CREATE TABLE contract_events (
+    -- An index/id for this event (row number).
+    index
+        BIGINT GENERATED ALWAYS AS IDENTITY
```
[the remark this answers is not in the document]
DOBEN: Didn't we discuss that it would be nice if these indexes were autogenerated and sequentially added (without skipping numbers)?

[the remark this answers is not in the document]
Victor-N-Suadicani: My understanding from [that discussion](https://github.com/Concordium/concordium-scan/pull/268#discussion_r1808299392) was that we concluded that using plain integers was fine since the insertion is sequential. But it's not hugely important.

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+    init_block_height as block_height,
+    transactions.events,
+    transactions.hash as transaction_hash,
+    accounts.address as creator,
+    version
+FROM contracts
+JOIN blocks ON init_block_height=blocks.height
+JOIN transactions ON init_block_height=transactions.block_height AND init_transaction_index=transactions.index
+JOIN accounts ON transactions.sender=accounts.index
+WHERE contracts.index=$1 AND contracts.sub_index=$2
```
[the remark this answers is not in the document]
DOBEN: I think that is a preference. To keep consistent with the rest of the queries, I keep it as it is.

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+            }
+
+            // Contract init transactions have exactly one top-level event.
+            let event = events.swap_remove(0_usize);
+
+            match event {
+                Event::ContractInitialized(_) => Ok(()),
+                _ => Err(ApiError::InternalError(format!(
+                    "Not ContractInitialized event; Wrong event enum tag: {:?}",
+                    mem::discriminant(&event)
```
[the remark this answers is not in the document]
DOBEN: The `Event` type does not implement `Debug` and since this error should never happen (also long we didn't corrupt the database), the enum_tag is mainly for internal debugging.

But long-term would be nice if somebody implements `Debug` on the `Event` type :).

backend-rust/src/graphql_api.rs
DOBEN: true, when using `skip/take` this could be also easily derived at the front-end. But sticking to the implementation/types that are currently used which have this pagination information.

backend-rust/src/indexer.rs
DOBEN: The type was removed now, and we will introduce an enum down the line to represent events from failed transactions.

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+            .await?
+            .ok_or(ApiError::NotFound)?;
+
+            let Some(events) = row.events else {
+                return Err(ApiError::InternalError("Missing events in database".to_string()));
+            };
+
+            let mut events: Vec<Event> = serde_json::from_value(events).map_err(|_| {
+                ApiError::InternalError("Failed to deserialize events from database".to_string())
+            })?;
```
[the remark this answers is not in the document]
DOBEN: I left it for now since we later need to cast the array back to a single event.

backend-rust/src/graphql_api.rs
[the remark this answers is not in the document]
DOBEN: There are dependent types that don't implement it so will leave it as it is for now.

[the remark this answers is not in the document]
Victor-N-Suadicani: I would just derive Debug on those too. It does no harm and should just be done.

backend-rust/src/graphql_api.rs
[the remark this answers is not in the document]
Victor-N-Suadicani: Not sure I understand - yes you only need the single event, that's why I'm saying you could bind as `[event]` so you don't need to pop. I.e.

```rust
let [event]: [Event; 1] = serde_json::from_value(events).map_err(|_| {
    ApiError::InternalError("Failed to deserialize events from database".to_string())
})?;
```
Due to the binding, you are "unpacking" the array immediately and the `event` variable only refers to the single event.

backend-rust/src/graphql_api.rs
DOBEN: I kept the current explicit lifetime since that is consistent with the rest of the code.

backend-rust/src/graphql_api.rs
DOBEN: I leave the lifetime name explicitly since it is consistent with the rest of the code currently.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -169,6 +169,7 @@ CREATE TABLE transactions(
         BOOLEAN
         NOT NULL,
     -- Transaction details. Events if success is true.
+    -- TODO: It would be nice to have every event in a separate row in a table to easily query a specific event from a specific transaction.
     events
         JSONB,
```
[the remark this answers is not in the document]
DOBEN: Removed the comment.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -309,11 +310,47 @@ CREATE TABLE contracts(
...
+        NOT NULL,
 
     -- Make the contract index and subindex the primary key.
     PRIMARY KEY (index, sub_index)
 );
 
+-- Every successful event associated to a contract.
+CREATE TABLE contract_events (
+    -- An index/id for this event (row number).
+    index
```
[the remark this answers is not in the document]
DOBEN: removed now.

backend-rust/src/graphql_api.rs
```diff
@@ -1313,14 +1331,209 @@ struct Contract {
...
+        let row_stream = sqlx::query!(
+            r#"
+                            SELECT * FROM (
+                                SELECT
+                                    contract_events.index,
+                                    contract_events.transaction_index,
+                                    trace_element_index,
+                                    contract_events.block_height AS event_block_height,
+                                    transactions.hash as transaction_hash,
+                                    transactions.events,
```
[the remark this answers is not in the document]
DOBEN: What error would appear, if `events` is `NULL` which by the database schema could happen?
Currently we check these edge cases and race an internal error should it ever happen.

backend-rust/src/graphql_api.rs
Victor-N-Suadicani: This is still the case, no?

Victor-N-Suadicani (approved):

backend-rust/src/graphql_api.rs
DOBEN: fixed

backend-rust/migrations/0001_initialize.up.sql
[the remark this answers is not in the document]
DOBEN: we not query the index, but it is used as a `primary key` to insure each event can be inserted into the database uniquely.

PR #272, opened 2024-10-21
Batch insert blocks

limemloh:
## Purpose

Reduce indexer catch-up time.

Benchmarked by running the indexer for exactly 10 minutes on a local set up, (i.e. indexer, database and concordium node (testnet) all running on the same machine) resetting the database in between:
- Without batching it managed to index to block height `86811`.
- With block batching it managed to index to block height `124171`.

## Changes

- When processing a batch, collect all the blocks and insert them as one query, reducing the chattiness with the database.
- Spawn separate tasks for preprocessing and processing, allowing for these to be done in parallel.


backend-rust/src/indexer.rs
```diff
@@ -138,11 +138,14 @@ SELECT height FROM blocks ORDER BY height DESC LIMIT 1
...
+        let traverse_future =
+            tokio::spawn(traverse_config.traverse(self.block_pre_processor, sender));
+        let process_future =
+            tokio::spawn(processor_config.process_event_stream(self.block_processor, receiver));
         info!("Indexing from block height {}", self.start_height);
-        let (result, ()) = futures::join!(traverse_future, process_future);
-        Ok(result?)
+        let (traverse_result, process_result) = futures::join!(traverse_future, process_future);
+        process_result?;
+        Ok(traverse_result??)
```
Victor-N-Suadicani, line 148: Using `try_join!` instead of `join!` might be more elegant here.

backend-rust/src/indexer.rs
```diff
@@ -415,6 +417,24 @@ LIMIT 1
         .await
         .context("Failed to query data for save context")?;
 
+        let last_block = sqlx::query!(
+            r#"
+SELECT
+  slot_time
+FROM blocks
+ORDER BY height DESC
+LIMIT 1
```
Victor-N-Suadicani, line 426: What if an insertion/update happens into the blocks table between this query and the previous one? Is that okay? If not, maybe a transaction with a lock would make sense.

backend-rust/src/indexer.rs
```diff
@@ -505,13 +523,14 @@ impl ProcessEvent for BlockProcessor {
...
-    /// This is needed in order to compute the finalization time of blocks.
-    last_finalized_height: i64,
     /// The last finalized block hash according to the latest indexed block.
-    /// This is needed in order to compute the finalization time of blocks.
-    last_finalized_hash:   String,
+    /// This is used when computing the finalization time.
+    last_finalized_hash:  String,
+    /// The slot time of the last processed block.
+    /// This is used when computing the block time.
+    last_block_slot_time: NaiveDateTime,
```
Victor-N-Suadicani, line 533: Why is it we do not use `DateTime<Utc>`?

backend-rust/src/indexer.rs
```diff
@@ -672,59 +691,86 @@ impl PreparedBlock {
         })
     }
 
-    async fn save(
-        &self,
+    async fn batch_save(
+        batch: &Vec<Self>,
```
Victor-N-Suadicani, line 695:
```suggestion
        batch: &[Self],
```

backend-rust/src/indexer.rs
```diff
@@ -672,59 +691,86 @@ impl PreparedBlock {
...
         context: &BlockProcessingContext,
         tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
-    ) -> anyhow::Result<Option<BlockProcessingContext>> {
+    ) -> anyhow::Result<BlockProcessingContext> {
+        let mut height = Vec::new();
```
Victor-N-Suadicani, line 699: I can't help but wonder: If we decided on some specific batch size, we could maybe stack-allocate all these using const generics as arrays `[T; N]`, instead of heap-allocating. Not sure if that would make any difference.

If nothing else, some of these could use `with_capacity` instead of new, I think.

backend-rust/src/indexer.rs
```diff
@@ -672,59 +691,86 @@ impl PreparedBlock {
         })
     }
 
-    async fn save(
-        &self,
+    async fn batch_save(
+        batch: &Vec<Self>,
         context: &BlockProcessingContext,
```
Victor-N-Suadicani, line 696: Maybe just take ownership of the context since you just clone it below. Let the caller clone if necessary.

Victor-N-Suadicani (approved):
Just a few suggestions/questions.

backend-rust/src/indexer.rs
```diff
@@ -672,59 +691,86 @@ impl PreparedBlock {
...
-        }
+FROM UNNEST($1::BIGINT[], $2::text[], $3::timestamp[]) AS finalizer(height, finalized, slot_time)
+JOIN blocks l ON finalizer.finalized = l.hash
+WHERE b.finalization_time IS NULL AND b.height <= l.height
+"#,
+            &finalizers,
+            &last_finalized,
+            &finalizers_slot_time
+        )
+        .execute(tx.as_mut())
```
DOBEN, line 772: Can we add a comment on what is happening here please?

backend-rust/src/indexer.rs
limemloh: The difference is that `join!` will wait for all of them to finish, but `try_join!` will return if just one of them results in `Err`. So this was my intent.

backend-rust/src/indexer.rs
```diff
@@ -399,12 +402,11 @@ impl BlockProcessor {
...
-            BlockProcessingContext,
+        let last_finalized_block = sqlx::query!(
             r#"
 SELECT
-  height as last_finalized_height,
-  hash as last_finalized_hash
+  height,
+  hash
 FROM blocks
 WHERE finalization_time IS NULL
```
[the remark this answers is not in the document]
limemloh: Right, I need the hash of the last finalized block, so this is wrong.
Good catch!

backend-rust/src/indexer.rs
limemloh: This would only happen if running multiple indexers, which we don't support for now.
At some later point, we should warn the user and prevent this from happening.

backend-rust/src/indexer.rs
```diff
@@ -466,22 +486,20 @@ impl ProcessEvent for BlockProcessor {
...
         let mut tx = self.pool.begin().await.context("Failed to create SQL transaction")?;
-        let mut override_context = None;
-        for data in batch {
-            let context = override_context.as_ref().unwrap_or(self.current_context.borrow());
-            let new_context = data.save(context, &mut tx).await.context("Failed saving block")?;
-            if new_context.is_some() {
-                override_context = new_context;
+
+        let new_context =
+            PreparedBlock::batch_save(batch, self.current_context.borrow(), &mut tx).await?;
```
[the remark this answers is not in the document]
limemloh: It is still done atomically since the entire batch and the block items are part of the same database transaction, so if something fails, we roll back the entire batch.
Note that the batch size will be 1 when the indexer is caught up to the chain.

backend-rust/src/indexer.rs
limemloh: We don't store the timezone in the database, since they are all UTC, therefore [sqlx provides `NaiveDateTime`](https://docs.rs/sqlx/latest/sqlx/postgres/types/index.html#chrono) when querying and I we don't really gain much from converting it here.

backend-rust/src/indexer.rs
limemloh: In case of something failing we need to rollback the context as well, hence the need for cloning it.
But I'll move it up and make it mutable here instead, to avoid returning anything.

backend-rust/src/indexer.rs
limemloh: The batches will vary in size, and we don't want to force a full batch since it would delay the indexer when it is caught up to the chain.
We could maybe improve a bit if we prepare the collecting in these vecs ahead of calling process, but doubt we would gain a lot from it.
Yes, I forgot to use `with_capacity`

backend-rust/src/indexer.rs
limemloh: I tried explaining it, but it is difficult without a drawing board and my ascii art skills are not what they used to be :sweat_smile:

backend-rust/src/indexer.rs
Victor-N-Suadicani: Why does it matter that both of them finishes if an error occurs in either of them? Are you saying that the side effects of one of them needs to be observed even when the other one fails? A comment explaining why would be good.

backend-rust/src/indexer.rs
Victor-N-Suadicani: Oh I see, we are using `timestamp` instead of `timestamptz`! We should definitely change that. Using timestamp is definitely wrong - there is almost never any reason to use `timestamp` over `timestamptz`. Note that timestamptz does _not_ store the timezone, it just takes time zones into account when saving/loading data (the SQL type `timestamp with time zone` is very badly named). We definitely want to take it into account. For instance, if a PostgreSQL client changed their locale, the input time would be wrong.

Please change all instances of `timestamp` to `timestamptz` and use `DateTime<Utc>` instead (or I/you can do this in a separate PR).

backend-rust/src/indexer.rs
Victor-N-Suadicani: You could still have a "batch" of size 1. Using const generics it would just be changing the const value. Anyways it's probably not important.

backend-rust/src/indexer.rs
limemloh: Let's fix this in a separate PR

backend-rust/src/indexer.rs
limemloh: It is a choice, and either are valid approaches to me. In this case, if traverse exits with an error, we still complete the current processing before exiting.

Victor-N-Suadicani: I definitely don't think this is just a meaningless distinction/choice. I would not say both are equally valid but that they should be used in different situations.

The fact that you do it like this suggests to the reader that the side effects matter. If the side effects matter, you definitely should write it like this and explain why the side effects matter in a comment. But if the side effects do not matter, then writing it like this is misleading, I feel. It makes it seem like the side effects matter, when they actually don't. Using `try_join!` would make it clear that the side effects do not matter. So I would not say they are equally valid, but that they should be used appropriately (`join!` when side effects matter, `try_join!` when they don't).

Also the error is obviously reported faster if you don't wait for the other task to finish, but I imagine the performance is not important here.

DOBEN (approved):

backend-rust/src/indexer.rs
limemloh: I did not call the choice meaningless.

The traverse task will result in an error if it fails to prepare blocks for long enough, this could be due to network issues with the concordium-nodes, in anycase the traverser decided to shut down, so allowing the process_task to gracefully shutdown seems better.
In the end we can always recover, but gracefully shutting down will mean the batch of blocks that are currently getting processed will complete before exit and I don't see why the traverser should prevent that.

PR #274, opened 2024-10-21
Implement transaction metrics query in new Rust backend

Victor-N-Suadicani:
## Purpose

The current transaction metrics query is broken. It times out when choosing the yearly period. We should fix this with the new backend.

## Changes

* Creates a new `transactions_metrics` module which will contain the logic for this query. This also serves as an example of how we can split up the logic in different modules more generally. Draft for now to show this.
* Adds an `index` column to the `transactions` database table (renaming the previous one to `block_index`) which will be a globally unique index and serves as the new primary key.

## Checklist

      hard-to-understand areas.


backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -0,0 +1,52 @@
+use async_graphql::{Context, Object, SimpleObject};
```
Victor-N-Suadicani, line 1: We could also consider making this a more general `transactions` module and include the already-existing transaction-related queries from `graphql_api.rs`.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -193,7 +199,7 @@ CREATE TABLE accounts(
         BIGINT
         NOT NULL,
     -- Connect the account with the transaction creating it.
-    FOREIGN KEY (created_block, created_index) REFERENCES transactions(block_height, index)
+    FOREIGN KEY (created_block, created_index) REFERENCES transactions (block_height, block_index)
```
Victor-N-Suadicani, line 202: Not sure if this should just use the new global index or if it is useful to keep the block/block index way.

backend-rust/src/graphql_api/context_ext.rs
```diff
@@ -0,0 +1,23 @@
+//! Simple extension trait for the GraphQL context to make it easier to extract
+//! certain data.
+
+use async_graphql::Context;
+use sqlx::PgPool;
+
+use super::{ApiError, ApiResult, ApiServiceConfig};
+
+pub(crate) trait ContextExt {
```
limemloh, line 9: Personally I am not a fan of this approach, it is adding an abstraction with one implementation and makes it look like `pool` and `config` is something which is part of the API of `async_graphql`.
To me, the function calls are more straight forward, and I imagine you would not introduce a module just for holding two functions.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -112,8 +112,14 @@ CREATE TABLE blocks(
 
 -- Every transaction on chain.
 CREATE TABLE transactions(
-    -- Index of the transaction within the block.
+    -- Global index of the transaction, i.e. across blocks.
+    -- This is used to quickly determine counts of transactions, so
+    -- this _must_ be strictly increasing without skipping any values.
     index
```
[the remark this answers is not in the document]
Victor-N-Suadicani: Hmm how would I define the next `total_transaction_count` value? Wouldn't I need to do a `COUNT(*)` over the `transactions` table with a `WHERE` filter for the block's index? Not that it would be a big or expensive count, but a +1 for every transaction row is simpler in that sense.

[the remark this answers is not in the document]
limemloh: We have all the transactions part of the block in memory when we prepare the block, so there it could just count and store the transaction count for each block in the prepared block.
When 'saving' we already have the `BlockProcessingContext` which you could extend to track the total transaction count of the previously inserted block and use that to compute the new total when inserting.

backend-rust/src/graphql_api/context_ext.rs
limemloh: I don't think this is self-contained, the only logic in here is the call to `.data()`, the relevant types and the logic for actually setting these values are still defined outside this module.
This logic is also tightly bound to how the `Context<'_>` is set up, so this module is not abstracting anything.

Victor-N-Suadicani: @DOBEN What do you think? Do you think this extension trait approach to get `context.pool()` is nice or do you prefer `get_pool(context)` with the free-standing functions?

DOBEN: If you ask me, I think we should keep the old implementation so we can focus on getting the remaining parts of the block explorer backend ready.

Victor-N-Suadicani:
I implemented the transaction metrics query now. It should work and should be efficient, though I haven't tested with data in the database. `EXPLAIN ANALYZE` suggests that it is fast with the added `slot_time` index though.

The behavior is not exactly the same either. The old behavior put the buckets at certain times, which lead to the last bucket covering into the future, which is strange and means that the last bucket is actually almost always smaller than the rest. The new behavior has the last bucket covering just up until the current moment. This makes the buckets even.

limemloh:
> The behavior is not exactly the same either. The old behavior put the buckets at certain times, which lead to the last bucket covering into the future, which is strange and means that the last bucket is actually almost always smaller than the rest. The new behavior has the last bucket covering just up until the current moment. This makes the buckets even.

This would mean that it behaves differently from the other metrics we have, and we would need to change those as well.
Not sure if this is better though, since the buckets shift in time, the graphs drawn from these metrics will probably start looking like waves as the count in each block shifts from one bucket to the next.

Victor-N-Suadicani:
It's a fair point but I personally feel that's okay. Yes, every bucket would change depending on when you look, but at least the last bucket doesn't look weird. The current solution makes it seem like the last bucket almost always has less activity, which is inaccurate. I feel this is the most accurate way to portray the data, at least if we are talking about a bar chart with buckets. A finer-resolution line chart would not have the same problem to the same extent I think?

One way to fix it would be to fix the buckets in a similar manner as before, but only consider bucket that do not overlap with the future. But then you don't see the most up-to-date data. I feel the solution proposed here is better.

limemloh (approved):
Let's stick to this for now then, we can then evaluate the end result when we have more of the backend ready and start connecting the frontend.

backend-rust/README.md
```diff
@@ -17,6 +17,11 @@ To run the services, the following dependencies are required to be available on
 
 ## Run the Indexer Service
 
+The indexer talks to a Concordium node in order to gather data about the chain, which it then inserts into a PostgreSQL database.
+Note that only one instance of the indexer may run at any one time, as multiple instances would conflict with each other.
+The indexer purposefully performs insertions in a sequential manner, such that table indices can be strictly increasing without skipping any values.
+Since no rows are ever deleted, this allows using the table indices to quickly calculate the number of rows in a table, without having to actually count all rows via a table scan.
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I moved it to a code doc comment on `IndexerService` instead - not sure if there is a better spot but at least it's there.

backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+    last_cumulative_transaction_count: i64,
+    /// Total number of transactions in the requested period.
+    transaction_count: i64,
+    buckets: TransactionMetricsBuckets,
+}
+
+#[derive(SimpleObject)]
+struct TransactionMetricsBuckets {
+    /// The width (time interval) of each bucket.
+    bucket_width: TimeSpan,
```
DOBEN, line 23: A general comment regarding the number of buckets: It seems to be hardcoded to 31 in ccdscan codebase. It would be better if the number of buckets is tracked here and could be general so the front-end has a choice how many buckets to display.

backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+        let last_cumulative_transaction_count = sqlx::query_scalar!(
+            "SELECT cumulative_num_txs FROM blocks ORDER BY height DESC LIMIT 1"
+        )
+        .fetch_one(pool)
+        .await?;
+
+        let interval: PgInterval = period
+            .as_duration()
+            .try_into()
+            .map_err(|e| ApiError::DurationOutOfRange(Arc::new(e)))?;
```
DOBEN, line 58: Just curious: why do we need `arc` on the error?

backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+        let transaction_count =
+            last_cumulative_transaction_count - cumulative_transaction_count_before_period;
+
+        let bucket_width = period.bucket_width();
+
+        // The full period interval, e.g. 7 days.
+        let period_interval: PgInterval = period
+            .as_duration()
+            .try_into()
+            .map_err(|err| ApiError::DurationOutOfRange(Arc::new(err)))?;
```
DOBEN, line 80: The `interval` value above is the same as `period_interval`.

backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+        )
+        .fetch_all(pool)
+        .await?;
+
+        let x_time = rows
+            .iter()
+            .map(|r| r.bucket_time.expect("generated by generate_series so never null"))
+            .collect();
+        let y_last_cumulative_transaction_count =
+            rows.iter().map(|r| r.end_cumulative_num_txs.expect("coalesced")).collect();
```
DOBEN, line 99: Suggestion: An additional explanation in the `expect` part would help to distinguish the error from similar errors further down. E.g. I would mention the field that caused the error: `end_cumulative_num_txs`

backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -0,0 +1,119 @@
...
+        let y_transaction_count = rows
+            .iter()
+            .map(|r| {
+                r.end_cumulative_num_txs.expect("coalesced")
+                    - r.start_cumulative_num_txs.expect("coalesced")
```
DOBEN, line 104: see above

backend-rust/src/graphql_api/transaction_metrics.rs
Victor-N-Suadicani: I was curious as well - the reason seems to be that `async_graphql` requires the return types of the graphql queries to be `Clone`, which means the error type must be clone. However some of the errors in the variants of the error enum are *not* `Clone`, so they are wrapped in `Arc` to make them `Clone`.

backend-rust/src/graphql_api/transaction_metrics.rs
Victor-N-Suadicani: Hmm I'm not sure that makes sense. Each period choice has a set bucket width. I.e. if you choose 7 days, each bucket is 6 hours. That will always determine a set number of buckets. I guess we could change it so you could choose *both* a period and a bucket width separately, but I don't think that would be too useful.

DOBEN: For now, let's keep it as it is, so we don't have to adjust the front-end. I just find it weird that we are fixed to display always `31` buckets on the front-end.

backend-rust/src/graphql_api/transaction_metrics.rs
Victor-N-Suadicani: First of all, the panic message would give the line number so it would be distinguished from the others.

But secondly, the explanation of the expect here is meant to say "this will never panic because this field is coalesced, i.e. using the `COALESCE` function in SQL, therefore it is never `NULL`, therefore this is always `Some`". So the error will never happen in the first place.

PR #283, opened 2024-10-30
Run P7 delegator migrations at every P7 block instead of only payday

limemloh:
## Purpose

Fix bug preventing the ingestion from proceeding.
The issue was that delegators with pending changes from P6 were not migrated immediately in P7, as the migration logic was only run at payday.
This PR changes the migration logic to be run at every block ingested in P7.

### DIFF
--- a/backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs
+++ b/backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs
@@ -29,20 +29,20 @@ public class DelegationImportHandler
         // meaning we can just return for blocks from a protocol version prior to that.
         if (payload.BlockInfo.ProtocolVersion < ProtocolVersion.P4) return resultBuilder.Build();
 
-        // Handle effective pending changes for delegators and update the resultBuilder.
-        if (importPaydayStatus is FirstBlockAfterPayday firstBlockAfterPayday) {
-            if (payload.BlockInfo.ProtocolVersion < ProtocolVersion.P7) {
-                // Stake changes only take effect from the first block in each payday.
-                await _writer.UpdateAccountsWithPendingDelegationChange(firstBlockAfterPayday.PaydayTimestamp,
-                    account => ApplyPendingChange(account, resultBuilder));
-            } else if (payload.BlockInfo.ProtocolVersion == ProtocolVersion.P7) {
-                // Starting from Concordium Protocol Version 7 stake changes are immediate,
-                // meaning no delegators are expected to have pending changes from this point.
-                // Only the first reward day in P7 should this do anything, afterwards it is a
-                // no-op, since no accounts with pending changes are expected.
-                await _writer.UpdateAccountsWithPendingDelegationChange(DateTimeOffset.MaxValue,
-                    account => ApplyPendingChange(account, resultBuilder));
-            }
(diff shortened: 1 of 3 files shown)


DOBEN (approved):

backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs
```diff
@@ -29,20 +29,20 @@ public async Task<DelegationUpdateResults> HandleDelegationUpdates(BlockDataPayl
...
+                account => ApplyPendingChange(account, resultBuilder));
+
+        } else if (payload.BlockInfo.ProtocolVersion == ProtocolVersion.P7) {
+            // Starting from Concordium Protocol Version 7 stake changes are immediate,
+            // meaning no delegators are expected to have pending changes from this point.
+            // Only the block in P7 should this do anything, afterwards it is a
+            // no-op, since no accounts with pending changes are expected.
+            await _writer.UpdateAccountsWithPendingDelegationChange(DateTimeOffset.MaxValue,
+                account => ApplyPendingChange(account, resultBuilder));
+
```
soerenbf, line 45: Not a blocker, but could be:
```suggestion
        } else if (payload.BlockInfo.ProtocolVersion >= ProtocolVersion.P7) {
            // Starting from Concordium Protocol Version 7 stake changes are immediate,
            // meaning no delegators are expected to have pending changes from this point.
            // Only the block in P7 should this do anything, afterwards it is a
            // no-op, since no accounts with pending changes are expected.
            await _writer.UpdateAccountsWithPendingDelegationChange(DateTimeOffset.MaxValue,
                account => ApplyPendingChange(account, resultBuilder));

```

Also seems to reflect the comment more precisely?

backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs
```diff
@@ -29,20 +29,20 @@ public async Task<DelegationUpdateResults> HandleDelegationUpdates(BlockDataPayl
...
+        if (payload.BlockInfo.ProtocolVersion < ProtocolVersion.P7 && importPaydayStatus is FirstBlockAfterPayday firstBlockAfterPayday) {
+            // Handle effective pending changes for delegators and update the resultBuilder.
+            // Stake changes only take effect from the first block in each payday.
+            await _writer.UpdateAccountsWithPendingDelegationChange(firstBlockAfterPayday.PaydayTimestamp,
+                account => ApplyPendingChange(account, resultBuilder));
+
+        } else if (payload.BlockInfo.ProtocolVersion == ProtocolVersion.P7) {
+            // Starting from Concordium Protocol Version 7 stake changes are immediate,
+            // meaning no delegators are expected to have pending changes from this point.
+            // Only the block in P7 should this do anything, afterwards it is a
```
soerenbf, line 41:
```suggestion
            // Only the first block in P7 should this do anything, afterwards it is a
```
?

backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs
limemloh: This is migrating state from P6 to P7, so when we get to P8 it does not need to be run.
Ideally we would not run it more than once, either but this was quick solution.

PR #285, opened 2024-10-31
Account metrics

Victor-N-Suadicani:
## Purpose

Implements the account metrics query.

## Changes

The implementation is very similar to the transaction metrics query. I needed to add an index to ensure efficiency. I have not thoroughly tested the query but I imagine we can do this later.

## Checklist

      hard-to-understand areas.


backend-rust/src/graphql_api/account_metrics.rs
```diff
@@ -0,0 +1,110 @@
...
+impl AccountMetricsQuery {
+    async fn account_metrics(
+        &self,
+        ctx: &Context<'_>,
+        period: MetricsPeriod,
+    ) -> ApiResult<AccountMetrics> {
+        let pool = get_pool(ctx)?;
+
+        let last_cumulative_accounts_created =
+            sqlx::query_scalar!("SELECT MAX(index) FROM accounts").fetch_one(pool).await?.unwrap();
```
limemloh, line 52: **Suggestion:** Avoid panicking.

Since this service will only read data from the database, it can't really do much damage, so I think crashing in the case of something being off is not ideal. It will take down every other query being processed. If we introduce some bug, which can trigger this, it would allow an attacker to easily do a denial-of-service attack.

Instead, I suggest returning an Internal error to the user, which we would then monitor for.

Victor-N-Suadicani: >It will take down every other query being processed.

That's not true - axum catches panics so that only that single request that panics is affected. It will not crash the entire web server, it will just drop the HTTP connection ungracefully. Not too dissimilar to a single thread panicking while other threads are unaffected.

I can definitely understand that you'd want to avoid panics if it would affect other requests or the server as a whole, but it does not. It could not be used as part of a DOS attack.

DOBEN: Reporting an error instead of panicking (even the database should never be corrupted) is better because it makes debugging easier in my opinion if that ever happens.

Victor-N-Suadicani: I don't disagree with that in general, but in this case, this error will essentially never happen and therefore an unwrap is fine I think. The panic will point to the line of code in question and should make it obvious what happened so I don't think, in this particular case, that it hurts debugging.

DOBEN (approved):

backend-rust/src/graphql_api/account_metrics.rs
DOBEN: The error case can happen if e.g. the indexer has not started yet (meaning no data in the database). This is an error and we should be easily debug it from the front-end to report to devOps that they need to start the indexing service.

limemloh: Can't find any information on what axum does when there is a panic, but shutting down the connection ungracefully is not nice either.
For better error message to the user and codebase consistency, it would be good to do the same as we do elsewhere and handle the error.

Victor-N-Suadicani: I'm not sure where exactly it's documented but that is how it works. If we don't like dropping the connection, we could use the [catch panic middleware layer](https://docs.rs/tower-http/latest/tower_http/catch_panic/index.html) that turns it into a 500 response instead. But I think that's outside the scope of this PR.

limemloh: I see that axum spawns a task per request and the panicking is handled by the tokio runtime then, so you are right in it not being a vulnerability.
I would still prefer consistency and a proper error messages for the user.

Victor-N-Suadicani: Alright, I opted to coalesce the max instead, to make it so there is no error and the endpoint will simply report 0 in the case of no data, which seems more correct than giving an error.

limemloh (approved):

PR #290, opened 2024-11-04
Avoid building frontend for each environment and configure using env

limemloh:
## Purpose

Make frontend docker image independent on the network, such that the same image can be used for mainnet, testnet, stagenet and so on.

## Changes

- Update Nuxt runtime configuration to take the relevant settings.
- Allow network selector to be configured at runtime using env variables.
- Provide .env files for different networks.
- Update readme with instructions for how to run the image.


frontend/nuxt.config.ts
```diff
@@ -45,27 +47,22 @@ export default defineNuxtConfig({
...
-
+	// Lock default values for Nuxt to what they were at this date.
 	compatibilityDate: '2024-11-01',
+	// TypeScript configurations.
+	// typescript: {
+	// 	// Enable strict checks.
+	// 	strict: true,
+	// 	// Enable type-checking at build time.
+	// 	typeCheck: true,
+	// },
```
DOBEN, line 67: Should the comment be removed?

DOBEN (approved):

frontend/nuxt.config.ts
limemloh: I enabled it and then there were a lot of type errors, I have addressed these in a [separate PR](https://github.com/Concordium/concordium-scan/pull/291)

PR #293, opened 2024-11-05
Add release workflow and prepare frontend release

limemloh:
## Purpose

Add flow for releasing the frontend using Github Actions and prepare a release.

## Changes

- Add CI checks for the frontend, (linting is failing so will be added later).
- Add release flow for frontend when tags `frontend/*` are pushed.


.github/workflows/frontend-release.yml
```diff
@@ -0,0 +1,37 @@
+name: Publish frontend docker image on DockerHub
+
+on:
+  push:
+    tags:
+      - frontend/*
```
DOBEN, line 6:
```suggestion
      - frontend/*.*.*'
```

.github/workflows/frontend-release.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          flavor: latest=false
+          images: concordium/ccdscan-frontend
+          tags: type=match,pattern=frontend/(.*),group=1
```
DOBEN, line 23: We can add some checks that the image does not already exists on docker hub and a check that the tagged version matches the version in the `package.json` file:
```
      - name: Extract tag from package.json
        id: meta
        run: |
          export VERSION=$(jq -r .version simpleAgeVerification/package.json)
          export FULL_IMAGE_TAG="${{ env.REGISTRY }}/concordium/$IMAGE_NAME:$VERSION"
          echo "::notice FULL_IMAGE_TAG=${FULL_IMAGE_TAG}"
          # Make sure the image does not exist. Abort if we can retrieve any metadata.
          if docker manifest inspect ${FULL_IMAGE_TAG} > /dev/null; then
             echo "::error ${FULL_IMAGE_TAG} already exists"
             exit 1
          elif [ ! "${EXPECTED_TAG}" = "simple-age-verification/${VERSION}" ]; then
             echo "::error Expected tag ${EXPECTED_TAG} does not match the version ${VERSION}."
             exit 1
```

.github/workflows/frontend-release.yml
limemloh: I don't see why this is better, currently I can specify anything for the image, and it does not lock us into some specific format.

.github/workflows/frontend-release.yml
limemloh: I don't see a high risk of overwriting images, we would have to overwrite the git tag first and force push, and in this case it seems desirable to upload a new image again.

limemloh: I'll add the check for the package version

DOBEN (approved):

PR #295, opened 2024-11-06
Account transactions

Victor-N-Suadicani:
## Purpose

Implements the account transactions query, which is used to retrieve transactions for a single account on the account page.

## Checklist

      hard-to-understand areas.


backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -176,16 +180,17 @@ CREATE TABLE transactions(
         JSONB,
 
     -- Within a single block, two transactions cannot share the same index.
-    PRIMARY KEY (block_height, index)
+    PRIMARY KEY (block_height, block_index)
 );
 
+CREATE INDEX transactions_index_idx ON transactions (index);
```
Victor-N-Suadicani, line 186: Not sure if we should just get rid of the block_height/block_index and only use the global index. :shrug:

backend-rust/src/graphql_api.rs
```diff
@@ -2810,14 +2810,99 @@ impl Account {
...
+        .fetch(pool);
+
+        let has_previous_page = sqlx::query_scalar!(
+            "SELECT true
+            FROM transactions
+            WHERE sender = $1 AND index <= $2
+            LIMIT 1",
+            self.index,
+            query.from,
+        )
```
Victor-N-Suadicani, line 2875: This isn't super elegant and requires an extra query but I'm not really sure how else to achieve it.

backend-rust/src/graphql_api.rs
```diff
@@ -2810,14 +2810,99 @@ impl Account {
...
+                    type as "tx_type: DbTransactionType",
+                    type_account as "type_account: AccountTransactionType",
+                    type_credential_deployment as "type_credential_deployment: CredentialDeploymentTransactionType",
+                    type_update as "type_update: UpdateTransactionType",
+                    success,
+                    events as "events: sqlx::types::Json<Vec<Event>>",
+                    reject as "reject: sqlx::types::Json<TransactionRejectReason>"
+                FROM transactions
+                WHERE
+                    sender = $1
```
Victor-N-Suadicani, line 2851: Is this correct? I.e. are transactions associated with the sender account? Or do we need to have a receiver column and use that?

DOBEN: The current CCDScan seems to return all kinds of transactions that the account is somehow involved in e.g. even if the account receives CCD (and is not the initiator/sender of the transaction). I even wonder if ContractUpdate transactions appear if the account gets CCD paid out to its address from a contract address (haven't seen that event yet, so not sure).

Transactions such as where the account receives CCD (not sends the CCD).
- Simple transfer 
- Simple CCD transfer with memo
- Simple CCD transfer with a schedule

backend-rust/migrations/0001_initialize.up.sql
Victor-N-Suadicani: I removed it.

backend-rust/src/graphql_api.rs
```diff
@@ -2810,14 +2810,99 @@ impl Account {
...
+        after: Option<String>,
+        #[graphql(desc = "Returns the last _n_ elements from the list.")] last: Option<u64>,
         #[graphql(desc = "Returns the elements in the list that come before the specified cursor.")]
-        before: String,
+        before: Option<String>,
     ) -> ApiResult<connection::Connection<String, AccountTransactionRelation>> {
-        todo_api!()
+        let config = get_config(ctx)?;
+        let pool = get_pool(ctx)?;
+        let query = ConnectionQuery::<i64>::new(
```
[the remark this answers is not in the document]
Victor-N-Suadicani: No, here `i64` corresponds to the PostgreSQL type, i.e. `BIGINT` which is `i64`. PostgreSQL has no support for unsigned integers. So in this particular case we need to use a signed integer.

backend-rust/src/graphql_api.rs
```diff
@@ -2810,14 +2810,99 @@ impl Account {
...
+            ) ORDER BY index ASC
+            "#,
+            self.index,
+            query.from,
+            query.to,
+            query.desc,
+            query.limit as i64,
+        )
+        .fetch(pool);
+
```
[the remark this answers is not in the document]
Victor-N-Suadicani: Do you think this is included now with the `affected_accounts` table?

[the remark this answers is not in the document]
limemloh: Looks good to me :+1: 
Good job both of you :1st_place_medal:

limemloh (approved):
Great work! Only some minor API details to address.

backend-rust/src/graphql_api.rs
limemloh: If we were to improve, it could be trying to combine some queries to reduce chatter, but this solution should still scale, so I would say this is fine :+1:

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -173,55 +173,64 @@ CREATE TABLE transactions(
...
-    -- Connect the account with the transaction creating it.
-    FOREIGN KEY (created_block, created_index) REFERENCES transactions (block_height, index)
+        NOT NULL
     -- credential_registration_id
 );
 
--- Important for performance when joining accounts with its associated creation block.
-CREATE INDEX accounts_created_block_idx ON accounts (created_block);
+-- Important for performance when joining accounts with its associated creation transaction.
+CREATE INDEX accounts_transaction_index_idx ON accounts (transaction_index);
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I think you're right, this isn't needed.

DOBEN (approved):

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -173,55 +173,64 @@ CREATE TABLE transactions(
...
+        NOT NULL
+        REFERENCES transactions,
+    -- An account affected by this transaction.
+    account_index
+        BIGINT
+        NOT NULL
+        REFERENCES accounts,
+
+    -- A transaction can only affect an account once.
+    PRIMARY KEY (transaction_index, account_index)
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I've tested the queries with `enable_seqscan = false` and it didn't do any sequential scans and used the existing indices. Sometimes PostgreSQL can use compound indices like this one even when only one of the fields are relevant. I'll leave it be for now, we can see if there are any performance issues later.

PR #299, opened 2024-11-13
Sorting and filtering on accounts

Victor-N-Suadicani:
## Purpose

The frontend allows the user to sort by various properties. We need to support this in the new backend.

## Changes

Modifies the SQL query for accounts significantly to filter/sort dynamically based on a desired ordering. Also adds a `num_txs` and `delegated_stake` field to keep track of some of the properties that can be sorted by. The `delegated_stake` field is however not populated by the indexer at the moment (neither is the `amount` field). That is left as an exercise to the reader :upside_down_face:

I have not tested the query thoroughly but I have tried an `EXPLAIN ANALYZE` to ensure the performance is okay - a few additional indices on the `accounts` table were necessary.

## Checklist

      hard-to-understand areas.


limemloh (approved):
Looks good to me :+1:

DOBEN (approved):

PR #307, opened 2024-11-26
Rust release flow

lassemand:
## Purpose

Make release flow

## Changes

Dockerfile
Release file

## Checklist

      hard-to-understand areas.


backend-rust/Dockerfile
```diff
@@ -0,0 +1,23 @@
+ARG build_image=rust:1.82-bookworm
+ARG base_image=debian:bookworm-slim
+FROM ${build_image} AS build
+
+WORKDIR /usr/app
+
+COPY Cargo.toml Cargo.lock ./
+COPY src ./src
+COPY concordium-rust-sdk /usr/app/concordium-rust-sdk
+COPY .sqlx .sqlx
```
lassemand, line 10: We should probably ensure in CI that sqlx has been updated.

.github/workflows/release-backend.yml
```diff
@@ -0,0 +1,13 @@
+on:
+  push:
+    tags:
+      - ccdscan-fungus/*
+
+jobs:
+  release-ccdscan-fungus:
+    uses: concordium/.github/.github/workflows/docker-release-workflow.yaml@v1
```
lassemand, line 8: It has been mentioned that we would rather do copy paste. I would for now rather have it like this so we do not diverge in the repository

.github/workflows/release-backend.yml
```diff
@@ -0,0 +1,13 @@
+on:
+  push:
+    tags:
+      - ccdscan-fungus/*
```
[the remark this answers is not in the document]
lassemand: https://concordium.slack.com/archives/C03DRCC9GR1/p1730893953908189

Seems to be consensus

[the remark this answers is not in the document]
DOBEN: I thought people were just joking and coming up with funny names. it could be more confusing why we call it `fungus/fungy`, so my suggestion is to be super boring and call it `backend/` or if we want to avoid clashes with previous tags `backend-rust`.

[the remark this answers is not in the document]
lassemand: sure, I have also adjusted the repository name on docker hub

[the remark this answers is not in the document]
lassemand: also I called it ccdscan-backend, since that correlates quite nicely with the ccdscan-frontend, and the current backends service name is ccdscan. So we can at a later point remove that to remove confusion

backend-rust/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN cargo build --release --locked
+
+
+FROM ${base_image}
+
+WORKDIR /usr/app
+
+COPY migrations /usr/app/migrations
+COPY --from=build /usr/app/target/release/ccdscan-api /usr/app/target/release/ccdscan-indexer /usr/bin/
+COPY --from=build /usr/local/cargo/bin/sqlx /usr/local/bin/sqlx
```
DOBEN, line 21:
```suggestion

```

backend-rust/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN chmod +x /usr/bin/ccdscan-api /usr/bin/ccdscan-indexer
```
DOBEN, line 23: I  think there should be two releases/dockerfiles (`indexer` and `graph-api` service). This means two independent images in the end since these services are run independently (e.g. we want to run several `graph-api` instances but only a single `indexer`).

similar to:
https://github.com/Concordium/concordium-dapp-examples/tree/main/trackAndTrace/dockerfiles

.github/workflows/release-backend.yml
DOBEN: It adds another level of complexity because people have to check out that above file and figure out what it does (since it is written generically while we only need a few lines of code here to push to docker).

backend-rust/Dockerfile
lassemand: I can indeed do both based on what I have here.

There is a few issues with the above mentioned dockerfiles though:

1. We are building the same code twice which is inefficient
2. We are building the same code without locking. This means that you have no guarantee that the code is actually the same.

lassemand: https://github.com/Concordium/concordium-scan/actions/runs/12048317579

Can see the effect of the altercation in the above mentioned link.

I just ended up using caching.

.github/workflows/release-backend.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Check if images exists
+        run: |
+          set +e
+          docker manifest inspect concordium/ccdscan-indexer:$TAG_VERSION
+          EXITCODE=$?
+          if [ $EXITCODE -eq "0" ]; then
+            echo "Error: ${{ env.FULL_IMAGE_TAG }} already exists."
+            exit 1
+          fi
+          docker manifest inspect concordium/ccdscan-backend:$TAG_VERSION
```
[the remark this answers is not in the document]
lassemand: I choosed this because I had already created the backend repo. Fine with changing as well

DOBEN (approved):

Issue #255, opened 2024-09-27
Implement Query::token


limemloh:
Implement `Query::token` in the Rust GraphQL API and relevant types.   ​<br>​Leave out implementation for `Token::accounts` and `Token::tokenEvents`.

For reference, the [C# GraphQL schema can be found here](https://github.com/Concordium/concordium-scan/blob/main/backend/Tests/Api/GraphQL/\__snapshots_\_/committed-schema.verified.graphql).

PR #313, opened 2024-11-28
Add tokens table

DOBEN:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/255

Related to:
https://github.com/Concordium/concordium-base/pull/588
https://github.com/Concordium/concordium-rust-sdk/pull/220

This PR adds tracking of `cis-2` tokens. `cis-1` tokens have not been considered yet.


backend-rust/src/indexer.rs
```diff
@@ -1810,14 +1844,18 @@ struct PreparedContractInitialized {
...
     amount:           i64,
     height:           i64,
+    cis2_events:      Vec<cis2::Event>,
 }
 
 impl PreparedContractInitialized {
-    fn prepare(
+    async fn prepare(
+        node_client: &mut v2::Client,
+        pool: &PgPool,
```
limemloh, line 1853: We should not access the database in the preprocessing phase, but it seems this is not used anyway.
Please remove

backend-rust/src/indexer.rs
```diff
@@ -1919,10 +2132,181 @@ impl PreparedContractUpdate {
...
+        )
+        .await
     }
 }
 
+async fn process_cis2_events(
+    cis2_events: &[cis2::Event],
+    contract_index: i64,
+    contract_sub_index: i64,
+    pool: &PgPool,
```
limemloh, line 2151: Use the parameter `tx` instead of introducing this `pool` parameter

backend-rust/src/indexer.rs
```diff
@@ -1919,10 +2132,181 @@ impl PreparedContractUpdate {
...
+    transaction_index: i64,
+    tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+) -> anyhow::Result<()> {
+    for log in cis2_events.iter() {
+        // The `total_supply` value of a token is inserted/updated in the database here.
+        // Only `Mint` and `Burn` events affect the `total_supply` of a
+        // token.
+        if let cis2::Event::Mint {
```
limemloh, line 2159: Use `match` statement instead of several `if lets` as this convey better the control flow and check for exhaustive matches

backend-rust/src/indexer.rs
```diff
@@ -1919,10 +2132,181 @@ impl PreparedContractUpdate {
...
 
+async fn process_cis2_events(
+    cis2_events: &[cis2::Event],
+    contract_index: i64,
+    contract_sub_index: i64,
+    pool: &PgPool,
+    transaction_index: i64,
+    tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+) -> anyhow::Result<()> {
+    for log in cis2_events.iter() {
```
limemloh, line 2155: Really minor suggestion and my opinion, but moving the loop out to be where this function is called, will make this function more readable.

backend-rust/src/indexer.rs
```diff
@@ -1919,10 +2132,181 @@ impl PreparedContractUpdate {
...
+                    DO UPDATE SET total_supply = EXCLUDED.total_supply",
+                token_address,
+                contract_index,
+                contract_sub_index,
+                new_total_supply,
+                token_id.to_string(),
+                transaction_index
+            )
+            .execute(tx.as_mut())
+            .await?;
```
limemloh, line 2210: You can do the computation as part of the update on conflict, saving first having to query
```suggestion
            let token_address =
                get_token_address(contract_index as u64, contract_sub_index as u64, token_id)?;

            // Note: Some `buggy` CIS2 token contracts might mint more tokens than the
            // MAX::TOKEN_AMOUNT specified in the CIS2 standard. The
            // `total_supply` eventually overflows in that case.
            let minted = BigDecimal::from_str(&amount.0.to_string())?;

            // If the `token_address` does not exist, insert the new token with its
            // `total_supply` set to `amount`. If the `token_address` exists,
            // update the `total_supply` value by adding the `amount` to the existing
            // value in the database.
            sqlx::query!(
                "
                    INSERT INTO tokens (index, token_address, contract_index, contract_sub_index, \
                 total_supply, token_id, init_transaction_index)
                    VALUES ((SELECT COALESCE(MAX(index) + 1, 0) FROM tokens), $1, $2, $3, $4, $5, \
                 $6)
                    ON CONFLICT (token_address)
                    DO UPDATE SET total_supply = tokens.total_supply + EXCLUDED.total_supply",
                token_address,
                contract_index,
                contract_sub_index,
                minted,
                token_id.to_string(),
                transaction_index
            )
            .execute(tx.as_mut())
            .await?;
```
Likewise for Burn below

backend-rust/src/indexer.rs
```diff
@@ -1919,10 +2132,181 @@ impl PreparedContractUpdate {
...
+            // otherwise use the `amount` as the new `total_supply`.
+            // Note: Some `buggy` CIS2 token contracts might mint more tokens than the
+            // MAX::TOKEN_AMOUNT specified in the CIS2 standard. The
+            // `total_supply` eventually overflows in that case.
+            let new_total_supply = if let Some(row) = row {
+                let current_total_supply: BigDecimal = row.total_supply;
+
+                current_total_supply + BigDecimal::from_str(&amount.0.to_string())?
+            } else {
+                BigDecimal::from_str(&amount.0.to_string())?
```
limemloh, line 2187: Use [BigDecimal::from_biguint](https://docs.rs/bigdecimal/latest/bigdecimal/struct.BigDecimal.html#method.from_biguint) instead of converting to a string

limemloh (changes requested):

backend-rust/src/indexer.rs
```diff
@@ -1860,19 +1948,64 @@ impl PreparedContractInitialized {
...
 }
 
+/// The token address is generated by using a `version byte 2` and concatenating
+/// it with the leb128 byte encoding of the contract index and the leb128 byte
+/// encoding of the contract subindex concatenated with the token id in bytes.
+/// Finally the whole byte array is base 58 check encoded.
+/// https://proposals.concordium.software/CIS/cis-2.html#token-address
+/// The token address is a unique identifier used to distinguish tokens across
+/// all smart contracts.
+pub fn get_token_address(
```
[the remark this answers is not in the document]
limemloh: If you need to update something in the rust sdk submodule, it is already pointing to a commit on this branch https://github.com/Concordium/concordium-rust-sdk/pull/219

backend-rust/src/indexer.rs
```diff
@@ -195,25 +203,27 @@ impl NodeMetricLabels {
 /// of [`Indexer`](concordium_rust_sdk::indexer::Indexer). Since several
 /// preprocessors can run in parallel, this must be `Sync`.
 struct BlockPreProcessor {
+    pool: PgPool,
```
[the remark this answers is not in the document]
DOBEN: good point.

backend-rust/src/indexer.rs
```diff
@@ -2022,17 +2097,33 @@ impl PreparedRejectModuleTransaction {
...
+/// Finally the whole byte array is base 58 check encoded.
+/// https://proposals.concordium.software/CIS/cis-2.html#token-address
+/// The token address is a unique identifier used to distinguish tokens across
+/// all smart contracts.
+pub fn get_token_address(
+    contract_address: ContractAddress,
+    token_id: TokenId,
+) -> Result<String, anyhow::Error> {
+    Ok(format!("{}", TokenAddress::new(contract_address, token_id)))
+}
```
limemloh, line 2112: Not sure this function is justified anymore, maybe just do `TokenAddress::new(contract_address, token_id).to_string()` everywhere instead

Issue #270, opened 2024-10-16
Log internal errors in API service


limemloh:
We should log when we return internal error in the API service as errors.<br><br>Consider implementing some middleware/extension for async-graphql service which catches when internal errors are returned to the user and log these as errors as well.

PR #317, opened 2024-12-04
Add decoding of contract log events

DOBEN:
## Purpose

Add decoding of contract log events in interrupt chain events

## Changes

Add decoding of contract log events in interrupt chain events

### DIFF
--- a/backend-rust/src/graphql_api.rs
+++ b/backend-rust/src/graphql_api.rs
@@ -820,61 +820,8 @@ LIMIT 30", // WHERE slot_time > (LOCALTIMESTAMP - $1::interval)
     // the elements in the list that come after the specified cursor." after:
     // String "Returns the last _n_ elements from the list." last: Int "Returns
     // the elements in the list that come before the specified cursor." before:
-    // String): TokensConnection
-    // TODO
-    // token(contractIndex: UnsignedLong!
+    // String): TokensConnection token(contractIndex: UnsignedLong!
     // contractSubIndex: UnsignedLong! tokenId: String!): Token!
-
-    async fn token<'a>(
-        &self,
-        ctx: &Context<'a>,
-        contract_address_index: ContractIndex,
-        contract_address_sub_index: ContractIndex,
-        token_id: String,
-    ) -> ApiResult<Token> {
-        let pool = get_pool(ctx)?;
-
-        // let row = sqlx::query!(
-        //     r#"SELECT
-        //         module_reference,
-        //         name as contract_name,
-        //         contracts.amount,
-        //         blocks.slot_time as block_slot_time,
-        //         transactions.block_height,
-        //         transactions.hash as transaction_hash,
-        //         accounts.address as creator
-        //     FROM contracts
-        //     JOIN transactions ON transaction_index = transactions.index
-        //     JOIN blocks ON transactions.block_height = blocks.height
-        //     JOIN accounts ON transactions.sender = accounts.index
-        //     WHERE contracts.index = $1 AND contracts.sub_index = $2"#,
-        //     contract_address_index.0 as i64,
-        //     contract_address_sub_index.0 as i64,
-        // )
-        // .fetch_optional(pool)
-        // .await?
-        // .ok_or(ApiError::NotFound)?;
-
-        // let snapshot = ContractSnapshot {
-        //     block_height: row.block_height,
-        //     contract_address_index,
-        //     contract_address_sub_index,
-        //     contract_name: row.contract_name,
-        //     module_reference: row.module_reference,
-        //     amount: row.amount,
-        // };
-
-        Ok(Token {
-            initial_transaction:        Transaction,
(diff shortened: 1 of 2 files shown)


backend-rust/src/graphql_api.rs
```diff
@@ -5061,15 +5062,70 @@ pub struct ChainUpdatePayload {
...
+        let mut connection = connection::Connection::new(true, true);
+
+        for (index, log) in self.contract_logs_raw.iter().enumerate() {
+            let decoded_log =
+                decode_value_with_schema(opt_event_schema.as_ref(), "event", log, "contract_log");
+
+            connection.edges.push(connection::Edge::new(index.to_string(), decoded_log));
+        }
+
+        // Nice-to-have: pagination info but not used at front-end currently.
```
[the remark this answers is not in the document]
DOBEN: Currently, the front-end does not paginate the `contract_logs` (usually there are only a few) and this backend just uses some `default` pagination setting when generating the `connection` for the `contract_logs` as such.

```
let mut connection = connection::Connection::new(true, true);
```

backend-rust/src/graphql_api.rs
```diff
@@ -5061,15 +5062,70 @@ pub struct ChainUpdatePayload {
...
+        .fetch_optional(pool)
+        .await?
+        .ok_or(ApiError::NotFound)?;
+
+        let opt_event_schema = row
+            .display_schema
+            .as_ref()
+            .and_then(|schema| VersionedModuleSchema::new(schema, &None).ok())
+            .and_then(|versioned_schema| {
+                versioned_schema.get_event_schema(&row.contract_name).ok()
```
lassemand, line 5113: Same for this and should probably remove curly brackets for consistency

backend-rust/src/graphql_api.rs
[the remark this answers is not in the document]
DOBEN: I changed the inline comment to `TODO`. I feel it is important to know why `pagination` is not handled here at the moment when reading the code.

backend-rust/src/graphql_api.rs
```diff
@@ -5061,15 +5062,70 @@ pub struct ChainUpdatePayload {
...
+            });
+
+        let mut connection = connection::Connection::new(true, true);
+
+        for (index, log) in self.contract_logs_raw.iter().enumerate() {
+            let decoded_log =
+                decode_value_with_schema(opt_event_schema.as_ref(), "event", log, "contract_log");
```
lassemand, line 5120: I assume "event" should be the display_name?

At the very least it seems to be unused in the method besides providing error messages. And for the error message it seem to say that no schema with such name exist?

backend-rust/src/graphql_api.rs
lassemand, line 5125:
```suggestion
        // TODO: pagination info but not used at front-end currently.
```

DOBEN: I use the comment. For prioritization it is important that it is currently not used and as such not so important to implement vs some other TODOS.
```
// TODO: pagination info but not used at front-end currently.
```

backend-rust/src/graphql_api.rs
lassemand: Based on conversation elsewhere this should not be logging error, but curly brackets can still be removed for consistency

backend-rust/src/graphql_api.rs
```diff
@@ -5105,24 +5152,33 @@ impl ContractInterrupted {
...
-            .and_then(|schema| VersionedModuleSchema::new(schema, &None).ok())
-            .and_then(|versioned_schema| {
-                versioned_schema.get_event_schema(&row.contract_name).ok()
-            });
+        // Get the event schema if it exists.
+        let opt_event_schema = if let Some(event_schema) = row.display_schema.as_ref() {
+            let versioned_schema =
+                VersionedModuleSchema::new(event_schema, &None).map_err(|_| {
+                    ApiError::InternalError(
+                        "Database bytes should be a valid VersionedModuleSchema".to_string(),
```
lassemand, line 5160: We should definitely also include this error to the logs.

lassemand (approved):

backend-rust/src/graphql_api.rs
DOBEN: Yes, I agree. We have an issue for building a middleware to trace all `internalError` in our service (not implemented yet). When addressing the below issues we likely will decide to log/trace more information internally (e.g. exact database connection issues) while reporting only the parts of the internal errors to the user that don't leak any problematic information.
https://github.com/Concordium/concordium-scan/issues/270

backend-rust/src/graphql_api.rs
DOBEN: Revised that part and decided to implement some prober types `SmartContractSchemaNames` for the names of schemas.

PR #323, opened 2024-12-06
Start migrating frontend

lassemand:


frontend/src/queries/useTransactionListQuery.ts
```diff
@@ -59,7 +59,11 @@ const TransactionsQuery = gql<TransactionListResponse>`
 export const useTransactionsListQuery = (
 	variables: Partial<QueryVariables>
 ) => {
+	const fungyApiUrl = inject<string>('fungyApiUrl')
 	const { data, executeQuery } = useQuery({
+		context: {
+			url: fungyApiUrl,
```
lassemand, line 65: We need to do this for all migrated queries

td202 (approved):
Looks reasonable to me.

frontend/src/queries/useAccountQuery.ts
```diff
@@ -150,19 +150,6 @@ address {
 }
 amount
 transactionCount
-baker {
```
limemloh, line 153: I don't think we should remove this

frontend/src/queries/useAccountListQuery.ts
```diff
@@ -64,6 +64,9 @@ const AccountsQuery = gql<AccountsListResponse>`
 
 export const useAccountsListQuery = (variables: AccountListVariables) => {
 	const { data } = useQuery({
+		context: {
+			url: inject<string>('fungyApiUrl'),
```
limemloh, line 68: This is not really that type safe, it would be better to just fetch the runtime config directly

Issue #246, opened 2024-09-27
Implement Account::releaseSchedule and AccountReleaseSchedule::schedule


limemloh:
Implement `Account::releaseSchedule` and `AccountReleaseSchedule::schedule` in the Rust GraphQL API and relevant types.<br>Consider whether `AccountReleaseSchedule::schedule` needs paging.

For reference, the [C# GraphQL schema can be found here](https://github.com/Concordium/concordium-scan/blob/main/backend/Tests/Api/GraphQL/\__snapshots_\_/committed-schema.verified.graphql).

limemloh:
Done in https://github.com/Concordium/concordium-scan/pull/324

PR #324, opened 2024-12-06
Index scheduled transfer events.

td202:
## Purpose

Relates to: #246

This implements indexing for scheduled transfers. The actual queries are not yet implemented.

## Changes

- Add `scheduled_releases` table.
- Add index by `(account_index, release_time)` for efficiently querying the pending releases on a particular account.
- Add handling for `TransferredWithSchedule` and `TransferredWithScheduleAndMemo` events to index the scheduled releases appropriately.

## Checklist

      hard-to-understand areas.


backend-rust/migrations/0003_release_schedule.up.sql
```diff
@@ -0,0 +1,31 @@
+-- Migration script for adding the scheduled release table.
```
limemloh, line 1: I think we should stick to one sql file for now, and only do migrations when we have to, since we will nuke the database for now anyway.

limemloh (approved):
Looks good to me, but we need some logic for these `scheduled_releases` to be resolved and update the account balances accordingly.
We can introduce that in a separate PR.

backend-rust/src/indexer.rs
```diff
@@ -526,6 +526,9 @@ impl ProcessEvent for BlockProcessor {
             }
             out.push_str(format!("\n- {}:{}", block.height, block.hash).as_str())
         }
+        process_release_schedules(new_context.last_block_slot_time, &mut tx)
```
lassemand, line 529: I assume it has been discussed where this should be placed. 

The alternative of handling this in the query and using an index to query it on seems to be the better solution to me. This would be in particular because if we have an index on release_time querying the value is the same performance as now, the api is less reliant on the indexer to be running with a certain frequency, and we are lowering the extra request per cycle that occurs now. 

If we are afraid of querying up too much data we can just run a periodical job cleaning it up. 

That said this does not seem too unreasonable to me, and I believe since the decision has been made we may want to just live with it.

lassemand: Although the change required to make the change is also rather small

backend-rust/src/indexer.rs
```diff
@@ -1894,3 +1922,69 @@ impl PreparedContractUpdate {
...
+            transaction_index,
+            self.account_address,
+            &self.release_times,
+            &self.amounts
+        )
+        .execute(tx.as_mut())
+        .await?;
+        sqlx::query!(
+            "UPDATE accounts
+               SET amount = amount + $1
```
lassemand, line 1981: This indention seems off

lassemand (approved):
Besides

https://github.com/Concordium/concordium-scan/pull/324#discussion_r1890360649

it looks good to me.

backend-rust/src/graphql_api.rs
```diff
@@ -2343,11 +2350,27 @@ struct AccountAddressAmount {
...
+            ApiError::InternalError(
+                "AccountReleaseScheduleItem: No transaction at transaction_index".to_string(),
+            ),
+        )
+    }
+
+    async fn timestamp(&self) -> DateTime { self.timestamp }
+
+    async fn amount(&self) -> Amount { self.amount }
 }
```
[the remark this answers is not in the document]
limemloh: It would expose the `index` and `transaction_index` as well in the API, so they would need to be skipped:
```
#[derive(SimpleObject)]
#[graphql(complex)]
struct AccountReleaseScheduleItem {
    #[graphql(skip)]
    index:             AccountReleaseScheduleItemIndex,
    #[graphql(skip)]
    transaction_index: TransactionIndex,
    timestamp:         DateTime,
    amount:            Amount,
}
```

But I prefer the current, mainly because it is more explicit and sort of isolate details about what goes into the API, especially since the `AccountReleaseScheduleItem` type is used to parse rows from the database.

backend-rust/src/indexer.rs
limemloh: Good point with the API now relying on the indexer running for the query to be correct, I'll fix that.

I think cleaning up is still reasonable during processing of the blocks, since the table won't grow unless the indexer is running anyway.

PR #325, opened 2024-12-10
Account delegation

lassemand:
## Purpose

Implement account delegation

## Changes

Added delegation to the api and indexer on rust ccd scan

## Checklist

      hard-to-understand areas.


backend-rust/src/graphql_api.rs
```diff
@@ -3075,15 +3077,17 @@ struct Account {
     /// The total number of transactions this account has been involved in or
     /// affected by.
     num_txs:           i64,
+    delegated_restake_earnings: Option<bool>,
+    delegated_target_baker_id: Option<i64>
     // Get baker information if this account is baking.
     // baker: Option<Baker>,
-    // delegation: Option<Delegation>,
```
lassemand, line 3080: Already presented on the object directly and not really a reason to abtract it into another object. I am doing that on the graphql though for backwards compatibility.

backend-rust/src/graphql_api.rs
```diff
@@ -3579,7 +3606,6 @@ struct Delegation {
     staked_amount:     Amount,
     restake_earnings:  bool,
     delegation_target: DelegationTarget,
-    pending_change:    PendingDelegationChange,
```
lassemand, line 3582: I should probably leave this for backwards compatility and just provide it with some placeholder value

lassemand: This issue has been fixed by not querying it on the frontend:

https://github.com/Concordium/concordium-scan/pull/326

backend-rust/migrations/0003_account_delegation.up.sql
limemloh, line 1: Let's avoid having multiple migrations until we actually have to, this is just making it harder during development. This by itself is not enough to migrate anyway as the already deployed service would have to nuke their database anyway, unless we implement more migration logic, which is what I don't want us to care about for now.

backend-rust/src/indexer.rs
```diff
@@ -1295,6 +1304,169 @@ impl PreparedAccountCreation {
...
+            }
+            PreparedAccountDelegationEvent::BakerRemoved {
+                baker_id,
+            } => {
+                sqlx::query!(
+                    r#"UPDATE accounts SET delegated_stake = 0, delegated_restake_earnings = false, delegated_target_baker_id = NULL WHERE index = $1"#,
+                    *baker_id
+                )
+                .execute(tx.as_mut())
+                .await?;
```
limemloh, line 1460: I don't this update is needed, since the _delegation event_ of baker removed would be followed by events updating these

limemloh (approved):
Looks good only one minor thing, besides that I would like to avoid dealing with migrations for now, so we should keep to the one big SQL file

backend-rust/src/indexer.rs
lassemand: I would like to keep it as is. This is the what CCDScan already does, and I would like to be the difference of behaviour between the two services as small as possible. Then we can factor it out later if need be.

Since this is only happening on the indexer, the change is speed would not impact the behaviour at all.

backend-rust/migrations/0003_account_delegation.up.sql
lassemand: I can do it in this PR. I will have to drop the database anyways.

lassemand: https://github.com/Concordium/concordium-infra-deployment-charts/pull/104/files#diff-9fc72415c8b50f695b00bc361e0877715fe8d63f1e51b4b93b409b18b7e942ca

I am however handling migrations, so please lets use them when we can. Running `sqlx migrade add` should not be too much of a burden upon the developers.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -206,11 +205,16 @@ CREATE TABLE accounts(
...
         BIGINT
         NOT NULL
-        DEFAULT 0
+        DEFAULT 0,
+    delegated_restake_earnings
+        BOOLEAN
+        NULL,
+    delegated_target_baker_id
+        BIGINT
+        NULL
```
limemloh, line 217: Please add documentation for these columns, it should include when the values would be NULL

limemloh (approved):

backend-rust/src/indexer.rs
DOBEN: I assume a baker cannot `delegate its stake` somewhere else since it is baking itself. If true, then the fields `delegated_...` should already be not set for a valid baker when the `BakerRemoved` event is logged.

DOBEN (approved):
Looks good.

backend-rust/src/indexer.rs
lassemand: https://github.com/Concordium/concordium-scan/blob/516a1c3256555f8a4cedc6f50bd8db231fe4276e/backend/Application/Api/GraphQL/Import/DelegationImportHandler.cs#L196

As mentioned this is what we do, I prefer to align as much as possible with existing behaviour. 

But yes it sounds as if we are just making a baker removed twice.

backend-rust/src/indexer.rs
```diff
@@ -1295,6 +1304,170 @@ impl PreparedAccountCreation {
...
+        account_id: i64,
+        staked:     i64,
+    },
+    StakeDecrease {
+        account_id: i64,
+        staked:     i64,
+    },
+    SetRestakeEarnings {
+        account_id:       i64,
+        restake_earnings: bool,
```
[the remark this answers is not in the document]
lassemand: I agree but the reason for naming it like this is to align with current namings. 

https://github.com/Concordium/concordium-scan/blob/516a1c3256555f8a4cedc6f50bd8db231fe4276e/backend-rust/migrations/0001_initialize.up.sql#L263

I rather it consistent worse namings as compared to inconsistent better and worse namings.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -206,17 +205,27 @@ CREATE TABLE accounts(
...
 );
 
 -- These are important for the sorting options on the accounts query.
 CREATE INDEX accounts_amount_idx ON accounts (amount);
 CREATE INDEX accounts_delegated_stake_idx ON accounts (delegated_stake);
 CREATE INDEX accounts_num_txs_idx ON accounts (num_txs);
+CREATE INDEX accounts_delegated_target_baker_id_index
```
limemloh, line 226: Please add a comment to when this index is interesting, these details can be hard to derive later

limemloh (approved):

backend-rust/src/indexer.rs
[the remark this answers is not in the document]
lassemand: We agreed to do this in a subsequent PR on both fields

backend-rust/migrations/0001_initialize.up.sql
lassemand: I removed this index. I introduced it because it was already present.

lassemand: Removing it

PR #326, opened 2024-12-10
Remove unused field in the frontend

lassemand:
## Purpose

I can ignore pendingChange as per https://linear.app/concordium/issue/CCD-68/implement-accountdelegation on fungy CCDScan as soon as we stop querying it on the frontend. It does not seem to be used.


lassemand:
Testing this against mainnet seems to be working just fine

td202 (approved):
Looks good.

PR #331, opened 2024-12-17
Fixing errors related to account and account delegation

lassemand:
## Purpose

```
{
  "data": null,
  "errors": [
    {
      "message": "Unknown field \"accountsMetrics\" on type \"Query\". Did you mean \"accountMetrics\", \"accountByAddress\"?",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

## Changes

Modified name to be accountsMetrics for backwards compatibility
Added missing events to database structs conversion
CreatedAt is taking in an option making it not working.

### DIFF
--- a/backend-rust/src/graphql_api.rs
+++ b/backend-rust/src/graphql_api.rs
@@ -3141,7 +3141,7 @@ impl Account {
     /// Timestamp of the block where this account was created.
     async fn created_at(&self, ctx: &Context<'_>) -> ApiResult<DateTime> {
         let slot_time = if let Some(transaction_index) = self.transaction_index {
-        sqlx::query_scalar!(
+            sqlx::query_scalar!(
                 "SELECT slot_time
                 FROM transactions
                 JOIN blocks ON transactions.block_height = blocks.height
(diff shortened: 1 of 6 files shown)


limemloh (approved):

backend-rust/src/graphql_api.rs
```diff
@@ -3140,16 +3140,25 @@ impl Account {
...
+                "SELECT slot_time
+                FROM transactions
+                JOIN blocks ON transactions.block_height = blocks.height
+                WHERE transactions.index = $1",
+                transaction_index
+            )
+            .fetch_one(get_pool(ctx)?)
+            .await?
+        } else {
+            sqlx::query_scalar!(
```
lassemand, line 3154: This could also be calculated upon startup and just be used as a constant. 

We have a constant amount of accounts which are root accounts, and hence I rather have the very limited speed degradation with the benefit of more clear scoping.

backend-rust/src/indexer.rs
```diff
@@ -1208,8 +1208,13 @@ impl PreparedEvent {
                 } => None,
 
                 AccountTransactionEffects::DelegationConfigured {
-                    data,
-                } => None,
+                    data: events,
+                } => Some(PreparedEvent::AccountDelegationEvents(
+                    events
```
lassemand, line 1213: This was forgotten as part of the previous PR.

PR #335, opened 2024-12-19
Lma/ccd 74/implement accountrewards

lassemand:
## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


backend-rust/src/indexer.rs
```diff
@@ -542,6 +561,9 @@ impl ProcessEvent for BlockProcessor {
             for item in block.prepared_block_items.iter() {
                 item.save(&mut tx).await?;
             }
+            for item in block.special_items.iter() {
```
lassemand, line 564: We could maybe use a trait here, but since we do not do that for events either, I am fine with this. Also the benefit here is rather small

backend-rust/src/indexer.rs
```diff
@@ -617,6 +639,7 @@ struct BlockData {
     chain_parameters:     ChainParameters,
     tokenomics_info:      RewardsOverview,
     total_staked_capital: Amount,
+    special_events:       Vec<SpecialTransactionOutcome>,
```
lassemand, line 642: We could make this a Stream (As it is provided by the SDK.).

Since it diverges with the rest of the properties on this struct, I would probably leave it as is.

backend-rust/src/indexer.rs
```diff
@@ -876,7 +915,186 @@ WHERE blocks.finalization_time IS NULL AND blocks.height <= last.height
...
+    transaction_type: AccountStatementEntryType,
+}
+
+impl PreparedBlockSpecialEvent {
+    fn prepare(
+        event: &SpecialTransactionOutcome,
+        slot_time: DateTime<Utc>,
+        block_height: AbsoluteBlockHeight,
+    ) -> anyhow::Result<Vec<Self>> {
+        let results: Box<dyn Iterator<Item = Result<Self, anyhow::Error>>> = match &event {
```
lassemand, line 932: This should be rewritten into a Vec. This is the better solution, but would have to be propagated into the database layer to make sense.

backend-rust/src/indexer.rs
```diff
@@ -876,7 +915,186 @@ WHERE blocks.finalization_time IS NULL AND blocks.height <= last.height
...
+            } => Box::new(std::iter::empty()),
+        };
+        results.collect::<Result<Vec<_>, _>>()
+    }
+
+    async fn save(
+        &self,
+        tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+    ) -> anyhow::Result<()> {
+        // Note that this does not include account creation. We handle that when saving
```
lassemand, line 1050: Remove comment

backend-rust/src/indexer.rs
```diff
@@ -876,7 +915,186 @@ WHERE blocks.finalization_time IS NULL AND blocks.height <= last.height
...
+            self.account_address
+        )
+        .execute(tx.as_mut())
+        .await?;
+
+        Ok(())
+    }
+}
+
+/// Prepared block item (transaction), ready to be inserted in the database
```
lassemand, line 1097: No shit

backend-rust/src/indexer.rs
```diff
@@ -767,6 +792,19 @@ impl PreparedBlock {
...
+            .iter()
+            .flat_map(|event| {
+                PreparedBlockSpecialEvent::prepare(
+                    event,
+                    data.block_info.block_slot_time,
+                    data.block_info.block_height,
+                )
+            })
+            .flatten()
+            .filter(|event| event.amount > 0)
```
lassemand, line 806: This differs from the current CCDScan

backend-rust/src/indexer.rs
```diff
@@ -746,6 +769,8 @@ struct PreparedBlock {
     block_last_finalized: String,
     /// Preprocessed block items, ready to be saved in the database.
     prepared_block_items: Vec<PreparedBlockItem>,
+    /// Preprocessed
```
lassemand, line 772: Need proper comment

backend-rust/src/indexer.rs
```diff
@@ -313,72 +314,90 @@ impl Indexer for BlockPreProcessor {
...
-                    .await?;
-                Ok(events)
-            };
+        let result =
+            async move {
+                let mut client1 = client.clone();
+                let mut client2 = client.clone();
+                let mut client3 = client.clone();
+                let mut client4 = client.clone();
+                let mut client5 = client.clone();
```
lassemand, line 323: This is really ugly. We could probably just use a mutex here instead.

backend-rust/src/graphql_api.rs
```diff
@@ -2323,31 +2337,66 @@ struct CollectionSegmentInfo {
...
-#[allow(clippy::enum_variant_names)]
-enum RewardType {
+    async fn block(&self, ctx: &Context<'_>) -> ApiResult<Block> {
+        Block::query_by_height(get_pool(ctx)?, self.block_height).await
+    }
+}
+
+#[derive(Enum, Copy, Clone, PartialEq, Eq, sqlx::Type)]
+#[sqlx(type_name = "reward_type")]
+pub enum RewardType {
```
lassemand, line 2361: There should a relationship between RewardTypes and AccountStatementEntries.

However I do not have any immediate good solutions here. As it is of less importance because it only relates to future maintainability I prefer to leave it be.

backend-rust/src/graphql_api.rs
```diff
@@ -3332,26 +3382,137 @@ impl Account {
...
+            query.limit,
+            &self.index
+        )
+        .fetch(pool);
+
+        let mut connection = connection::Connection::new(false, false);
+        let mut min_index = None;
+        let mut max_index = None;
+        while let Some(statement) = rewards.try_next().await? {
+            min_index = min(min_index, Some(statement.id));
```
lassemand, line 3494: Not working correctly

backend-rust/src/graphql_api.rs
```diff
@@ -3332,26 +3382,137 @@ impl Account {
...
+
+        let mut account_statements = sqlx::query_as!(
+            AccountStatementEntry,
+            r#"
+                SELECT id, amount, entry_type as "entry_type: AccountStatementEntryType", timestamp, account_balance, transaction_id, block_height
+                FROM account_statements
+                WHERE account_index = $4
+                  AND id > $1 AND id < $2
+                ORDER BY id ASC
+                LIMIT $3
```
lassemand, line 3411: Somehow limited to 10 atm.

lassemand: Because variable has been set

backend-rust/src/graphql_api.rs
```diff
@@ -3332,26 +3382,151 @@ impl Account {
...
+            "#,
+            query.from,
+            query.to,
+            query.limit,
+            &self.index
+        )
+        .fetch(pool);
+
+        let mut connection = connection::Connection::new(false, false);
```
lassemand, line 3497: I will generalise this in another PR. I would like to have the sorting function included as well

limemloh: I am not sure that I follow, is there a sorting feature for this query?

backend-rust/src/indexer.rs
limemloh: I disagree, introducing a trait would add more complexity than it hides.

backend-rust/src/indexer.rs
limemloh: Since this step is done in the preprocessing phase and therefore in parallel, I doubt that it would provide a speed-up.

backend-rust/src/indexer.rs
```diff
@@ -876,7 +915,182 @@ WHERE blocks.finalization_time IS NULL AND blocks.height <= last.height
...
+                    slot_time,
+                    transaction_type: AccountStatementEntryType::FinalizationReward,
+                },
+            ],
+            SpecialTransactionOutcome::BlockAccrueReward {
+                ..
+            }
+            | SpecialTransactionOutcome::PaydayPoolReward {
+                ..
+            } => Vec::new(),
```
limemloh, line 1039: I am surprised these are not implemented as these also include rewards for accounts

backend-rust/src/indexer.rs
```diff
@@ -876,7 +915,182 @@ WHERE blocks.finalization_time IS NULL AND blocks.height <= last.height
...
+        &self,
+        tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+    ) -> anyhow::Result<()> {
+        sqlx::query!(
+            "WITH account_info AS (
+            SELECT index AS account_index, amount + $4 AS current_balance
+            FROM accounts
+            WHERE address = $1
+        )
+        INSERT INTO account_statements (
```
limemloh, line 1054: We should also probably track transaction fees and the amount part of different types of transfers in this table, but maybe that is for another PR, please add a task in that case.

backend-rust/src/indexer.rs
lassemand: Lets implement that

backend-rust/src/indexer.rs
lassemand: I will add task for transaction fees, 

For the other types of transfers I will add those as part of this PR.

lassemand: https://linear.app/concordium/issue/CCD-138/accountstatementtransactionfees

backend-rust/src/graphql_api.rs
```diff
@@ -2393,7 +2442,7 @@ struct AccountToken {
     token_id:           String,
     balance:            BigInteger,
     token:              Token,
-    account_id:         i64,
+    account_index:      i64,
```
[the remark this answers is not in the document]
lassemand: Year this seems like some refactoring being too offensive.

backend-rust/src/indexer.rs
lassemand: There seem to be some issue here. 

This was what we did on the old CCDScan, however traversing the database we are getting these into the database. 

I where those stem from and why we differ on those.

lassemand: As this deviates from the existing flow I would like to wait with this. I added a TODO, and will do that after I have validated correctness between the two instances of CCDScan.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -565,3 +577,60 @@ $trigger$ LANGUAGE plpgsql;
...
+        NOT NULL,
+    -- Used as reference for all account statements not of type reward type
+    transaction_id
+        BIGINT
+        NULL
+);
+
+CREATE INDEX account_statements_entry_type_idx ON account_statements (entry_type);
+
+CREATE VIEW account_rewards AS
```
limemloh, line 623: Not sure how this performs, since the query we need is the rewards for some `account_index` and this view is first filtering rewards across every account, and then we filter the ones related to the account (scanning). Maybe postgres is smart enough to rearrange to avoid scanning, but a more direct approach would be to have the index:
```
account_statements (account_index, entry_type)
```
And then query directly: 
```sql
SELECT ... FROM account_statements WHERE account_index = $1 AND entry_type IN (...)
```
Which I expect to scale better

backend-rust/src/graphql_api.rs
```diff
@@ -3466,26 +3547,171 @@ impl Account {
...
+            after,
+            last,
+            before,
+            config.account_statements_connection_limit,
+        )?;
+
+        let mut account_statements = sqlx::query_as!(
+            AccountStatementEntry,
+            r#"
+            SELECT
```
limemloh, line 3571: I think we are missing the subquery for handling the `last` query parameter, currently it seems that it will behave identically to providing `first` as we are not ordering a subquery according to `query.desc` (which is true when `last` is provided, the bad naming of `query.desc` is on me)

backend-rust/src/graphql_api.rs
```diff
@@ -3466,26 +3547,171 @@ impl Account {
...
+        let pool = get_pool(ctx)?;
+        let query = ConnectionQuery::<i64>::new(
+            first,
+            after,
+            last,
+            before,
+            config.reward_connection_limit,
+        )?;
+        let mut rewards = sqlx::query_as!(
+            AccountReward,
```
limemloh, line 3656: Similar to above, I think we are missing handling of `last` query parameter

backend-rust/src/indexer.rs
limemloh: Pretty sure these outcomes were introduced in a later protocol version, so this might be why the code for handling it is scattered in the old backend.

backend-rust/migrations/0001_initialize.up.sql
lassemand: A view is essentially a stored query that you can treat almost like a table. So performance will be the same. 

Even if performance would be an issue this is to be used on a single account. This is by no means a bottleneck of anything, lets not waste time on pseudo optimisations.

limemloh: My comment is more about the query than the view, so I should probably have put this comment at the query instead.

What I mean is that it is not super obvious to me that this will perform well, since the query is essentially this (flattening the view):

```sql
SELECT ... FROM (
    SELECT ... FROM account_statements WHERE entry_type IN ...
) WHERE account_index = $4 AND id > $1 AND id < $2
```

The behavior could be the sub query is using the above index to filter only relevant `entry_type` which should perform fine, but the result contains rows for every account and could be quite large.
Then we filter on the `account_index` and `id`, but since we are working with the result of the subquery postgresql does not have any index to use, so it must scan through the entire result of the subquery to find the relevant rows.

If we instead did:

```sql
SELECT ... FROM account_statements 
WHERE account_index = $4 AND entry_type IN (...) AND id > $1 AND id < $2
```
And introduced an index `account_statements (account_index, entry_type, id)`
The entire query can rely on this index.

lassemand: Sure that is true. I will make the change in a bit, as I need a bit of a break.

lassemand: I am pretty certain it extends the query and does not make a subquery.

https://enterprisedb.com/blog/how-do-postgresql-securitybarrier-views-work

limemloh: We would be relying on the optimizer for this, so it would be good to verify it using `EXPLAIN ANALYZE ...`

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -582,3 +594,45 @@ $trigger$ LANGUAGE plpgsql;
...
+        BIGINT
+        REFERENCES blocks(height)
+        NOT NULL,
+    -- Used as reference for all account statements not of type reward type
+    transaction_id
+        BIGINT
+        NULL
+);
+
+CREATE INDEX account_statements_entry_type_idx ON account_statements (account_index, entry_type);
```
limemloh, line 638: Don't we want the `id` as part of the index as well?
```suggestion
CREATE INDEX account_statements_entry_type_idx ON account_statements (account_index, entry_type, id);
```

backend-rust/src/graphql_api.rs
```diff
@@ -1094,26 +1174,197 @@ impl Account {
...
+            });
+            connection.edges.push(connection::Edge::new(statement.id.to_string(), statement));
+        }
+
+        if let (Some(page_min_id), Some(page_max_id)) = (min_index, max_index) {
+            let result = sqlx::query!(
+                r#"
+                    SELECT MAX(id) as max_id, MIN(id) as min_id
+                    FROM account_rewards
+                    WHERE account_index = $1
```
limemloh, line 1356:
```suggestion
                    WHERE account_index = $1
                        AND entry_type IN (
                            'FinalizationReward',
                            'FoundationReward',
                            'BakerReward',
                            'TransactionFeeReward'
                        )
```

limemloh (approved):

backend-rust/migrations/0001_initialize.up.sql
lassemand:
```

                    AND id > $1
                    AND id < $2
```

We only need because of that, and such index already exists because of the field being primary key

limemloh: I won't be able to use the primary key index because of the other WHERE statements in the query

PR #345, opened 2025-01-02
Fix bug for accounts query, where last and first did not behave

limemloh:
## Purpose

Fix [bug found here](https://github.com/Concordium/concordium-scan/pull/336/files#r1900832176)


backend-rust/src/graphql_api.rs
```diff
@@ -693,21 +693,25 @@ impl BaseQuery {
             -- to select page items from the start/end of the range.
             -- Each page must still independently be ordered ascending.
             -- See also https://relay.dev/graphql/connections.htm#sec-Edge-order
-            ORDER BY CASE
-                WHEN $3 THEN index
-                WHEN $4 THEN amount
-                WHEN $5 THEN num_txs
-                WHEN $6 THEN delegated_stake
-            END ASC",
+            ORDER BY
```
lassemand, line 696: Ehh. Dont we just double sort this?

limemloh: Some of these are redundant, yes, but I chose not to special case these as the query is already plenty complex.

backend-rust/src/graphql_api.rs
```diff
@@ -693,21 +693,25 @@ impl BaseQuery {
...
             query.to,
             matches!(order.field, AccountOrderField::Age),
             matches!(order.field, AccountOrderField::Amount),
             matches!(order.field, AccountOrderField::TransactionCount),
             matches!(order.field, AccountOrderField::DelegatedStake),
             filter.map(|f| f.is_delegator).unwrap_or_default(),
-            matches!(order.dir, OrderDir::Desc),
+            query.desc != matches!(order.dir, OrderDir::Desc),
             query.limit,
+            matches!(order.dir, OrderDir::Desc),
```
DOBEN, line 714: The comments in lines 648 - 651 and lines 692-695 need to be updated since the returned node order is now ASC or DESC.

DOBEN (approved):

backend-rust/src/graphql_api.rs
DOBEN: I was hoping for something to dynamically apply `ASC or DESC` when the query parameter is inserted to avoid some of these cases. But haven't found support for it yet :(.

Looking for something like this:

```
let mut accounts = sqlx::query_as!(
    Account,
    r"
    
    .....     

    ORDER BY
        CASE WHEN $3 THEN index END $10,
        CASE WHEN $4 THEN amount END $10,
        CASE WHEN $5 THEN num_txs  END $10,
        CASE WHEN $6 THEN delegated_stake END $10
    ",
    query.from,
    query.to,
    
    ....
    
    if matches!(order.dir, OrderDir::Desc) { "DESC" } else { "ASC" },
)
.fetch(pool);
```

limemloh: Since these will be prepared statements, it is not possible to change keywords dynamically, only the query parameters can be dynamic, so we have to do something like this.

backend-rust/src/graphql_api.rs
limemloh: I have updated the comments, hopefully making the logic more clear as well

PR #348, opened 2025-01-03
Split graphql api

limemloh:
## Purpose

First PR moving content of graphql_api.rs into separate files, this should not add/remove any logic


backend-rust/src/event/mod.rs
```diff
@@ -0,0 +1,609 @@
+use async_graphql::Union;
```
lassemand, line 1: I believe this file is a mistake?

lassemand (approved):
Looks good besides both `event/mod.rs` and `transaction_event/mod.rs` contains the same data.

backend-rust/src/event/mod.rs
limemloh: yes, you are right

backend-rust/src/transaction_event/credentials.rs
```diff
@@ -0,0 +1,50 @@
...
+
+        if let Ok(v) = u8::try_from(v) {
+            Ok(Self(v))
+        } else {
+            Err(InputValueError::expected_type(value))
+        }
+    }
+
+    fn to_value(&self) -> Value { Value::Number(self.0.into()) }
+}
```
DOBEN, line 26: Maybe moving into the `types.rs` folder where the other ScalarTypes are.

backend-rust/src/types.rs
```diff
@@ -0,0 +1,151 @@
+use anyhow::Context;
+use async_graphql::{scalar, InputValueError, InputValueResult, Scalar, ScalarType, Value};
+
```
DOBEN, line 3: Suggestion: Calling the file `scalar_types`.

DOBEN (approved):

PR #358, opened 2025-01-06
[DRAFT] Accountrewards transactions

lassemand:
## Purpose

TODO

## Changes

TODO

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


backend-rust/src/indexer.rs
```diff
@@ -803,15 +803,28 @@ impl PreparedBlock {
...
                 .push(PreparedBlockItem::prepare(node_client, data, item_summary, item).await?)
         }
-        let special_items: Vec<_> = data
+        let prepared_account_statements_iter =
+            data.events.iter().zip(data.items.iter()).map(|(item_summary, item)| {
+                PreparedAccountStatement::prepare_from_block_item(
+                    item_summary,
+                    data.block_info.block_height,
+                )
+            });
```
limemloh, line 812: I think we will have an issue with this approach of collecting them all in one list like this and then process them.
We need them to be intertwined with the other events when storing to the database. When we `save` these to the database it selects the balance of the account, and therefore if we update the balance from anywhere else when processing the block item, like an in a contract update, we are in trouble.

limemloh (changes requested):

backend-rust/src/indexer.rs
lassemand:
```
        let combined_statements: Vec<_> =
            prepared_account_statements_iter.chain(special_items_iter).collect();
```
I am doing that?

backend-rust/src/indexer.rs
```diff
@@ -2907,15 +2961,18 @@ async fn process_cis2_event(
 
 struct PreparedScheduledReleases {
     account_address: Arc<String>,
-    release_times:   Vec<DateTime<Utc>>,
-    amounts:         Vec<i64>,
-    update_balance:  PreparedUpdateAccountBalance,
+    release_times: Vec<DateTime<Utc>>,
+    amounts: Vec<i64>,
+    target_account_balance_update: PreparedUpdateAccountBalance,
+    source_account_balance_update: PreparedUpdateAccountBalance,
```
lassemand, line 2967: It might be we only have to make this a PreparedAccountStatement, I just do not understand why we are not updating the account as well.

limemloh: I think this should update balance of both sender and receiver immediately, so it look correct now :+1:

backend-rust/src/indexer.rs
```diff
@@ -3007,25 +3097,39 @@ impl PreparedCcdTransferEvent {
...
-                account_address: receiver_address.clone(),
-                change:          amount,
+                account_address:   receiver_address.clone(),
+                change:            amount,
+                account_statement: PreparedAccountStatement {
+                    account_address: receiver_address.clone().to_string(),
+                    amount,
+                    block_height: block_height.height.try_into()?,
+                    transaction_type: AccountStatementEntryType::TransferIn,
+                },
```
limemloh, line 3121: I think these two could use `PreparedUpdateAccountBalance::prepare` introduced above

backend-rust/src/indexer.rs
```diff
@@ -2403,6 +2424,34 @@ impl PreparedContractUpdate {
             } => vec![],
         };
 
+        let prepared_account_statement = match event {
+            ContractTraceElement::Updated {
+                data,
+            } => {
+                if let Address::Account(account_address) = data.instigator {
+                    Some(PreparedAccountStatement {
```
limemloh, line 2432: I think this should be a `PreparedUpdateAccountBalance` now

backend-rust/src/indexer.rs
```diff
@@ -2403,6 +2424,34 @@ impl PreparedContractUpdate {
...
+                    })
+                } else {
+                    None
+                }
+            }
+            ContractTraceElement::Transferred {
+                amount,
+                to,
+                ..
+            } => Some(PreparedAccountStatement {
```
limemloh, line 2446: Likewise here, this should be a `PreparedUpdateAccountBalance` now

limemloh (approved):
Looks good to me, only small comments

Issue #258, opened 2024-09-27
Implement Token::tokenEvents, Token::accounts, Accounts::tokens and Contracts::tokens


limemloh:
Implement:

- [x] `Token::tokenEvents`
- [X] `Token::accounts` [https://github.com/Concordium/concordium-scan/pull/359/files](https://github.com/Concordium/concordium-scan/pull/359/files)
- [x] `Account::tokens` 
- [X] `Contracts::tokens`

in the Rust GraphQL API and relevant types.

For reference, the [C# GraphQL schema can be found here](https://github.com/Concordium/concordium-scan/blob/main/backend/Tests/Api/GraphQL/\__snapshots_\_/committed-schema.verified.graphql).

PR #359, opened 2025-01-07
Add token-account relationship

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-scan/issues/258

## Changes

Add `Token::accounts` query.


backend-rust/src/graphql_api/token.rs
```diff
@@ -75,6 +78,61 @@ impl Token {
     async fn contract_address_formatted(&self, ctx: &Context<'_>) -> ApiResult<String> {
         Ok(format!("<{},{}>", self.contract_index, self.contract_sub_index))
     }
+
+    async fn accounts(
```
limemloh, line 82: Not sure what the current behavior is in the API, but we are including accounts with zero balance here, as long as they have been involved with the token at some point, not sure whether this is the desired behavior.

If we choose to filter out the accounts with zero balance then we cannot do the `index_per_token` trick...

backend-rust/src/indexer.rs
```diff
@@ -2565,3 +2751,146 @@ impl PreparedScheduledReleases {
         Ok(())
     }
 }
+
+// Extends the `Account` type by including the `AccountIndex`.
+pub struct AccountExtended {
```
limemloh, line 2756: Can we avoid having these types by fetching the index from the table during processing?
[Just like we do here](https://github.com/Concordium/concordium-scan/blob/9d6255d16e5b1ece5326030f9d6087ffbd28ab52/backend-rust/src/indexer.rs#L1035)

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -540,6 +540,59 @@ CREATE INDEX token_idx ON tokens (contract_index, contract_sub_index, token_id);
...
+    -- encoding of the contract subindex concatenated with the token id in bytes.
+    -- Finally the whole byte array is base 58 check encoded.
+    -- https://proposals.concordium.software/CIS/cis-2.html#token-address
+    token_address
+        TEXT
+        NOT NULL,
+    -- Token id of the token.
+    token_id
+        TEXT
+        NOT NULL,
```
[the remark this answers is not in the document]
DOBEN: good suggestion

backend-rust/src/graphql_api/token.rs
DOBEN: I exclude tokens with a 0 balance now while still displaying tokens with a negative balance.
Making "buggy" negative balances visible helps smart contract developers to debug their contracts.

backend-rust/src/graphql_api/token.rs
```diff
@@ -75,6 +78,61 @@ impl Token {
...
+            items.pop();
+        }
+        let has_previous_page = min_index > 0;
+
+        Ok(AccountsCollectionSegment {
+            page_info: CollectionSegmentInfo {
+                has_next_page,
+                has_previous_page,
+            },
+            total_count: items.len().try_into()?,
```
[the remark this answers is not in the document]
DOBEN: Changed it in the new code. let's create a separate PR for the other areas.

backend-rust/src/graphql_api/token.rs
```diff
@@ -124,12 +142,28 @@ impl Token {
         }
         let has_previous_page = min_index > 0;
 
+        let total_count: i32 = sqlx::query_scalar!(
+            "SELECT
+                MAX(index_per_token)
```
limemloh, line 147: I think this is off by one, since the `index_per_token` starts from `0`.
```suggestion
                MAX(index_per_token) + 1
```

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -540,6 +540,38 @@ CREATE INDEX token_idx ON tokens (contract_index, contract_sub_index, token_id);
...
+    -- The accumulated balance of the token by the above account calculated by considering all `MintEvents`, `BurnEvents` and 
+    -- `TransferEvents` associated to the token and account. If no balance is specified when inserting a new row in the table,
+    -- the default balance 0 is used.
+    balance
+        NUMERIC
+        NOT NULL
+        DEFAULT 0,
+
+    -- Ensure that each token_index and account_index pair is unique.
+    CONSTRAINT unique_token_account_relationship UNIQUE (token_index, account_index)
```
limemloh, line 572: It seems that we never use the `index` column for this table, so we can get the constraint if we make these primary keys
```suggestion
    PRIMARY KEY (token_index, account_index)
```

limemloh (approved):

backend-rust/migrations/0001_initialize.up.sql
limemloh: no longer relevant

PR #361, opened 2025-01-08
Update account balance for transfer events

limemloh:
## Purpose

Update account balance for various transfer events


DOBEN (approved):
Looks good.

Actions that affect the CCD balance of an account that are not covered yet (can be a different PR):
- transaction fee
- a smart contract sents CCD to an account (ContractTraceElement::Transferred)

PR #363, opened 2025-01-08
Support indexing P8 new suspended and resume baker events

limemloh:
## Purpose

Basic support for new P8 baker events as the rust SDK is updated.

Still missing implementation for API and frontend


backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -312,7 +312,19 @@ CREATE TABLE bakers(
...
+    -- Transaction used for self suspending.
+    -- This is not null only when a baker have send the transaction for self suspending.
+    -- This should always be null, when `inactive_suspended` column is not null.
+    self_suspended
+        BIGINT
+        REFERENCES transactions,
+    -- Block which suspended this baker due to inactivity.
+    -- This is not null only when a baker got suspended due to inactivity.
+    -- This should always be null, when `self_suspended` column is not null.
+    inactive_suspended
```
DOBEN, line 325: From a user's perspective: capturing the `prime_for_suspension` event and displaying it at the front-end, so people have 24 hours to fix their node, might be good to add here similar to this final `inactive-suspension` state (I assume that is for next PR).

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -312,7 +312,19 @@ CREATE TABLE bakers(
     -- Fraction of finalization rewards charged by the pool owner.
     -- Stored as a fraction of an amount with a precision of `1/100_000`.
     finalization_commission
+        BIGINT,
+    -- Transaction used for self suspending.
+    -- This is not null only when a baker have send the transaction for self suspending.
```
DOBEN, line 317:
```suggestion
    -- Transaction index used for self-suspending.
    -- This is not null only when a baker has sent the transaction for self-suspending.
```

DOBEN (approved):

backend-rust/migrations/0001_initialize.up.sql
limemloh: Right, I'll add this and the missing indexing, but leave out the API part of this

backend-rust/src/indexer.rs
```diff
@@ -3501,81 +3638,155 @@ impl PreparedUpdateAccountBalance {
...
 }
 
-/// Represent the event of a transfer of CCD from one account to another.
-pub struct PreparedCcdTransferEvent {
-    /// Updating the sender account balance.
-    update_sender:   PreparedUpdateAccountBalance,
-    /// Updating the receivers account balance.
-    update_receiver: PreparedUpdateAccountBalance,
+/// Represent the potential event of baker being "unprimed" for suspension.
+/// The baker of the block, plus the signers of the quarum certificate when
```
td202, line 3720:
```suggestion
/// The baker of the block, plus the signers of the quorum certificate when
```

backend-rust/src/indexer.rs
```diff
@@ -3501,81 +3638,155 @@ impl PreparedUpdateAccountBalance {
...
-        self.update_sender.save(tx, Some(transaction_index)).await?;
-        self.update_receiver.save(tx, Some(transaction_index)).await?;
+        sqlx::query!(
+            "UPDATE bakers
+                SET primed_for_suspension = NULL
+                WHERE
+                    primed_for_suspension IS NOT NULL
+                    AND id = ANY ($1)",
+            &self.baker_ids,
+        )
```
td202, line 3754: It might be worth caching (some of the information from) the bakers table, so that there's no need to make an update query at all in the common case where nothing is being updated.

backend-rust/src/indexer.rs
```diff
@@ -2053,6 +2080,37 @@ impl PreparedBakerEvent {
...
+            } => {
+                sqlx::query!(
+                    "UPDATE bakers
+                     SET
+                         self_suspended = NULL,
+                         inactive_suspended = NULL,
+                         primed_for_suspension = NULL
+                     WHERE id=$1",
+                    baker_id
+                )
```
td202, line 2110: Self-suspending and resuming shouldn't affect the `primed_for_suspension` status. The node itself only resets the primed status when a validator shows activity. Because the status only applies to current validators (i.e. in the committee, as opposed to active ones, which just have validator records on the accounts) when a snapshot is taken, this effectively clears the primed status, because the validator will now be suspended. I think the short of this is that you should just clear the `primed_for_suspension` when the validator is automatically suspended, and not in these cases.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -312,7 +312,27 @@ CREATE TABLE bakers(
...
+    -- Block which suspended this baker due to inactivity.
+    -- This is not null only when a baker got suspended due to inactivity.
+    -- This should always be null, when `self_suspended` or `primed_for_suspension` column is not
+    -- null.
+    inactive_suspended
+        BIGINT
+        REFERENCES blocks,
+    -- Block which is primed for suspension this baker due to inactivity.
+    -- This is not null only when a baker got primed for suspension due to inactivity.
+    -- This should always be null, when `self_suspended` or `inactive_suspended` column is not null.
```
td202, line 332: See other comments, but I think it's OK for both this and `self_suspended` to be present.

backend-rust/src/indexer.rs
```diff
@@ -3501,81 +3638,155 @@ impl PreparedUpdateAccountBalance {
...
     }
 }
 
-/// Represent the event of a transfer of CCD from one account to another.
-pub struct PreparedCcdTransferEvent {
-    /// Updating the sender account balance.
-    update_sender:   PreparedUpdateAccountBalance,
-    /// Updating the receivers account balance.
-    update_receiver: PreparedUpdateAccountBalance,
+/// Represent the potential event of baker being "unprimed" for suspension.
```
DOBEN, line 3719:
```suggestion
/// Represent the potential event of bakers being "unprimed" for suspension.
```

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -312,7 +312,27 @@ CREATE TABLE bakers(
...
+    primed_for_suspension
```
DOBEN, line 333: Since we don't have any front-end design it is a bit unclear what makes sense to collect in the database. I see two options:
- We replicate the same data as the node captures (I think that is what Thomas suggests)
- We capture data as we want to display it at the front-end in the database, e.g. I suspect we do not overload the state of the baker at the front-end. A baker can only be displayed suspended on-time and the latest suspension event is the one that counts:

    ```
NotSuspended
Suspended(blockheight)
PrimedForSuspension(blockheight)
    ```

```
enum SuspensionCause {
            SelfSuspended(blockheight)
          InactivitySuspended( blockheight)
}
```

    ```

backend-rust/src/indexer.rs
limemloh: Yes, you are probably right.
To avoid too much complexity, I'll add a partial index on the bakers where primed_for_suspension is not null.

backend-rust/migrations/0001_initialize.up.sql
limemloh: Since the node tracks state for self-suspension and primed for suspensions separately, we have to do the same.
If a validator is primed for suspension, then self-suspends and resume, they would still be primed. With the above state we would not know to show them as primed in this case.

PR #375, opened 2025-01-10
Add account tokens query

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-scan/issues/258

## Changes

Add `Account::Tokens`


backend-rust/src/graphql_api.rs
```diff
@@ -1105,22 +1104,127 @@ impl Account {
...
+        // Tokens with 0 balance are filtered out. We still display tokens with a
+        // negative balance (buggy cis2 smart contract) to help smart contract
+        // developers to debug their smart contracts.
+        // Excluding tokens with 0 balances does not scale well currently. We might need
+        // to introduce a specific index to optimize this query in the future.
+        let mut row_stream = sqlx::query_as!(
+            AccountTokenInterim,
+            "
+            SELECT 
+                token_id,
```
lassemand, line 1143: You might as well just do *

lassemand (commented):
LGTM

lassemand (approved):

backend-rust/src/graphql_api.rs
```diff
@@ -1105,22 +1104,127 @@ impl Account {
...
+                    account_id,
+                    row_num
+                FROM (
+                    SELECT
+                        token_id,
+                        contract_index,
+                        contract_sub_index,
+                        balance AS raw_balance,
+                        account_index AS account_id,
+                        ROW_NUMBER() OVER (ORDER BY account_tokens.index) - 1 AS row_num
```
limemloh, line 1164: I am not so sure whether this will scale, but since this is a cursor connection we can avoid counting row numbers by having the cursor capture everything we need to find the row of `before` and `after`

backend-rust/src/graphql_api.rs
```diff
@@ -1105,22 +1104,127 @@ impl Account {
...
+                    FROM account_tokens
+                    JOIN tokens
+                        ON tokens.index = account_tokens.token_index
+                    WHERE account_tokens.balance != 0 
+                        AND account_tokens.account_index = $5
+                ) AS filtered_tokens
+                WHERE $1 < row_num AND row_num < $2
+                ORDER BY
```
limemloh, line 1172: We order by the row number here, but these were not assigned in any particular order, so it seems a bit random what order we get in the end.
Do you know the order of the .NET backend?

backend-rust/src/graphql_api/token.rs
```diff
@@ -23,6 +23,35 @@ impl QueryToken {
     ) -> ApiResult<Token> {
         let pool = get_pool(ctx)?;
```
limemloh, line 24: The variable `pool` seems to never be used
```suggestion
```

backend-rust/src/graphql_api.rs
```diff
@@ -1105,22 +1104,127 @@ impl Account {
...
+                        contract_index,
+                        contract_sub_index,
+                        balance AS raw_balance,
+                        account_index AS account_id,
+                        ROW_NUMBER() OVER (ORDER BY account_tokens.index) - 1 AS row_num
+                    FROM account_tokens
+                    JOIN tokens
+                        ON tokens.index = account_tokens.token_index
+                    WHERE account_tokens.balance != 0 
+                        AND account_tokens.account_index = $5
```
limemloh, line 1169: If we are sticking to this solution, we should probably introduce the partial index:
```sql
CREATE INDEX non_zero_account_token_idx ON account_tokens (account_index) WHERE balance != 0;
```

backend-rust/src/graphql_api.rs
DOBEN: https://github.com/Concordium/concordium-scan/blob/main/backend/Application/Api/GraphQL/Accounts/Account.cs#L84

backend-rust/src/graphql_api.rs
DOBEN: We currently don't have an index (or can create one easily) to achieve this. We would need an index `account_token_per_account` and delete (remove the row) from `account_tokens` whenever an `account_index` balance what go down to 0.

backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -597,11 +600,17 @@ CREATE TABLE account_tokens (
...
+    -- Every time an `account_token` is inserted or updated in this table, a sequential index is assigned 
+    -- to the operation and tracked in this sequence.
+    -- This sequence is used to sort/filter the newest tokens transferred to an account address.
+    change_seq BIGINT DEFAULT nextval('account_tokens_update_seq'),
 
     -- Ensure that each token_index and account_index pair is unique.
     CONSTRAINT unique_token_account_relationship UNIQUE (token_index, account_index)
 );
 
+CREATE INDEX non_zero_account_token_idx ON account_tokens (account_index) WHERE balance != 0;
```
limemloh, line 612: We need it to include the `change_seq` as well
```suggestion
CREATE INDEX non_zero_account_token_idx ON account_tokens (account_index, change_seq) WHERE balance != 0;
```

backend-rust/src/graphql_api.rs
```diff
@@ -1105,22 +1104,110 @@ impl Account {
...
+        // negative balance (buggy cis2 smart contract) to help smart contract
+        // developers to debug their smart contracts.
+        // The tokens are sorted/filtered by the newest tokens transferred to an account
+        // address.
+        let mut row_stream = sqlx::query_as!(
+            AccountTokenInterim,
+            "
+            SELECT * FROM (
+                SELECT * FROM (
+                    SELECT
```
limemloh, line 1144: Since we are not using row_number any more we can merge the two inner subqueries into one,

backend-rust/src/graphql_api.rs
```diff
@@ -1105,22 +1104,110 @@ impl Account {
...
+                    FROM account_tokens
+                    JOIN tokens
+                        ON tokens.index = account_tokens.token_index
+                    WHERE account_tokens.balance != 0 
+                        AND account_tokens.account_index = $5
+                ) AS filtered_tokens
+                WHERE $1 < change_seq AND change_seq < $2
+                ORDER BY
+                    CASE WHEN $4 THEN change_seq END DESC,
+                    CASE WHEN NOT $4 THEN change_seq END ASC
```
limemloh, line 1160: If we want the most recent changes to be at the front of the collection, it needs to be opposite (`query.desc` was badly named by me)
```suggestion
                    CASE WHEN NOT $4 THEN change_seq END DESC,
                    CASE WHEN $4 THEN change_seq END ASC
```

limemloh (approved):

Issue #369, opened 2025-01-09
Implement RejectedReceive::message.


limemloh:
Optional schema decoding of the message send as parameter for the contract which rejected.

PR #385, opened 2025-01-13
Implement RejectedReceive::message

limemloh:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/369

Implement `RejectedReceive::message`.


backend-rust/src/transaction_reject.rs
```diff
@@ -525,322 +544,432 @@ pub struct PoolClosed {
...
+            .context("Invalid receive name for RejectedReceive")?;
+        let schema_type = match schema.get_receive_param_schema(
+            receive_name.contract_name(),
+            receive_name.entrypoint_name().into(),
+        ) {
+            Ok(t) => t,
+            Err(VersionedSchemaError::NoContractInModule)
+            | Err(VersionedSchemaError::NoReceiveInContract) => return Ok(None),
+            Err(err) => {
+                anyhow::bail!("Database bytes should be a valid VersionedModuleSchema: {}", err);
```
DOBEN, line 965: This error is an `InternalError` in my opinion since the indexer should not insert an invalid schema into the database in the first place. I suggest that the function `process_message` returns the tuple (message, message_parsing_error) and returns an error on `InternalError`.

backend-rust/src/transaction_reject.rs
```diff
@@ -525,322 +544,432 @@ pub struct PoolClosed {
...
+                    ON smart_contract_modules.module_reference = contracts.module_reference
+            WHERE index = $1 AND sub_index = $2",
+            i64::try_from(self.contract_address.index.0)?,
+            i64::try_from(self.contract_address.sub_index.0)?
+        )
+        .fetch_one(tx.as_mut())
+        .await?;
+        let Some(schema) = schema else {
+            // No schema found in the smart contract module.
+            return Ok(None);
```
DOBEN, line 951: For a smart contract developer, it can be helpful for debugging if the parsing is not successful due to if no `schema` is in the contract module or `NoContractInModule`, `NoReceiveInContract`, or `NoParamsInReceive` error occurred (these cases are all logged as `None`). Should we consider logging them as `message_parsing_error`.

limemloh: I have changed the error into a status message instead.
Please have a look

DOBEN (approved):
LGTM

PR #387, opened 2025-01-14
Account search

lassemand:
## Purpose

Implement search functions for accounts


backend-rust/src/graphql_api.rs
```diff
@@ -1834,14 +1835,112 @@ impl SearchResult {
...
+                    transaction_index,
+                    address,
+                    amount,
+                    delegated_stake,
+                    num_txs,
+                    delegated_restake_earnings,
+                    delegated_target_baker_id
+                FROM accounts
+                WHERE
+                    address LIKE $5 || '%'
```
lassemand, line 1894: I am looking into https://www.postgresql.org/docs/current/pgtrgm.html to optimise this

PR #391, opened 2025-01-14
Fungy dashboard

lassemand:
## Purpose

The time is still used even if multiple blocks was part of the batch.

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -453,8 +453,6 @@ async fn compute_total_stake_capital(
 struct BlockProcessor {
     /// Database connection pool
     pool: PgPool,
-    /// Metric counting how many blocks was saved to the database successfully.
-    blocks_processed: Counter,
     /// Metric counting the total number of failed attempts to process
     /// blocks.
     processing_failures: Counter,
@@ -509,12 +507,6 @@ LIMIT 1
             last_cumulative_num_txs: last_block.cumulative_num_txs,
         };
 
-        let blocks_processed = Counter::default();
-        registry.register(
-            "blocks_processed",
-            "Number of blocks save to the database",
-            blocks_processed.clone(),
-        );
         let processing_failures = Counter::default();
         registry.register(
             "processing_failures",
@@ -532,7 +524,6 @@ LIMIT 1
         Ok(Self {
             pool,
             current_context: starting_context,
-            blocks_processed,
             processing_failures,
             processing_duration_seconds,
             max_successive_failures,
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -579,7 +579,7 @@ impl ProcessEvent for BlockProcessor {
         tx.commit().await.context("Failed to commit SQL transaction")?;
         // Update metrics.
         let duration = start_time.elapsed();
-        self.processing_duration_seconds.observe(duration.as_secs_f64() / batch.len() as f64);
+        self.processing_duration_seconds.observe(duration.as_secs_f64());
         self.blocks_processed.inc_by(u64::try_from(batch.len())?);
         // Update the current context when we are certain that nothing failed during
         // processing.
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -453,15 +453,13 @@ async fn compute_total_stake_capital(
 struct BlockProcessor {
     /// Database connection pool
     pool: PgPool,
-    /// Metric counting how many blocks was saved to the database successfully.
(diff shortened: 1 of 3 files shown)


backend-rust/src/indexer.rs
```diff
@@ -557,30 +548,28 @@ impl ProcessEvent for BlockProcessor {
...
+            self.processing_duration_seconds.observe(duration.as_secs_f64());
         }
         process_release_schedules(new_context.last_block_slot_time, &mut tx)
             .await
             .context("Processing scheduled releases")?;
         tx.commit().await.context("Failed to commit SQL transaction")?;
-        // Update metrics.
-        let duration = start_time.elapsed();
-        self.processing_duration_seconds.observe(duration.as_secs_f64() / batch.len() as f64);
-        self.blocks_processed.inc_by(u64::try_from(batch.len())?);
```
lassemand, line 583: already there now as part of bucket.

backend-rust/src/indexer.rs
```diff
@@ -557,30 +548,28 @@ impl ProcessEvent for BlockProcessor {
...
             for item in block.prepared_block_items.iter() {
                 item.save(&mut tx).await?;
             }
             for item in block.special_items.iter() {
                 item.save(&mut tx, None).await?;
             }
-            out.push_str(format!("\n- {}:{}", block.height, block.hash).as_str())
+            out.push_str(format!("\n- {}:{}", block.height, block.hash).as_str());
+            let duration = start_time.elapsed();
+            self.processing_duration_seconds.observe(duration.as_secs_f64());
```
lassemand, line 567: Fixes bucketing

backend-rust/src/indexer.rs
```diff
@@ -557,30 +548,28 @@ impl ProcessEvent for BlockProcessor {
...
         }
         process_release_schedules(new_context.last_block_slot_time, &mut tx)
             .await
             .context("Processing scheduled releases")?;
         tx.commit().await.context("Failed to commit SQL transaction")?;
-        // Update metrics.
-        let duration = start_time.elapsed();
-        self.processing_duration_seconds.observe(duration.as_secs_f64() / batch.len() as f64);
```
lassemand, line 582: Dividing here makes implicit assumption about how this is going to be used, in the end throwing away information.

Furthermore the bucketing is useless as currently defined

backend-rust/src/indexer.rs
limemloh: Not exactly, because if it fails halfway through a batch it will restart and the above bucket would count some twice then.
Where incrementing at this point means that nothing can fail anymore.

backend-rust/src/indexer.rs
limemloh: I agree the division is not right, but I still think it is interesting to measure the processing time for a batch.
Maybe capturing the batch size in another dimension of the histogram is a better solution?

lassemand: That will be too complex.

Made two histograms, where one contains the batch size and the other contains the processing time.
indexer_processor_batch_size_sum is the successfully written blocks
indexer_processor_batch_size_count is the successfully written batches
```
# HELP indexer_processor_batch_size Batch sizes.
# TYPE indexer_processor_batch_size histogram
indexer_processor_batch_size_sum 14.0
indexer_processor_batch_size_count 14
indexer_processor_batch_size_bucket{le="1.0"} 14
indexer_processor_batch_size_bucket{le="2.0"} 14
indexer_processor_batch_size_bucket{le="3.0"} 14
indexer_processor_batch_size_bucket{le="4.0"} 14
indexer_processor_batch_size_bucket{le="5.0"} 14
indexer_processor_batch_size_bucket{le="6.0"} 14
indexer_processor_batch_size_bucket{le="7.0"} 14
indexer_processor_batch_size_bucket{le="8.0"} 14
indexer_processor_batch_size_bucket{le="9.0"} 14
indexer_processor_batch_size_bucket{le="10.0"} 14
indexer_processor_batch_size_bucket{le="+Inf"} 14
```

backend-rust/src/indexer.rs
```diff
@@ -509,12 +509,6 @@ LIMIT 1
             last_cumulative_num_txs: last_block.cumulative_num_txs,
         };
 
-        let blocks_processed = Counter::default();
-        registry.register(
```
limemloh, line 513: I think this is still interesting to have

limemloh (approved):

backend-rust/src/indexer.rs
lassemand: Exactly the same as:

indexer_processor_batch_size_sum

PR #392, opened 2025-01-15
Update .NET SDK and make .NET backend compatible with P8

limemloh:
## Purpose

Ensure .NET backend is compatible with P8.

## Changes

- Update .NET SDK to P8 support.
- Introduce `ChainParametersV3` in graphql API.


backend/Application/Api/GraphQL/ChainParametersV3.cs
```diff
@@ -0,0 +1,51 @@
...
+        return other != null &&
+               base.Equals(other) &&
+               RewardPeriodLength == other.RewardPeriodLength;
+    }
+
+    public override bool Equals(object? obj)
+    {
+        if (ReferenceEquals(null, obj)) return false;
+        if (ReferenceEquals(this, obj)) return true;
+        return obj.GetType() == GetType() && Equals(obj as ChainParametersV2);
```
lassemand, line 34: You have to validate that obj is of type ChainParametersV2.

lassemand: I assume you basically just have to cast to V3 instead.

limemloh: Nice catch

lassemand (approved):

PR #393, opened 2025-01-15
Simplify query

DOBEN:
## Purpose

Simplify query since fields are not used by the front-end.

Related to https://github.com/Concordium/concordium-scan/pull/388

### DIFF
--- a/frontend/src/queries/useTokenQuery.ts
+++ b/frontend/src/queries/useTokenQuery.ts
@@ -26,7 +26,7 @@ transaction {
 }
 event {
   __typename
-  ...on Cis2BurnEvent {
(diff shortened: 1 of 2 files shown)


frontend/src/queries/useTokenQuery.ts
```diff
@@ -26,7 +26,7 @@ transaction {
 }
 event {
   __typename
-  ...on CisBurnEvent {
+  ...on Cis2BurnEvent {
```
limemloh, line 29: Can we avoid the breaking change of the name here?
So far we are backwards compatible.

lassemand (approved):

PR #397, opened 2025-01-16
Order blocks descendingly and compute the next/previous page

limemloh:
## Purpose

Adjust the block list query to use descending order.
Compute the flags for next and previous page as well.

### DIFF
--- a/backend-rust/src/graphql_api/block.rs
+++ b/backend-rust/src/graphql_api/block.rs
@@ -85,6 +85,7 @@ impl QueryBlocks {
             false
         };
         let has_next_page = if let Some(last) = rows.last() {
+            // Genesis block have height 0, so we check whether the last block is higher.
             last.height > 0
         } else {
             false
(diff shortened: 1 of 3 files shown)


backend-rust/src/graphql_api/block.rs
```diff
@@ -64,29 +64,35 @@ impl QueryBlocks {
...
+            sqlx::query_scalar!("SELECT true FROM blocks WHERE height > $1 LIMIT 1", first.height)
+                .fetch_optional(pool)
+                .await?
+                .flatten()
+                .unwrap_or_default()
+        } else {
+            false
+        };
+        let has_next_page = if let Some(last) = rows.last() {
+            last.height > 0
```
lassemand, line 88: Can we name 0 with something meaningful? 

For example: 
```
let genesis_height = 0;
last.height > genesis_height
```

lassemand (approved):

backend-rust/src/graphql_api/block.rs
limemloh: Added a comment

PR #402, opened 2025-01-17
Fix total count values

DOBEN:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/401


limemloh (approved):
These `Count(*)` will probably not scale, but we can access the performance impact during next week and fix it.

DOBEN:
I analyzed them with `EXPLAIN ANALYZE` and they don't do a `sequential scan` on the table.
Even so they have no dedicated index for their query, they made use of a `particalIndex/primaryKeyIndex` already present to achieve an `Index Only Scan`. Let's if the performance gets to bad we will add dedicated indexes.

PR #403, opened 2025-01-17
Track balance updates for accounts and contracts

limemloh:
## Purpose

Add remaining tracking of balances for accounts and contracts.

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -51,7 +51,7 @@ use sqlx::PgPool;
 use std::{convert::TryInto, sync::Arc, time::Duration};
 use tokio::{time::Instant, try_join};
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, info};
 
 /// Service traversing each block of the chain, indexing it into a database.
 ///
@@ -785,7 +785,7 @@ struct PreparedBlock {
     /// Preprocessed block items, ready to be saved in the database.
     prepared_block_items: Vec<PreparedBlockItem>,
     /// Preprocessed block special items, ready to be saved in the database.
-    special_items:        Vec<PreparedAccountStatement>,
+    special_items:        Vec<PreparedUpdateAccountBalance>,
 }
 
 impl PreparedBlock {
@@ -812,13 +812,13 @@ impl PreparedBlock {
             .special_events
             .iter()
             .flat_map(|event| {
-                PreparedAccountStatement::prepare_from_special_transaction(
+                PreparedUpdateAccountBalance::prepare_from_special_transaction(
                     event,
                     data.block_info.block_height,
                 )
             })
             .flatten()
-            .filter(|event| event.amount > 0)
+            .filter(|event| event.change > 0)
             .collect();
 
         Ok(Self {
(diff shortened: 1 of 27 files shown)


backend-rust/src/indexer.rs
```diff
@@ -932,121 +932,13 @@ WHERE blocks.finalization_time IS NULL AND blocks.height <= last.height
 }
 
 struct PreparedAccountStatement {
-    account_address:  String,
+    account_address:  Arc<String>,
```
lassemand, line 935: I do not really get the benefit of this but sure.

backend-rust/src/indexer.rs
```diff
@@ -1293,298 +1176,353 @@ impl PreparedBlockItem {
 }
 
 /// Different types of block item events that can be prepared.
-enum PreparedEvent {
+enum PreparedBlockItemEvent {
     /// A new account got created.
     AccountCreation(PreparedAccountCreation),
+    /// An account transaction event.
+    AccountTransaction(Box<PreparedAccountTransaction>),
```
lassemand, line 1183: Please remove this Box

backend-rust/src/indexer.rs
```diff
@@ -1293,298 +1176,353 @@ impl PreparedBlockItem {
...
+        data: &BlockData,
+        item_summary: &BlockItemSummary,
+        item: &BlockItem<EncodedPayload>,
+    ) -> anyhow::Result<Self> {
+        match &item_summary.details {
+            BlockItemSummaryDetails::AccountCreation(details) => Ok(
+                PreparedBlockItemEvent::AccountCreation(PreparedAccountCreation::prepare(details)?),
+            ),
+            BlockItemSummaryDetails::AccountTransaction(details) => {
+                let sender = Arc::new(details.sender.to_string());
```
lassemand, line 1200: I see no value in this extra step. I would just pass the string

backend-rust/src/indexer.rs
limemloh: The point is to avoid allocating strings everywhere when cloning, this will instead point to the same allocation and should reduce memory usage.
But I'll admit that I don't have any numbers to back this up.

backend-rust/src/indexer.rs
limemloh: It is mainly there because clippy complained about the variants in the enum being of different sizes.
I could probably be placed better deeper in the `PreparedAccountTransaction` but we can think about that later.

backend-rust/src/indexer.rs
limemloh: As mentioned in another comment, this ensures that we only allocate the string representation of the sender once in memory and the rest just refer to this same location.

backend-rust/src/indexer.rs
lassemand: atleast it is consistent.

backend-rust/src/indexer.rs
```diff
@@ -1293,298 +1176,353 @@ impl PreparedBlockItem {
...
+                sender,
+                data.amount,
+                height,
+                false,
+            )?),
+            AccountTransactionEffects::TransferredToPublic {
+                amount,
+                ..
+            } => PreparedEvent::EncryptedBalance(PreparedUpdateEncryptedBalance::prepare(
+                sender, *amount, height, true,
```
lassemand, line 1493: This would be cleaner if it was an enum instead of a bool

backend-rust/src/indexer.rs
```diff
@@ -2554,13 +2570,39 @@ impl PreparedContractUpdate {
...
+}
+
+impl PreparedContractTraceEvent {
+    async fn save(
+        &self,
+        tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+        transaction_index: i64,
+    ) -> anyhow::Result<()> {
+        match self {
+            PreparedContractTraceEvent::Upgrade(event) => event.save(tx, transaction_index).await,
```
lassemand, line 2591: Please ensure this here makes sense. I have limited domain knowledge on this particular field about expected behaviour in the underlying types.

lassemand (approved):
The freshly introduced Contract related event types are very hard for me to know how they are supposed to be working. Please do an extra check on behaviour there.

backend-rust/src/indexer.rs
```diff
@@ -1293,298 +1176,353 @@ impl PreparedBlockItem {
...
+            ),
+            AccountTransactionEffects::ContractInitialized {
+                data: event_data,
+            } => PreparedEvent::ContractInitialized(
+                PreparedContractInitialized::prepare(node_client, data, event_data, sender).await?,
+            ),
+            AccountTransactionEffects::ContractUpdateIssued {
+                effects,
+            } => PreparedEvent::ContractUpdate(
+                PreparedContractUpdates::prepare(node_client, data, effects).await?,
```
lassemand, line 1385: Should the fees for making contract updates also be added here?

limemloh: They are all included in the transaction fee by the protocol, so we don't have to do anything

DOBEN (approved):

backend-rust/src/indexer.rs
```diff
@@ -2385,13 +2411,38 @@ impl PreparedContractUpdate {
...
+                update.amount,
+                data.finalized_block_info.height,
+            )?),
+            ContractTraceElement::Transferred {
+                from,
+                amount,
+                to,
+            } => PreparedContractTraceEvent::Transfer(PreparedTraceEventTransfer::prepare(
+                *from,
+                Arc::new(to.to_string()),
```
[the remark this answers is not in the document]
DOBEN: Similarly, I am a bit more in favor of parsing `amount` as `Amount` type to the `prepare` functions rather than `i64/u64`.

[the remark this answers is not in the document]
limemloh: Only the account address is represented using a Arc<String>, but we could wrap this is a AccountAddress newtype instead.

The issue is not passing the string around, but rather cloning the string storing the same account address in memory several times.

It is a small optimization though, but not an uncommon thing in rust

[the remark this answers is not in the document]
limemloh: We can discuss what to do and fix it in another PR

PR #404, opened 2025-01-19
Bump frontend version

lassemand:


limemloh (commented):
It would be good to update det change log with this version

limemloh (approved):

PR #406, opened 2025-01-20
Sanitize metadata url

limemloh:
## Purpose

The database is complaining about unsupported unicode characters again, so I moved the sanitizing code into where we construct the types instead of right before inserting.

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -3195,10 +3195,6 @@ async fn process_cis2_token_event(
             )
             .to_string();
 
-            // Since PostgreSQL Text data type does not support NUL we must replace these
-            // before inserting. These are replaced by the a Unicode 'REPLACEMENT CHARACTER'
-            // (U+FFFD).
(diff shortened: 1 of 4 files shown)


backend-rust/src/indexer.rs
```diff
@@ -3223,7 +3219,7 @@ async fn process_cis2_token_event(
                 token_address,
                 contract_index,
                 contract_sub_index,
-                metadata_url.as_str(),
+                metadata_url.url(),
```
lassemand, line 3222: pub fn url(&self) -> &str { &self.url }

This is still just returning the string?

lassemand, line 3222: I would rather do the replacement on the indexing. I am not even sure this will work

limemloh: It will be done as part of indexing. The replacement is done in the conversion code, which is used by the indexer

lassemand (approved):

DOBEN (approved):

PR #416, opened 2025-01-21
Only fetch blocks when starting from P6

limemloh:
## Purpose

Fix error where we were fetching block certificates for blocks prior to P6, causing invalid argument gRPC error.
## Changes

- Only update bakers unmarking primed for suspension starting from P8

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -30,8 +30,8 @@ use concordium_rust_sdk::{
         AbsoluteBlockHeight, AccountStakingInfo, AccountTransactionDetails,
         AccountTransactionEffects, BakerId, BlockItemSummary, BlockItemSummaryDetails,
         ContractAddress, ContractInitializedEvent, ContractTraceElement, DelegationTarget,
-        PartsPerHundredThousands, RejectReason, RewardsOverview, SpecialTransactionOutcome,
-        TransactionType,
+        PartsPerHundredThousands, ProtocolVersion, RejectReason, RewardsOverview,
+        SpecialTransactionOutcome, TransactionType,
     },
     v2::{
         self, BlockIdentifier, ChainParameters, FinalizedBlockInfo, QueryError, QueryResult,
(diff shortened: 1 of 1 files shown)


backend-rust/src/indexer.rs
```diff
@@ -3728,6 +3745,14 @@ struct PreparedUnmarkPrimedForSuspension {
 
 impl PreparedUnmarkPrimedForSuspension {
     fn prepare(data: &BlockData) -> anyhow::Result<Self> {
+        if data.block_info.protocol_version < ProtocolVersion::P8 {
+            // Baker suspension was introduced as part of Concordium Protocol Version 8,
+            // meaning for blocks prior to that no baker can be primed for
+            // suspension.
+            return Ok(Self {
```
lassemand, line 3752: This sounds a lot like an error

backend-rust/src/indexer.rs
```diff
@@ -676,6 +692,7 @@ struct BlockData {
     tokenomics_info:      RewardsOverview,
     total_staked_capital: Amount,
     special_events:       Vec<SpecialTransactionOutcome>,
+    /// Certificates included in the block.
```
lassemand, line 695: Thisis kind of given by the name. :)

backend-rust/src/indexer.rs
```diff
@@ -335,96 +335,112 @@ impl Indexer for BlockPreProcessor {
...
+                        RewardsOverview::V1 {
+                            total_staked_capital,
+                            ..
+                        } => Ok(total_staked_capital),
+                    }
+                };
+                let get_block_certificates = async move {
+                    // Fetching the block certificates prior to P6 results in a NotFound gRPC error,
+                    // so we produce the empty type of certificates instead.
+                    if block_info.response.protocol_version < ProtocolVersion::P6 {
```
lassemand, line 406: It should really be done in the SDK

backend-rust/src/indexer.rs
```diff
@@ -335,96 +335,112 @@ impl Indexer for BlockPreProcessor {
...
-                }
-                RewardsOverview::V1 {
-                    total_staked_capital,
-                    ..
-                } => total_staked_capital,
-            };
-            let node_response_time = start_fetching.elapsed();
-            self.node_response_time.get_or_create(label).observe(node_response_time.as_secs_f64());
+                let (total_staked_capital, certificates) =
+                    try_join!(get_total_staked_capital, get_block_certificates)?;
```
lassemand, line 419: These should be done within the block where it is needed so we only wait on that instead of waiting on all of the async blocks to finish before we can proceed with this one

backend-rust/src/indexer.rs
```diff
@@ -335,96 +335,112 @@ impl Indexer for BlockPreProcessor {
...
+                    // so we produce the empty type of certificates instead.
+                    if block_info.response.protocol_version < ProtocolVersion::P6 {
+                        Ok(BlockCertificates {
+                            quorum_certificate:       None,
+                            timeout_certificate:      None,
+                            epoch_finalization_entry: None,
+                        })
+                    } else {
+                        let response = client7.get_block_certificates(fbi.height).await?;
+                        Ok(response.response)
```
DOBEN, line 414: Currently, we only use the fetched data for ProtocolVersion::P8 and above. If it stays like this we could avoid fetching the data here for protocolVersions below (minor optimization).

DOBEN (approved):

backend-rust/src/indexer.rs
limemloh: This is not a bug, but an explicit choice for the gRPC API. Might not have been the right choice though.
Fixing it in the SDK, would mean we first have to fetch the block info to find the protocol version and then fetch it, which is not ideal either, so we do it here since we have the block_info anyway

backend-rust/src/indexer.rs
limemloh: Yes, you are right.
Will improve this

backend-rust/src/indexer.rs
limemloh: Well, it could have been the certificates related to this block and not the ones are included in the block.

PR #418, opened 2025-01-21
Add epoch duration query

DOBEN:
## Purpose

Expose epoch duration.

### DIFF
--- a/backend-rust/src/graphql_api.rs
+++ b/backend-rust/src/graphql_api.rs
@@ -473,6 +473,23 @@ impl BaseQuery {
         }
     }
 
+    async fn import_state<'a>(&self, ctx: &Context<'a>) -> ApiResult<ImportState> {
+        let epoch_duration =
+            sqlx::query_scalar!("SELECT epoch_duration FROM current_consensus_status")
+                .fetch_optional(get_pool(ctx)?)
+                .await?
+                .ok_or(ApiError::NotFound)?;
+
+        Ok(ImportState {
+            epoch_duration: TimeSpan(
+                Duration::try_milliseconds(epoch_duration).ok_or(ApiError::InternalError(
+                    "Epoch duration in database should be valid duration in milliseconds."
+                        .to_string(),
+                ))?,
+            ),
+        })
+    }
+
     async fn tokens(
         &self,
         ctx: &Context<'_>,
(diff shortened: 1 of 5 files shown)


lassemand:
I do not know what this is supposed to be used for. :)

DOBEN:
It is used to calculate the `payDay` duration:
https://github.com/Concordium/concordium-scan/blob/main/frontend/src/components/BakerDetails/BakerDetailsContainer.vue#L17

backend-rust/src/graphql_api.rs
```diff
@@ -473,6 +473,23 @@ impl BaseQuery {
...
+        let epoch_duration =
+            sqlx::query_scalar!("SELECT epoch_duration FROM current_consensus_status")
+                .fetch_optional(get_pool(ctx)?)
+                .await?
+                .ok_or(ApiError::NotFound)?;
+
+        Ok(ImportState {
+            epoch_duration: TimeSpan(
+                Duration::try_milliseconds(epoch_duration).ok_or(ApiError::InternalError(
+                    "Epoch duration in database should be valid duration in milliseconds."
```
lassemand, line 486: Perhaps include the epoch_duration into the error message.

backend-rust/src/indexer.rs
```diff
@@ -707,6 +707,8 @@ async fn save_genesis_data(endpoint: v2::Endpoint, pool: &PgPool) -> anyhow::Res
     let mut tx = pool.begin().await.context("Failed to create SQL transaction")?;
     let genesis_height = v2::BlockIdentifier::AbsoluteHeight(0.into());
     {
+        let epoch_duration = client.get_consensus_info().await?.epoch_duration.num_milliseconds();
```
lassemand, line 710: As per talk I do believe this could be stored inmem on the api upon startup. Maybe even just hardcode the values per env.

If you find that too annoying just keep it as is, since this will not be performance impacting at all.

lassemand (approved):

backend-rust/src/indexer.rs
DOBEN: Based on some discussion. We do not want `envs` because they need to be set at startup `correctly` and locked-up for each network as such.

To look up the value at start-up in the `graphQL_API` service is a good suggestions and we will consider adding a node client to the service if we have more such cases. 

For now we keep this look-up in the indexer.

PR #419, opened 2025-01-22
Cleanup listen on tasks

lassemand:
## Purpose

I am about to add another task on the API.

I want to avoid `n^(n-1)` calls for listening on other services to cancel. :)


docker-compose.yaml
```diff
@@ -6,9 +6,10 @@ version: "3"
 services:
   backend:
     image: ${CCDSCAN_BACKEND_IMAGE-concordium/ccdscan:test}
-    build: ./backend
-    ports:
-    - "${CCDSCAN_BACKEND_PORT-5000}:5000"
+    build:
+      context: ./backend
+      args:
+        DOTNET_VERSION: '6.0'
```
lassemand, line 12: I have no idea why it does not set the args per default tbh.

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -116,41 +116,32 @@ async fn main() -> anyhow::Result<()> {
     // Await for signal to shutdown or any of the tasks to stop.
     tokio::select! {
         _ = tokio::signal::ctrl_c() => {
-            info!("Received signal to shutdown");
```
limemloh, line 119: I think this info is useful, since it provides feedback to the user and since it is the graceful shutdown it might take some time

limemloh (approved):

PR #424, opened 2025-01-23
Add transactions per block query

DOBEN:
## Purpose

https://linear.app/concordium/issue/CCD-81/implement-blocktransactions

### DIFF
--- a/backend-rust/migrations/0001_initialize.up.sql
+++ b/backend-rust/migrations/0001_initialize.up.sql
@@ -214,6 +214,9 @@ CREATE TABLE transactions(
         JSONB
 );
 
+-- Important for quickly filtering transactions in a given block.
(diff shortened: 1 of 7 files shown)


backend-rust/migrations/0001_initialize.up.sql
```diff
@@ -214,6 +214,9 @@ CREATE TABLE transactions(
         JSONB
 );
 
+-- Important for quickly filtering transactions in a given block.
+CREATE INDEX transactions_block_idx ON transactions (block_height);
```
limemloh, line 218: We also filter on the `index` of the transaction, so we need it to be part of the index.
```suggestion
CREATE INDEX transactions_block_idx ON transactions (block_height, index);
```

Issue #423, opened 2025-01-23
Collection of small API changing improvements


DOBEN:
**Task description**

Collection of API breaking improvements:

**Sub-tasks**
- [ ] Use `rewardPeriodLength` directly in `latestChainParameters` (remove `ChainParametersV1`) in https://github.com/Concordium/concordium-scan/blob/main/frontend/src/queries/useBakerQuery.ts
- [ ] Fix typos `Cis2` instead of `Cis`  in  https://github.com/Concordium/concordium-scan/blob/main/frontend/src/queries/useTokenQuery.ts

PR #427, opened 2025-01-24
Add payday summary logic

DOBEN:
## Purpose

https://linear.app/concordium/issue/CCD-72/implement-querypaydaystatus

### DIFF
--- a/backend-rust/src/graphql_api.rs
+++ b/backend-rust/src/graphql_api.rs
@@ -32,11 +32,12 @@ use anyhow::Context as _;
 use async_graphql::{
     http::GraphiQLSource,
     types::{self, connection},
-    Context, EmptyMutation, Enum, MergedObject, Object, Schema, SimpleObject, Subscription, Union,
+    ComplexObject, Context, EmptyMutation, Enum, MergedObject, Object, Schema, SimpleObject,
+    Subscription, Union,
 };
 use async_graphql_axum::GraphQLSubscription;
 use block::Block;
-use chrono::Duration;
+use chrono::{Duration, TimeDelta, Utc};
 use concordium_rust_sdk::{
     base::contracts_common::schema::VersionedSchemaError, id::types as sdk_types,
 };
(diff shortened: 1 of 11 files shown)


backend-rust/src/graphql_api.rs
```diff
@@ -787,6 +816,69 @@ pub enum LatestChainParameters {
...
+pub struct PaydayStatus {
+    next_payday_time:             DateTime,
+    #[graphql(skip)]
+    opt_last_payday_block_height: Option<i64>,
+}
+
+#[ComplexObject]
+impl PaydayStatus {
+    // Future improvement (breaking changes): The front end only uses the
+    // `lastPaydayBlock` which is returned here.
```
limemloh, line 837: We should probably create an issue instead of this comment

backend-rust/src/indexer.rs
```diff
@@ -3593,6 +3593,9 @@ impl PreparedCcdTransferEvent {
 enum PreparedSpecialTransactionOutcome {
     /// Distribution of various rewards.
```
limemloh, line 3594:
```suggestion
    /// Distribution of various rewards. Excluding mint rewards.
```

backend-rust/src/indexer.rs
```diff
@@ -3846,6 +3856,45 @@ struct PreparedValidatorSuspension {
     block_height: i64,
 }
 
+/// Represents a payday block with its CCD mint rewards and the associated block
+/// height. The block was identified by observing the
+/// `SpecialTransactionOutcome::Mint` event.
+struct PreparedPayDayBlock {
+    block_height: i64,
+    rewards:      Vec<PreparedUpdateAccountBalance>,
```
limemloh, line 3864: It seems there will only ever be a vector of one element here, so maybe just:
```suggestion
    mint_reward:      PreparedUpdateAccountBalance,
```

limemloh (approved):

backend-rust/src/graphql_api.rs
DOBEN: It is already tracked in the issue:
https://github.com/Concordium/concordium-scan/issues/423

backend-rust/src/indexer.rs
DOBEN: I added `Excluding CCD mint rewards.`

PR #428, opened 2025-01-24
Introduce basic migration logic

limemloh:
## Purpose

Introduce migration logic into the `ccdscan-indexer` binary, allowing us to run migrations from rust.
`ccdscan-api` now ensures the database schema version is compatible and `ccdscan-indexer` ensures it is the latest before running.

## Changes

- Change the `DATABASE_URL` environment variable for both binaries to allow for disabling compile-time checked queries while running the service and to follow the naming convention.
- Update the `.env.template` with more settings and comments
- Update readme with new instructions on how to run and introduce migrations.
- Removed default value for node endpoint CLI arg


backend-rust/src/migrations/m0001-initialize.sql
```diff
@@ -0,0 +1,782 @@
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
lassemand, line 1: Why do we have this file? It seems to be pretty much the same as what is already there.

I assume it at the very least means we have to remove the old ones?

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -60,6 +64,9 @@ async fn main() -> anyhow::Result<()> {
         .connect(&cli.database_url)
         .await
         .context("Failed constructing database connection pool")?;
+    // Ensure the database schema is compatible with supported schema version.
+    migrations::ensure_compatible_schema_version(&pool, SUPPORTED_SCHEMA_VERSION).await?;
```
lassemand, line 68: Just do the same migration here instead.

During it this way we have an annoying situation where we are dependent on the api is initialising after the migration has occurred on the indexer which is a pretty annoying feature to implement deployment wise. 

At least it cannot be done as is, since it happens application side.

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
+use anyhow::Context;
```
lassemand, line 1: Instead of having schema versions, why dont we just ensure that the sql is the same as what is defined within the .sql files?

backend-rust/src/bin/ccdscan-api.rs
limemloh: I really don't think this should be doing migrations.

Most database migrations require changes to the indexer, so if the API starts doing migrations, it might crash it.
New columns need to be populated by the indexer, before the API should start using them anyway.
Removing columns would crash the indexer.

We also don't want several migrations to be running at once, so since we only run one indexer instance, this is probably a better fit.

backend-rust/src/migrations.rs
limemloh: Not sure, I understand what you are suggesting.
If you mean ask the database for the schema and then compare with the sql files, it seems very complex solution when we start altering tables and remove stuff.

backend-rust/src/migrations/m0001-initialize.sql
limemloh: Yes, the old one needs to be removed and I will do that right before merging.
Merging in changes to the old file from main, is just simpler when I keep this around.

backend-rust/README.md
```diff
@@ -141,41 +164,42 @@ Given that one wants follow the logs of the database:
...
-To introduce a new one run:
+To setup the database schema when developing the service run:
 
 ```
-sqlx migrate add '<description>'
+cargo run --bin ccdscan-indexer -- --migrate-only
 ```
 
-where `<description>` is replaced by a short description of the nature of the migration.
+NOTE: Having compile-time checked queries will cause issues, since the queries are invalid until the database have been properly migrated. This is done by _not_ having the `DATABASE_URL` environment variable set until after running the migrations or using:
```
DOBEN, line 175: This `Note` could be added to the ` ccdscan-api` service further up as well with the blow cli command as a work around since it can happen in the  ` ccdscan-api` service as well.

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
...
+
+/// Represents database schema versions.
+/// Whenever migrating the database a new variant should be introduced.
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, derive_more::Display)]
+#[repr(i64)]
+pub enum SchemaVersion {
+    #[display("0000:Empty database with no tables yet.")]
+    Empty            = 0,
+    #[display(
+        "0001:Initial schema version with tables for blocks, transaction, accounts, contracts, \
```
DOBEN, line 127:
```suggestion
        "0001:Initial schema version with tables for blocks, transactions, accounts, contracts, \
```

backend-rust/src/bin/ccdscan-api.rs
lassemand: Currently you are not making changes to the indexer. This is database relevant only. 

As is it will cause failures to the deployment because the SQL schema will be incompatible. The simple solution is to do what I suggested above.

An alternative could be to provide the information about whether the api schema has been applied, and then provide such information. This could for example be in the health endpoint.

backend-rust/src/migrations.rs
lassemand: Basically this relates quite heavily to the other conversation so I am closing this.

backend-rust/src/bin/ccdscan-api.rs
lassemand: A simpler approach yet is to just keep retrying until it is valid

backend-rust/Dockerfile
```diff
@@ -20,9 +19,7 @@ WORKDIR /usr/app
 RUN apt-get update && \
     apt-get -y install ca-certificates
 
-COPY migrations /usr/app/migrations
```
lassemand, line 23: We cannot move the sql away from the image

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
...
+         tokens and more."
+    )]
+    InitialFirstHalf = 1,
```
DOBEN, line 130:
```suggestion
    InitialFirstHalf,
```

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    /// Parse version number into a database schema version.
+    /// None if the version is unknown.
+    fn from_version(version: i64) -> Option<SchemaVersion> {
+        match version {
+            0 => Some(SchemaVersion::Empty),
+            1 => Some(SchemaVersion::InitialFirstHalf),
+            _ => None,
+        }
+    }
```
DOBEN, line 144: Implementing the `try_from` trait  might be more common than a custom conversion function.

```suggestion
impl TryFrom<i64> for SchemaVersion {
    type Error = anyhow::Error;

    fn try_from(version: i64) -> Result<Self, Self::Error> {
        match version {
            0 => Ok(SchemaVersion::Empty),
            1 => Ok(SchemaVersion::InitialFirstHalf),
            _ => anyhow::bail!("Unknown database schema version"),
        }
    }
}
```

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
...
+            SchemaVersion::InitialFirstHalf => unimplemented!(
+                "No migration implemented for database schema version {}",
+                self.as_i64()
+            ),
+        };
+        let end_time = chrono::Utc::now();
+        insert_migration(&mut tx, &new_version.into(), start_time, end_time).await?;
+        tx.commit().await?;
+        Ok(new_version)
+    }
```
DOBEN, line 179:
```suggestion
  fn next_version(self) -> anyhow::Result<SchemaVersion> {
        (self.as_i64() + 1).try_into()
    }

    /// Run migrations for this schema version to the next.
    async fn migration_to_next(&self, pool: &PgPool) -> anyhow::Result<SchemaVersion> {
        let mut tx = pool.begin().await?;
        let start_time = chrono::Utc::now();
        match self {
            SchemaVersion::Empty => {
                // Set up the initial database schema.
                tx.as_mut()
                    .execute(sqlx::raw_sql(include_str!("./migrations/m0001-initialize.sql")))
                    .await?;
            }
            SchemaVersion::InitialFirstHalf => unimplemented!(
                "No migration implemented for database schema version {}",
                self.as_i64()
            ),
        };
        let new_version = self.next_version()?;
        let end_time = chrono::Utc::now();
        insert_migration(&mut tx, &new_version.into(), start_time, end_time).await?;
        tx.commit().await?;
        Ok(new_version)
    }
```

limemloh: Being explicit about the next version allowed to implement migrations which "skips" other database migrations. Say we implement a new feature which require traversing blocks again in a migration and then release this on stagenet, then we find a bug or some missing information and we need to do another migration and schema version fixing it traversing the chain again.
We can now provide a migration which skips the faulty version and only travers once for testnet and mainnet.

backend-rust/src/migrations.rs
DOBEN: ok, it makes also sense to group them into the implementation trying to minimize the occurance that we `forget` to update the function in the future when introducing new schemaVersions.

backend-rust/.env.template
```diff
@@ -1,2 +1,16 @@
-PGPASSWORD=$DB_PASSWORD
-DATABASE_URL=postgres://postgres:$DB_PASSWORD@localhost/ccd-scan
+PGPASSWORD=${DB_PASSWORD}
+
+# gRPC interface of the node.
+CCDSCAN_INDEXER_GRPC_ENDPOINTS=http://localhost:20000
```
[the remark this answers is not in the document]
limemloh: I am not sure the public nodes support everything we need, but maybe we should extend both services with a `--dotenv <filename>` arg and provide `.env.stagetnet`, `.env.testnet` and `.env.mainnet` just like we do for the frontend.
We can do this in another PR

backend-rust/README.md
```diff
@@ -27,6 +27,26 @@ For instructions on how to use the indexer run:
 ccdscan-indexer --help
 ```
 
+<!-- TODO When service become stable: add documentation of arguments and environment variables. -->
+
+## Database schema setup and migrations
+
+To set up the database schema either from an empty database or migration from an older release of `ccdscan-indexer` run:
```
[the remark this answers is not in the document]
limemloh: I assume you mean using `sqlx` in which case I don't think it belongs here, since this section is for users of the service and they probably don't have the `sqlx-cli`.
But I'll add the sqlx command to the dev section.

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
...
+impl SchemaVersion {
+    /// The latest known version of the schema.
+    const LATEST: SchemaVersion = SchemaVersion::InitialFirstHalf;
+
+    /// Parse version number into a database schema version.
+    /// None if the version is unknown.
+    fn from_version(version: i64) -> Option<SchemaVersion> {
+        match version {
+            0 => Some(SchemaVersion::Empty),
+            1 => Some(SchemaVersion::InitialFirstHalf),
```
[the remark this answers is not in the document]
limemloh: We can automate the implementation of this function, but did not do it since it require adding a dependency, but if this is a real concern to us, then I will add it, meaning we get compile-time errors for everything else when a new schema version is introduced.

backend-rust/src/bin/ccdscan-api.rs
limemloh: This was discussed offline, so resolving this.
Let me know if you disagree

backend-rust/src/scalar_types.rs
```diff
@@ -3,7 +3,6 @@ use async_graphql::{scalar, InputValueError, InputValueResult, Scalar, ScalarTyp
 
 pub type Amount = UnsignedLong;
 pub type TokenId = String;
-pub type RewardPeriodLength = i64;
```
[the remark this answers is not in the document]
limemloh: I seems still to be on main https://github.com/Concordium/concordium-scan/blob/625dfff715f873b4285457f5f5dfc5b98932ad15/backend-rust/src/scalar_types.rs#L6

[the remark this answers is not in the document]
DOBEN: yes, exactly. What I mean was that this `removeOfALine` will not happen after you rebase.

backend-rust/.env.template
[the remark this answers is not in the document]
DOBEN: We discussed this and public nodes seem to support what we need for now.

backend-rust/README.md
```diff
@@ -13,56 +13,81 @@ The service is split to allow for running several instances of the GraphQL API a
...
+### Configuring
+
+The indexer service have several options for configuration, and these can be provided as command-line arguments or/and environment variables, where the command-line arguments take precedence.
+By default the service also reads environment variables a `.env` directory if present in the current directory of the running process.
+
+The file `.env.template` in this project repository can be used as a starting point.
+
+The required configurations are:
+
+- `--database-url <url>` (env `CCDSCAN_INDEXER_DATABASE_URL=<url>`): where `<url>` is the database connection in the format of a URL ex. `postgres://user:passwork@localhost/ccdscan`.
```
DOBEN, line 33:
```suggestion
- `--database-url <url>` (env `CCDSCAN_INDEXER_DATABASE_URL=<url>`): where `<url>` is the database connection in the format of a URL ex. `postgres://user:password@localhost/ccdscan`.
```

backend-rust/README.md
```diff
@@ -13,56 +13,81 @@ The service is split to allow for running several instances of the GraphQL API a
...
+
+The GraphQL API service have several options for configuration, and these can be provided as command-line arguments or/and environment variables, where the command-line arguments take precedence.
+By default the service also reads environment variables a `.env` directory if present in the current directory of the running process.
+
+The file `.env.template` in this project repository can be used as a starting point.
+
+The required configurations are:
 
-## Run the GraphQL API Service during development
+- `--database-url <url>` (env `CCDSCAN_API_DATABASE_URL=<url>`): where `<url>` is the database connection in the format of a URL ex. `postgres://user:passwork@localhost/ccdscan`.
```
DOBEN, line 78:
```suggestion
- `--database-url <url>` (env `CCDSCAN_API_DATABASE_URL=<url>`): where `<url>` is the database connection in the format of a URL ex. `postgres://user:password@localhost/ccdscan`.
```

backend-rust/README.md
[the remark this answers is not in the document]
DOBEN: I was thinking of providing an example command like below. Can be discussed for a follow up PR.
```
docker run -p 5432:5432 -e POSTGRES_PASSWORD=password -e POSTGRES_DB="ccdscan" --rm postgres
```

DOBEN (approved):
LGTM

backend-rust/README.md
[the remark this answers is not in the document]
limemloh: I have added this command in the dev section :+1:

backend-rust/src/migrations.rs
```diff
@@ -0,0 +1,264 @@
...
+    Empty            = 0,
+    #[display(
+        "0001:Initial schema version with tables for blocks, transaction, accounts, contracts, \
+         tokens and more."
+    )]
+    InitialFirstHalf = 1,
+}
+impl SchemaVersion {
+    /// The latest known version of the schema.
+    const LATEST: SchemaVersion = SchemaVersion::InitialFirstHalf;
```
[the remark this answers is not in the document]
limemloh: I'll leave this for now.
If this becomes an issue, we can address it

PR #430, opened 2025-01-27
Query node status

lassemand:
## Purpose

Implement query node status in mem

## Changes

- Introduce a background task that periodically fetches the state of the node collector backend.
- The task will run continuously to ensure the most up-to-date information is maintained in memory.

### Update Health Endpoint

- Enhance the /health endpoint to provide information about whether the node status information has been successfully fetched externally

### Multiple Sources of Truth

- Each pod has its own source of truth.
- During redeployment, the system must wait for the first fetch from the node collector backend to ensure the in-memory state is populated.

### Redis Considerations

While Redis could be used to ensure consistency between pods, it is deemed unnecessary for this implementation:

### Reasons:

- The list of node statuses is mutable and can be modified by any user.
- Paging requests can still encounter modifications between the first call and the last, regardless of Redis.

### Approach:

In-memory state will suffice, as eventual consistency is acceptable for this use case.

## Checklist

      hard-to-understand areas.


backend-rust/src/graphql_api/node_status.rs
```diff
@@ -0,0 +1,497 @@
...
+                NodeSortField::Uptime => a.uptime.partial_cmp(&b.uptime),
+            };
+
+            match sort_direction {
+                NodeSortDirection::Asc => ordering.unwrap_or(Equal),
+                NodeSortDirection::Desc => ordering.unwrap_or(Equal).reverse(),
+            }
+        });
+
+        let after_cursor_index = if let Some(after_cursor) = after {
```
lassemand, line 69: Just use https://github.com/Concordium/concordium-scan/pull/432/files#r1932205040 instead

backend-rust/src/graphql_api.rs
```diff
@@ -148,6 +151,8 @@ pub struct ApiServiceConfig {
     token_holder_addresses_collection_limit: u64,
     #[arg(long, env = "CCDSCAN_API_CONFIG_TOKEN_EVENTS_COLLECTION_LIMIT", default_value = "100")]
     token_events_collection_limit: u64,
+    #[arg(long, env = "CCDSCAN_API_CONFIG_NODE_STATUSES_CONNECTION_LIMIT", default_value = "100")]
+    node_statuses_collection_limit: u64,
```
limemloh, line 155: Since we have all of it in memory, this limit is not really protecting the service from anything, but it does not hurt either

backend-rust/src/graphql_api/node_status.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub finalization_period_emsd: Option<f64>,
+    pub consensus_baker_id: Option<u64>,
+    pub blocks_received_count: Option<u64>,
+}
+
+#[ComplexObject]
+impl NodeStatus {
+    async fn id(&self) -> types::ID { types::ID::from(self.node_id.clone()) }
+
+    async fn client_version(&self) -> String { self.client.to_string() }
```
limemloh, line 223:
```suggestion
    async fn client_version(&self) -> &String { &self.client }
```

backend-rust/src/graphql_api/node_status.rs
```diff
@@ -0,0 +1,497 @@
...
+                NodeSortField::ClientVersion => a.client.partial_cmp(&b.client),
+                NodeSortField::ConsensusBakerId => {
+                    a.consensus_baker_id.partial_cmp(&b.consensus_baker_id)
+                }
+                NodeSortField::FinalizedBlockHeight => {
+                    a.finalized_block_height.partial_cmp(&b.finalized_block_height)
+                }
+                NodeSortField::NodeName => a.node_name.partial_cmp(&b.node_name),
+                NodeSortField::PeersCount => a.peers_count.partial_cmp(&b.peers_count),
+                NodeSortField::Uptime => a.uptime.partial_cmp(&b.uptime),
```
limemloh, line 60: Really minor thing here, and it probably does not matter so feel free to ignore, but we could use `cmp` and [`f64::total_cmp`](https://doc.rust-lang.org/std/primitive.f64.html#method.total_cmp) to avoid the `unwrap_or` later

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -92,6 +127,8 @@ async fn main() -> anyhow::Result<()> {
 
     let (subscription, subscription_listener) =
         graphql_api::Subscription::new(cli.database_retry_delay_secs);
+    let (block_sender, block_receiver) = tokio::sync::watch::channel(None);
+    let block_receiver_health = block_receiver.clone();
```
limemloh, line 131: The naming of these channel endpoints seems a bit off, right? :sweat_smile: 
These are about the node status

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -155,26 +214,39 @@ async fn main() -> anyhow::Result<()> {
...
     // Ensure all tasks have stopped
-    let _ = tokio::join!(monitoring_task, queries_task, pgnotify_listener);
+    let _ = tokio::join!(monitoring_task, queries_task, pgnotify_listener, node_collector_task);
     Ok(())
 }
 
+#[derive(Clone)]
+struct HealthState {
+    pool:                 sqlx::PgPool,
+    node_status_receiver: Receiver<Option<Vec<NodeStatus>>>,
```
limemloh, line 234: Doc comments would be good here

limemloh, line 234: Consider introducing a type alias `type Receiver = Receiver<Option<Vec<NodeStatus>>` in `graphql_api/node_status.rs`.

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -155,26 +214,39 @@ async fn main() -> anyhow::Result<()> {
...
-            "database": "connected"
-        })),
-        Err(err) => axum::extract::Json(json!({
-            "status": "error",
-            "database": format!("not connected: {}", err)
-        })),
-    }
+    let node_status_connected = state.node_status_receiver.borrow().is_some();
+    let database_connected =
+        migrations::ensure_compatible_schema_version(&state.pool, SUPPORTED_SCHEMA_VERSION)
```
limemloh, line 244: The error message here contains information of why it is failing, so maybe it would be useful to keep it.
But it depends a bit on how the health endpoint is used, I guess.

limemloh (approved):
LGTM, only some minor comments.

backend-rust/src/graphql_api.rs
lassemand: I will remove it. It was actually also the plan, since I have the Content-Length check

backend-rust/src/bin/ccdscan-api.rs
```diff
@@ -53,6 +54,35 @@ struct Cli {
     /// Non-zero exit code is returned when incompatible.
     #[arg(long, env = "CCDSCAN_API_CHECK_DATABASE_COMPATIBILITY_ONLY")]
     check_database_compatibility_only: bool,
+    #[arg(
+        long,
+        env = "CCDSCAN_API_NODE_COLLECTOR_BACKEND_ORIGIN",
+        default_value = "https://dashboard.stagenet.concordium.com"
```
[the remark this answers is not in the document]
lassemand: It needs to be set no matter what. Sure I can use the .env.template to achieve the same thing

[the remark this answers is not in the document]
limemloh: I would prefer an error if a user forgot to set it, instead of silently getting stagenet data.
The readme lists the required CLI arguments, so this should be on that list, (only database url currently on that list)

backend-rust/src/bin/ccdscan-api.rs
lassemand: It is basically solely for the kubernetes control plane. It is not even externally exposed.

So relevant errors should be logged.

PR #431, opened 2025-01-27
Add baker transactions query

DOBEN:
## Purpose

https://linear.app/concordium/issue/CCD-65/implement-bakertransactions-and-accountbaker


backend-rust/src/graphql_api/baker.rs
```diff
@@ -137,11 +147,118 @@ impl Baker {
...
+                    type_account as "type_account: AccountTransactionType",
+                    type_credential_deployment as "type_credential_deployment: CredentialDeploymentTransactionType",
+                    type_update as "type_update: UpdateTransactionType",
+                    success,
+                    events as "events: sqlx::types::Json<Vec<Event>>",
+                    reject as "reject: sqlx::types::Json<TransactionRejectReason>"
+                FROM transactions
+                WHERE transactions.sender_index = $5
+                AND index > $1 AND index < $2
+                AND type_account IN ('AddBaker', 'RemoveBaker', 'UpdateBakerStake',
```
lassemand, line 197: Where is this list obtained from?

Can we somehow map enum types into this list?
And if not could we reuse the names provided as part of TransactionType? I assume it would be those which are used to generate the type.

lassemand (approved):

backend-rust/src/graphql_api/baker.rs
lassemand: The advantage of doing it this way is that if someone renames the enums, then it will still work.

DOBEN: The enum is is a subset of the all possible `account_transaction_type`:

https://github.com/Concordium/concordium-scan/blob/main/backend-rust/src/migrations/m0001-initialize.sql#L3

limemloh: compile-time checked queries would catch this

backend-rust/src/graphql_api/baker.rs
```diff
@@ -137,11 +147,124 @@ impl Baker {
...
+            query.desc,
+            query.limit,
+            self.id.0,
+            &[
+                AccountTransactionType::AddBaker,
+                AccountTransactionType::RemoveBaker,
+                AccountTransactionType::UpdateBakerStake,
+                AccountTransactionType::UpdateBakerRestakeEarnings,
+                AccountTransactionType::UpdateBakerKeys,
+                AccountTransactionType::ConfigureBaker
```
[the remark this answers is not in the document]
DOBEN: True. On the other hand, I don't expect most bakers to re-use the same account for sending frequent transactions on chain (having different keys for different purposes seems a wise way to operate - e.g. one baker account (key is well secured and not exposed often), another account for sending frequent transactions on-chain (key is online often and potentially compromised)).

Writing out that list is more explicit and it is immediately clear which transaction types are returned by that query. People might start re-ordering the enum in the future and it is hard to track if we use ranges at various places.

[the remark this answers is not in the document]
limemloh: We should of course document this for the enum. Postgres does not allow reordering, so it would mean introducing a new enum during the migration which hopefully would result in a compile error here.
I expect this to perform well enough, so I fine with being explicit.
If this become slow we can investigate the impact

limemloh (approved):

PR #432, opened 2025-01-28
Index and expose block special events

limemloh:
## Purpose

Index block special events and implement `Block::special_events` API call

### DIFF
--- a/backend-rust/src/block_special_event.rs
+++ b/backend-rust/src/block_special_event.rs
@@ -1,0 +1,421 @@
+use crate::{
+    address::AccountAddress,
+    graphql_api::{ApiError, ApiResult},
+    scalar_types::{Amount, BakerId},
+};
+use async_graphql::{connection, types, Object, SimpleObject, Union};
+use concordium_rust_sdk::{
+    common::types as sdk_common_types, id::types as sdk_id_types, types::SpecialTransactionOutcome,
+};
+
+#[derive(Union, serde::Serialize, serde::Deserialize)]
+#[allow(clippy::enum_variant_names)]
+pub enum SpecialEvent {
+    MintSpecialEvent(MintSpecialEvent),
+    FinalizationRewardsSpecialEvent(FinalizationRewardsSpecialEvent),
+    BlockRewardsSpecialEvent(BlockRewardsSpecialEvent),
+    BakingRewardsSpecialEvent(BakingRewardsSpecialEvent),
+    PaydayAccountRewardSpecialEvent(PaydayAccountRewardSpecialEvent),
+    BlockAccrueRewardSpecialEvent(BlockAccrueRewardSpecialEvent),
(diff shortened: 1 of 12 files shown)


backend-rust/src/block_special_event.rs
```diff
@@ -0,0 +1,421 @@
...
+                ..
+            } => Self::ValidatorSuspended,
+            O::ValidatorPrimedForSuspension {
+                ..
+            } => Self::ValidatorPrimedForSuspension,
+        }
+    }
+}
+
+fn connection_from_slice<A: async_graphql::OutputType>(
```
lassemand, line 384: This does not have to be specific to blocks.

I should use this from the nodes collector querying.

backend-rust/src/block_special_event.rs
```diff
@@ -0,0 +1,421 @@
...
+    collection: &[A],
+    first: Option<usize>,
```
lassemand, line 386: This should be u64 so as to avoid having different preconditions for the caller based on the underlying architecture.

backend-rust/src/block_special_event.rs
```diff
@@ -0,0 +1,421 @@
...
+    if first.is_some() && last.is_some() {
+        return Err(ApiError::QueryConnectionFirstLast);
+    }
+    let mut start: usize = if let Some(after) = after {
+        after.parse::<usize>()? + 1
+    } else {
+        0
+    };
+    let mut end: usize = if let Some(before) = before {
+        before.parse::<usize>()?
```
lassemand, line 400: should be min with the length

backend-rust/src/block_special_event.rs
lassemand: I like this function. Gj

backend-rust/src/block_special_event.rs
```diff
@@ -0,0 +1,421 @@
...
+    BlockAccrueRewardSpecialEvent(BlockAccrueRewardSpecialEvent),
+    PaydayFoundationRewardSpecialEvent(PaydayFoundationRewardSpecialEvent),
+    PaydayPoolRewardSpecialEvent(PaydayPoolRewardSpecialEvent),
+    ValidatorSuspended(ValidatorSuspended),
+    ValidatorPrimedForSuspension(ValidatorPrimedForSuspension),
+}
+
+impl SpecialEvent {
+    pub fn from_special_transaction_outcome(
+        block_height: i64,
```
lassemand, line 28: Why is this converted into a i64? Is this to fit it into the sql types?

lassemand (approved):
Looks good to me

backend-rust/src/block_special_event.rs
limemloh: The slice `&[A]` itself is enforcing these preconditions, so it won't change by using an u64 here.

PR #438, opened 2025-01-29
Provide option to point to different .env file

limemloh:
## Purpose

Add `--dovenv <file>`  option to both binaries, useful when running the service for several environments.


DOBEN (approved):
After adding the port it works.

PR #439, opened 2025-01-29
TLS and timeouts for endpoints when fetching genesis data

limemloh:
## Purpose

Fix issue when fetching genesis information from node endpoint depending on TLS.

## Changes

We already handle TLS, but not until the main indexer start running, so this PR move the handling earlier so it applies for the genesis data fetching as well.

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -122,6 +122,28 @@ SELECT height FROM blocks ORDER BY height DESC LIMIT 1
         .await?
         .map(|r| r.height);
 
+        // Handle TLS configuration and set timeouts according to the configuration for
+        // every endpoint.
+        let endpoints: Vec<v2::Endpoint> = endpoints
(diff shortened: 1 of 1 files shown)


backend-rust/src/indexer.rs
```diff
@@ -122,6 +122,28 @@ SELECT height FROM blocks ORDER BY height DESC LIMIT 1
         .await?
         .map(|r| r.height);
 
+        // Handle TLS configuration and set timeouts according to the configuration for
```
DOBEN, line 125: Can we move this above `last_height_stored` fetched, since `start_height` below is related to it.

DOBEN (approved):

DOBEN:
Suggestion: we could add some clarification to the error that the connection error is from the `node` and not the `databsae`.
```
Error: transport error

Caused by:
    0: error trying to connect: tls handshake eof
    1: tls handshake eof
```

PR #449, opened 2025-01-30
Migration issues

lassemand:
## Purpose

We need to return a none zero exit code given migration issues occurs


backend-rust/src/bin/ccdscan-indexer.rs
```diff
@@ -105,8 +105,7 @@ async fn main() -> anyhow::Result<()> {
             },
             result = &mut migration_task => {
                 if let Err(err) = result? {
-                    error!("Migration error: {}", err);
-                    return Ok(())
+                    return Err(anyhow::format_err!("Migration error: {}", err))
```
lassemand, line 108: I assume you want a custom error here, however I have no idea which one.

limemloh (approved):

PR #456, opened 2025-01-31
Initial setup for Query::bakers query

limemloh:
## Purpose

Implement `Query::bakers`


backend-rust/src/indexer.rs
```diff
@@ -2269,8 +2358,23 @@ impl PreparedBakerEvent {
             PreparedBakerEvent::RemoveDelegation {
                 delegator_id,
             } => {
+                // Update total pool stake of old pool (if not the passive pool).
+                sqlx::query!(
+                    "UPDATE bakers
+                     SET pool_total_staked = pool_total_staked - accounts.delegated_stake
```
DOBEN, line 2364: Also update the `pool_delegation_count` value.

backend-rust/src/migrations/m0007-cumulate-pool-info.sql
```diff
@@ -0,0 +1,46 @@
...
+    WHERE
+        pool_delegator_count IS NULL
+        AND pool_total_staked IS NULL;
+
+-- Mark new columns as NOT NULL
+ALTER TABLE bakers
+    ALTER COLUMN pool_total_staked SET NOT NULL,
+    ALTER COLUMN pool_delegator_count SET NOT NULL,
+    ADD CONSTRAINT check_pool_total_staked_positive CHECK (pool_total_staked > 0);
+
```
DOBEN, line 40: Good to add to find bugs in our accounting:

 ADD CONSTRAINT check_pool_delegator_count_not_negative CHECK (pool_delegator_count >= 0);

backend-rust/src/indexer.rs
```diff
@@ -1846,8 +1849,20 @@ impl PreparedAccountDelegationEvent {
                 account_id,
                 staked,
             } => {
+                // Update total stake of the pool first.
```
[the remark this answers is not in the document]
lassemand: So the reason for the later part (`- accounts.delegated_stake`) is that because we can only delegate to one node? Meaning if we then set delegation the old delegation will be removed? In such case why do we only remove the delegation from the targeted baker? Shouldnt we remove from any baker in which this account is currently delegating to?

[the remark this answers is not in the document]
lassemand: Ahh. So stake decrease/increase is strictly a change on the given node. So in practice, since we can only delegate to a single node, one could actually remove `bakers.id = accounts.delegated_target_baker_id AND` since it must be the case that this delegation belongs to this baker?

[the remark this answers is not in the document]
lassemand: Ahh and` accounts.delegated_target_baker_id` is there to fix passive delegation. :)

backend-rust/src/indexer.rs
```diff
@@ -4084,23 +4188,38 @@ impl RestakeEarnings {
...
+                             SET pool_total_staked = pool_total_staked + $2
+                         WHERE id = $1",
+                        pool,
+                        self.amount,
+                    )
+                    .execute(tx.as_mut())
+                    .await?;
+                }
+            }
+        } else {
```
lassemand, line 4214:
```suggestion
        if let Some(true) = account_row.delegated_restake_earnings {
              // Restake is enabled.
              if let Some(pool) = account_row.delegated_target_baker_id {
                  // delegating to a pool (and not the passive pool).
                  sqlx::query!(
                      "UPDATE bakers
                           SET pool_total_staked = pool_total_staked + $2
                       WHERE id = $1",
                      pool,
                      self.amount,
                  )
                  .execute(tx.as_mut())
                  .await?;
              }
        } else {
```

lassemand: Also I would probably just name let Some(pool) = account_row.delegated_target_baker_id to 

`let restake_enabled  = let Some(pool) = account_row.delegated_target_baker_id`

lassemand: I would also merge the two if statements, but I am fine with both

backend-rust/src/indexer.rs
```diff
@@ -4084,23 +4188,38 @@ impl RestakeEarnings {
...
+                        "UPDATE bakers
+                             SET pool_total_staked = pool_total_staked + $2
+                         WHERE id = $1",
+                        pool,
+                        self.amount,
+                    )
+                    .execute(tx.as_mut())
+                    .await?;
+                }
+            }
```
lassemand, line 4213:
```suggestion
            if {
                if restake && let Some(pool) = account_row.delegated_target_baker_id {
                  // delegating to a pool (and not the passive pool).
                  sqlx::query!(
                      "UPDATE bakers
                           SET pool_total_staked = pool_total_staked + $2
                       WHERE id = $1",
                      pool,
                      self.amount,
                  )
                  .execute(tx.as_mut())
                  .await?;
            }
```

backend-rust/src/indexer.rs
```diff
@@ -4084,23 +4188,38 @@ impl RestakeEarnings {
...
             self.amount
         )
         .fetch_one(tx.as_mut())
         .await?;
-        // When delegated_restake_earnings is None the account is not delegating, so it
-        // might be baking.
-        if account_row.delegated_restake_earnings.is_none() {
+        if let Some(restake) = account_row.delegated_restake_earnings {
+            // Account is delegating.
+            if restake {
```
lassemand, line 4199: Why is it relevant that we are restaking?

If we are not restaking, wouldnt a change to the state also have to occur?

backend-rust/src/migrations/m0007-cumulate-pool-info.sql
lassemand: I also wonder how it has to be positive given that we default to 0?

backend-rust/src/indexer.rs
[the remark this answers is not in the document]
lassemand: So in practice we could instead also check `accounts.delegated_target_baker_id != null`, (not saying we should)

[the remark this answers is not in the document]
limemloh: @DOBEN This works in the `DelegationEvent::Added` case even if the target got updated to something not null first, since the `accounts.delegated_stake` would be `0`.

@lassemand Each account can only delegate to one pool at a time, so yes the change in stake is for that target pool. The query is updating the total stake of the pool you are currently delegating to (which might be NULL due to it being the passive pool or NULL because you just added the delegation and this is not yet set)
you should understand the line:
```
SET pool_total_staked = pool_total_staked + $1 - accounts.delegated_stake
```
as add the new current stake of the delegator to the pool (`+ $1`) and remove what the previous delegated stake (`- accounts.delegated_stake`).

backend-rust/src/indexer.rs
limemloh: Yes, great catch!

backend-rust/src/indexer.rs
limemloh: This database operation `RestakeEarnings` only represents the stake changes and not the change in balance.
When restake earnings is enabled, the rewards are put as staked immediately, so we update the pool total stake. If restake earnings is not enabled we don't have to do any changes for this operation.

backend-rust/src/indexer.rs
limemloh: `if <condition> && let ...` is not valid syntax in rust, but I can change it to:
```rust
if let (true, Some(pool)) = (restake, account_row.delegated_target_baker_id) {
    ...
}
```
Removes one indentation level, so maybe better :shrug:

backend-rust/src/migrations/m0007-cumulate-pool-info.sql
limemloh: `pool_total_staked` does not default to `0`, it is a column in table `bakers` and for a row to be present in that table you must have staked some amount and the pool total stake is at least the bakers staked amount.

DOBEN (approved):
LGTM

lassemand (approved):

PR #459, opened 2025-02-01
Add index for blocks without cumulative time for indexer performance

limemloh:
## Purpose

Due to missing index for the recently added cumulative_finaliziation_time on blocks in the database schema, the processing of blocks starts slowing down as the block table grows.
This PR introduces a new database schema version with the index added.

Also, changed the log level for log of preprocessing block started, from INFO to DEBUG.


backend-rust/src/migrations.rs
```diff
@@ -201,7 +207,15 @@ impl SchemaVersion {
                     .await?;
                 SchemaVersion::InitialFirstHalf
             }
-            SchemaVersion::InitialFirstHalf => unimplemented!(
+            SchemaVersion::InitialFirstHalf => {
+                tx.as_mut()
+                    .execute(sqlx::raw_sql(include_str!(
+                        "./migrations/m0002-block-cumulative-time-index.sql"
```
DOBEN, line 213:
```suggestion
                        "./migrations/m0002-block-cumulative-fin-time-index.sql"
```

DOBEN (approved):

PR #460, opened 2025-02-02
Search function blocks

lassemand:
## Purpose

Add search function on blocks

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend-rust/src/graphql_api.rs
+++ b/backend-rust/src/graphql_api.rs
@@ -896,14 +896,93 @@ impl SearchResult {
 
     async fn blocks(
         &self,
-        #[graphql(desc = "Returns the first _n_ elements from the list.")] _first: Option<i32>,
+        ctx: &Context<'_>,
+        #[graphql(desc = "Returns the first _n_ elements from the list.")] first: Option<u64>,
         #[graphql(desc = "Returns the elements in the list that come after the specified cursor.")]
-        _after: Option<String>,
-        #[graphql(desc = "Returns the last _n_ elements from the list.")] _last: Option<i32>,
+        after: Option<String>,
+        #[graphql(desc = "Returns the last _n_ elements from the list.")] last: Option<u64>,
         #[graphql(desc = "Returns the elements in the list that come before the specified cursor.")]
-        _before: Option<String>,
+        before: Option<String>,
     ) -> ApiResult<connection::Connection<String, Block>> {
-        todo_api!()
+        let account_address_regex: Regex = Regex::new(r"^[a-fA-F0-9]{1,64}$")
+            .map_err(|_| ApiError::InternalError("Invalid regex".to_string()))?;
+        let pool = get_pool(ctx)?;
+        let config = get_config(ctx)?;
+        let query =
+            ConnectionQuery::<i64>::new(first, after, last, before, config.block_connection_limit)?;
+        let mut connection = connection::Connection::new(false, false);
+        if !account_address_regex.is_match(&self.query) {
+            return Ok(connection);
+        }
+        let param = self.query.to_lowercase();
+        let rows = sqlx::query_as!(
+            block::Block,
+            "SELECT * FROM (
+                SELECT
+                    hash,
+                    height,
+                    slot_time,
+                    block_time,
+                    finalization_time,
+                    baker_id,
+                    total_amount
+                FROM blocks
+                WHERE
+                    height = $5
+                    OR hash LIKE $6 || '%'
+                    AND height > $1
+                    AND height < $2
+                ORDER BY
+                    (CASE WHEN $4 THEN height END) ASC,
+                    (CASE WHEN NOT $4 THEN height END) DESC
+                LIMIT $3
+            ) ORDER BY height DESC",
+            query.from,
+            query.to,
+            query.limit,
+            query.desc,
+            param.parse::<i64>().ok(),
+            param
+        )
+        .fetch_all(pool)
+        .await?;
+
+        let mut min_index = None;
+        let mut max_index = None;
+        for block in rows {
+            min_index = Some(match min_index {
+                None => block.height,
+                Some(current_min) => min(current_min, block.height),
+            });
+
+            max_index = Some(match max_index {
+                None => block.height,
+                Some(current_max) => max(current_max, block.height),
+            });
+            connection.edges.push(connection::Edge::new(block.height.to_string(), block));
+        }
+
+        if let (Some(page_min_id), Some(page_max_id)) = (min_index, max_index) {
+            let result = sqlx::query!(
+                r#"
+                    SELECT MAX(height) as max_height, MIN(height) as min_height
+                    FROM blocks
+                    WHERE
+                        height = $1
+                        OR hash LIKE $2 || '%'
+                "#,
+                param.parse::<i64>().ok(),
+                param,
+            )
+            .fetch_one(pool)
+            .await?;
+
+            connection.has_previous_page =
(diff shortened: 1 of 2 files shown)


backend-rust/src/graphql_api.rs
```diff
@@ -896,14 +896,93 @@ impl SearchResult {
 
     async fn blocks(
```
DOBEN, line 897: Changelog please.

backend-rust/src/graphql_api.rs
```diff
@@ -896,14 +896,93 @@ impl SearchResult {
...
-        _after: Option<String>,
-        #[graphql(desc = "Returns the last _n_ elements from the list.")] _last: Option<i32>,
+        after: Option<String>,
+        #[graphql(desc = "Returns the last _n_ elements from the list.")] last: Option<u64>,
         #[graphql(desc = "Returns the elements in the list that come before the specified cursor.")]
-        _before: Option<String>,
+        before: Option<String>,
     ) -> ApiResult<connection::Connection<String, Block>> {
-        todo_api!()
+        let account_address_regex: Regex = Regex::new(r"^[a-fA-F0-9]{1,64}$")
```
DOBEN, line 907:
```suggestion
        let block_hash_regex: Regex = Regex::new(r"^[a-fA-F0-9]{1,64}$")
```

backend-rust/src/graphql_api.rs
```diff
@@ -896,14 +896,93 @@ impl SearchResult {
...
+                    height,
+                    slot_time,
+                    block_time,
+                    finalization_time,
+                    baker_id,
+                    total_amount
+                FROM blocks
+                WHERE
+                    height = $5
+                    OR hash LIKE $6 || '%'
```
DOBEN, line 932: We probably need an index for that to efficiently search for this combination since the `hash` is not part of an any index so far.

backend-rust/src/graphql_api.rs
```diff
@@ -896,14 +896,93 @@ impl SearchResult {
...
+            .map_err(|_| ApiError::InternalError("Invalid regex".to_string()))?;
+        let pool = get_pool(ctx)?;
+        let config = get_config(ctx)?;
+        let query =
+            ConnectionQuery::<i64>::new(first, after, last, before, config.block_connection_limit)?;
+        let mut connection = connection::Connection::new(false, false);
+        if !account_address_regex.is_match(&self.query) {
+            return Ok(connection);
+        }
+        let param = self.query.to_lowercase();
```
DOBEN, line 917: Suggestion: It would be nice to give it a suitable name.

```suggestion
        let search_string = self.query.to_lowercase();
```

backend-rust/src/graphql_api.rs
lassemand: Also it has to be a more advanced index, since it is a search query on a string. Indexing on the string values provides limited value here.

lassemand: Did that on accounts as well

DOBEN: Fine, if we postpone the index creation for PR once we are done with the other search result implementations.

backend-rust/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Database schema version: 2
 ### Changed
 
 - Make the `Query::transaction_metrics` use fixed buckets making it consistent with behavior of the old .NET backend.
+- Implement SearchResult::bakers and add relevant index to database
```
DOBEN, line 17:
```suggestion
- Implement `SearchResult::bakers` and add relevant index to database
```

DOBEN (dismissed):

backend-rust/src/migrations/m0002-block-cumulative-fin-time-index.sql
```diff
@@ -3,3 +3,6 @@
 CREATE INDEX blocks_height_null_cumulative_finalization_time ON blocks (height)
     WHERE blocks.cumulative_finalization_time IS NULL
     AND blocks.finalization_time IS NOT NULL;
+-- Used to efficiently perform partial string matching on the `hash` column,
+-- allowing fast lookups when searching for blocks by their hash prefix using `LIKE`.
+CREATE INDEX blocks_hash_gin_trgm_idx ON blocks USING gin(hash gin_trgm_ops);
```
limemloh, line 8: gin seems a bit heavy for just a prefix search, but it is probably okay.
Mostly concerned about the indexing time here.

backend-rust/src/graphql_api.rs
```diff
@@ -905,14 +905,93 @@ impl SearchResult {
...
+                    height,
+                    slot_time,
+                    block_time,
+                    finalization_time,
+                    baker_id,
+                    total_amount
+                FROM blocks
+                WHERE
+                    height = $5
+                    OR hash LIKE $6 || '%'
```
limemloh, line 941: Mintor thing, but the `starts_with` function is a bit more readable
```suggestion
                    OR starts_with(hash, $6)
```

limemloh (dismissed):

backend-rust/src/graphql_api.rs
```diff
@@ -905,14 +905,92 @@ impl SearchResult {
...
+                        height = $1
+                        OR starts_with(hash, $2)
+                "#,
+                lower_case_query.parse::<i64>().ok(),
+                lower_case_query,
+            )
+            .fetch_one(pool)
+            .await?;
+            connection.has_previous_page =
+                result.max_height.map_or(false, |db_max| db_max > page_max_height);
```
lassemand, line 989: We are moving from big to small here. 

This means that we have to reverse it all

DOBEN (approved):

backend-rust/src/graphql_api.rs
```diff
@@ -905,14 +905,92 @@ impl SearchResult {
...
+            });
+
+            max_height = Some(match max_height {
+                None => block.height,
+                Some(current_max) => max(current_max, block.height),
+            });
+            connection.edges.push(connection::Edge::new(block.height.to_string(), block));
+        }
+
+        if let (Some(page_min_height), Some(page_max_height)) = (min_height, max_height) {
```
limemloh, line 974: Minor suggestion: We can take the first and last edge instead, something like this:
```suggestion
        for block in rows {
            connection.edges.push(connection::Edge::new(block.height.to_string(), block));
        }

        if let (Some(page_min_height), Some(page_max_height)) = (connection.edges.last(), connection.edges.first()) {
```

backend-rust/src/graphql_api.rs
```diff
@@ -905,14 +905,92 @@ impl SearchResult {
...
+                LIMIT $3
+            ) ORDER BY height DESC",
+            query.from,
+            query.to,
+            query.limit,
+            query.desc,
+            lower_case_query.parse::<i64>().ok(),
+            lower_case_query
+        )
+        .fetch_all(pool)
```
limemloh, line 956: The advantage of `fetch` over `fetch_all` is that we can avoid allocating the Vec first.
Minimal optimization, but results in the same amount of code, so why not.

limemloh (approved):

backend-rust/src/graphql_api.rs
lassemand: This cannot be done. We are altering the order based on whether last has been set or not.

limemloh: No we are not, the ordering of the page must be the same, so if this is the case it is a bug

PR #465, opened 2025-02-03
Improve version reporting

DOBEN:
## Purpose

- Add `database_schema_version` and `api_supported_database_schema_version` to `versions` endpoint.
- Fix typo in `versions` endpoint.

### DIFF
--- a/backend-rust/src/graphql_api.rs
+++ b/backend-rust/src/graphql_api.rs
@@ -24,6 +24,7 @@ pub(crate) use todo_api;
 
 use crate::{
     connection::ConnectionQuery,
+    migrations::{self, current_schema_version},
     scalar_types::{BlockHeight, DateTime, TimeSpan, UnsignedLong},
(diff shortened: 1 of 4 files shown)


backend-rust/src/graphql_api.rs
```diff
@@ -63,6 +64,11 @@ use tracing::{error, info};
 use transaction::Transaction;
 
 const VERSION: &str = clap::crate_version!();
+/// The minimum supported database schema version for the API.
+/// Fails at startup if any breaking database schema versions have been
+/// introduced since this version.
+pub const SUPPORTED_SCHEMA_VERSION: migrations::SchemaVersion =
+    migrations::SchemaVersion::InitialFirstHalf;
```
limemloh, line 71: Maybe move this to the `migrations.rs` file instead, like the `SchemaVersion::LATEST` const, then we have all of it together.

limemloh (approved):

PR #474, opened 2025-02-05
Fix prev and next on contracts

lassemand:
## Purpose

Underlying task: https://linear.app/concordium/issue/CCD-166/contract-prev-and-next-page

When desc is set using this means that is indeed desc and not something we interpolate in the caller code. 

We can migrate a single endpoint at a time, because when ascending_default_order is set we are backwards compatible.


backend-rust/src/connection.rs
```diff
@@ -100,24 +114,27 @@ impl<A> ConnectionQuery<A> {
...
 
         let limit =
             first.or(last).map_or(connection_limit, |limit| connection_limit.min(limit)) as i64;
-
         Ok(Self {
             from,
             to,
             limit,
-            desc: last.is_some(),
+            desc: !(last.is_some() ^ ascending_default_order),
```
limemloh, line 137: It seems to me that this would be changing the meaning of this variable, which would break the other places it is used. The new meaning fits better with the name, but it is probably better to rename it to better fit what it means instead of changing it everywhere.

backend-rust/src/connection.rs
```diff
@@ -100,24 +114,27 @@ impl<A> ConnectionQuery<A> {
...
         };
 
         let to = if let Some(b) = before {
             b.parse::<A>().map_err(|e| e.into())?
-        } else {
+        } else if ascending_default_order {
             A::MAX
+        } else {
+            A::MIN
         };
```
limemloh, line 129: Since `A` is generic we can just use another type where `MIN` and `MAX` are switched around.
Something like:

```
struct DescI64(i64);

impl ConnectionCursor for DescI64 {
    const MIN = i64::MAX;
    const MAX = i64::MIN;
}
```

lassemand: I think this is a pretty odd way of doing it. We already have a variable indicating whether this is descending or ascending. Lets not have two ways in a single function to indicate the same thing

limemloh: The `query.desc` variable is really just describing whether the query was asking for `last` and not `first`, which is why the naming is bad.
The generic `A` represents the cursor, which doesn't really necessarily have a notion of ascending or descending, but so far we are mostly using `i64` as the cursor and this does have that notion.
For more complex cursors this would not be the case.

backend-rust/src/connection.rs
```diff
@@ -69,17 +69,20 @@ impl ConnectionCursor for i64 {
...
     /// If the `last` elements are requested instead of the `first` elements
     /// (indicated by the `last` key being set when creating a new
     /// `ConnectionQuery`), the edges/nodes should be ordered in reverse
     /// (DESC) order before applying the range. This allows the range from
     /// `from` to `to` to be applied starting from the last element.
-    pub desc:  bool,
+    /// TODO: Migrate away from this and use descending instead
+    pub desc:       bool,
+    /// Reflects whether the query should be descending or ascending
+    pub descending: bool,
```
limemloh, line 85: This is not about the connection itself being descending, but rather the subquery should be descending or ascending based on whether last/first was provided. The connection page can be descending while this field is true.

backend-rust/src/connection.rs
lassemand: I believe your point is: We completely remove the new variable and the extra constructor, and instead checks on whether A is of type DescI64 in the bottom of the function as well, and then avoid the extra constructor. That at least sounds ok to me, however it was not super clear in the initial proposal.

limemloh: I have implemented what I mean in [this PR](https://github.com/Concordium/concordium-scan/pull/479/files#diff-995eb760e8b294e181df7486e1f3a89eb8adef6efa89b39dc1a8fa1a0b0f1ef2R72), since I needed something similar

lassemand: I actually like this solution more. 

I find it pretty ugly that we are doing ASC queries when desc is true. Since this solution seems to have been forced upon me, I will adjust I guess.

limemloh: I understand why you dislike it, but to me the issue is more the naming of these variables are misleading.
`ConnectionCursor::MIN` and `ConnectionCursor::MAX` should rather be named `ConnectionCursor::StartBound` and `ConnectionCursor::EndBound` and `query.desc` should be named `query.is_last`.

lassemand: That is indeed another possibility. Can I change that in another PR?

DOBEN (approved):

PR #475, opened 2025-02-06
Contract rejected event

lassemand:
## Purpose

Task: https://linear.app/concordium/issue/CCD-169/contract-rejected-event-size


limemloh (approved):
LGTM assuming the changelog is updated

PR #481, opened 2025-02-06
Return null for contract update message when parameter is empty

### LINKED ISSUE #472: Error display when parsing the input parameter/contractEvents fails
The new Rust backend now reports the exact `error` (when parsing the input parameter/contractEvents fails with the schema). The front-end has no error component for that case yet but some improvements can be done without too much front-end work:

* Button to display the error in a pop-up should work (or not displayed). It should be possible to see the full error message.
* The text should contain `Error` to distinguish the case from correctly parsed parameters/contractEvents.<br><br><br>https://stagenet.ccdscan.io/contracts?dcount=1&dentity=contract&dcontractAddressIndex=0&dcontractAddressSubIndex=0

Versions: Rust-backend 0.1.19, frontend 1.7.0<br><br>/home/db/Desktop/errors/Screenshot from 2025-02-03 23-03-07.png

limemloh:

### DIFF
--- a/backend-rust/CHANGELOG.md
+++ b/backend-rust/CHANGELOG.md
@@ -26,6 +26,8 @@ Database schema version: 2
 - Issue for `Query::transaction_metrics` producing an internal error when query period is beyond the genesis block.
 - Next and previous page on contracts
 - Contract rejected event skips in the correct way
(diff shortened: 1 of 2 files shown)


DOBEN (approved):
Good catch of the issue. The original issue seems still relevant (not fixed by the PR) in case an actual schema parsing error appears.

PR #482, opened 2025-02-06
Fix events being of wrong order

lassemand:
## Purpose

https://linear.app/concordium/issue/CCD-167/contract-events-wrong-order


limemloh (approved):
LGTM, don't forget the changelog

PR #500, opened 2025-02-10
Fix misleading variable names in connection.rs

limemloh:
## Purpose

Rename some variables in `connection.rs` which are named differently from how these are used.
This should not change any semantics of the code.

### DIFF
--- a/backend-rust/src/connection.rs
+++ b/backend-rust/src/connection.rs
@@ -55,14 +55,19 @@ pub fn connection_from_slice<T: AsRef<[A]>, A: async_graphql::OutputType + Clone
     Ok(connection)
 }
 
-/// Upper and lower limits for the Cursor in a GraphQL Cursor Connection.
+/// Bounds for the Cursor in a GraphQL Cursor Connection, used as the fallback
+/// when no explicit range is provided as `after`/`before`.
 pub trait ConnectionCursor {
-    const MIN: Self;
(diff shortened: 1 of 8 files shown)


backend-rust/src/connection.rs
```diff
@@ -72,26 +77,26 @@ impl ConnectionCursor for i64 {
...
-    pub to:    A,
+    pub to:      A,
     /// The limit to use for the SQL query.
-    pub limit: i64,
+    pub limit:   i64,
     /// If the `last` elements are requested instead of the `first` elements
     /// (indicated by the `last` key being set when creating a new
     /// `ConnectionQuery`), the edges/nodes should be ordered in reverse
     /// (DESC) order before applying the range. This allows the range from
     /// `from` to `to` to be applied starting from the last element.
```
DOBEN, line 98:
```suggestion
    /// `ConnectionQuery`), the edges/nodes should be ordered in reverse
    /// order to the returned order of the connection before applying the range. This allows the range from
    /// `from` to `to` to be applied starting from the last element.
```

Similar comments exist on some of the queries using this boolean.

DOBEN (approved):

backend-rust/src/connection.rs
limemloh: We can fix them as we find them :smile:

PR #511, opened 2025-02-11
Fix date

lassemand:
## Purpose

We are not using the same date across multiple calls to the database.


backend-rust/src/graphql_api/block_metrics.rs
```diff
@@ -171,7 +174,7 @@ SELECT
     ) AS y_finalization_time_avg_s,
     bucket_last_block.total_staked AS y_last_total_micro_ccd_staked
 FROM
-    date_bin_series($2::INTERVAL, NOW() - $1::INTERVAL, NOW()) AS bucket
+    date_bin_series($2::INTERVAL, $3::TIMESTAMPTZ - $1::INTERVAL, $3) AS bucket
```
lassemand, line 177: Basically we are calculating based on different limits. Now they are fixed and should be fine.

It does not matter what the limit are so long as they are the same.

limemloh (approved):
An alternative would be to use an SQL transaction for the queries.

also don't forget the changelog

Issue #518, opened 2025-02-12
Display `blockHeight` on the detail block page.


DOBEN:
**Task description**

Display `blockHeight` on the detail page of a block.

E.g.:
https://stagenet.ccdscan.io/blocks?dcount=1&dentity=block&dhash=a5d256fe5585981895d55d88b5c79a370f02468c03705dec36a517a27efc80d2

PR #519, opened 2025-02-12
Add block height display

DOBEN:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/518

## Changes

Add `blockHeight` display at block details page.


lassemand (approved):

frontend/src/components/BlockDetails/BlockDetailsHeader.vue
```diff
@@ -15,6 +15,16 @@
 					icon-size="h-5 w-5"
 					tooltip-class="font-sans"
 				/>
+				<h1 class="inline-block text-2xl numerical">
+					{{ String(block?.id) }}
+				</h1>
+				<TextCopy
+					:text="String(block.id)"
```
limemloh, line 22: I think this should be using `block.blockHeight` instead of `block.id`.
The ID type in graphql is not meant to convey information, but is simply an identifier for the resource.

limemloh: Also the .NET backend does not use block height as `block.id`

limemloh (approved):

Issue #521, opened 2025-02-13
Transaction page to block issue


lassemand:
When opening a block from a transaction we have to emit the hash of the block and not the id.

PR #522, opened 2025-02-13
Remove id from query

lassemand:
## Purpose

When going from transaction to block we are getting an error. This remediates that by no longer be using id.

Closing https://github.com/Concordium/concordium-scan/issues/521


frontend/src/components/BlockDetails/BlockDetailsContainer.vue
```diff
@@ -36,15 +36,12 @@ const paginationVars = {
 }
 
 type Props = {
-	id?: string
 	hash?: string
```
limemloh, line 39: We need the hash be present now, so it should not be optional
```suggestion
	hash: string
```

frontend/src/queries/useBlockQuery.ts
```diff
@@ -97,22 +77,15 @@ const BlockQueryByHash = gql<BlockByBlockHashResponse>`
...
-			hash?: Ref<string>
-	  }
-	| {
-			hash: Ref<string>
-			id?: Ref<string>
-	  }
-) & {
+type QueryParams = {
+	hash?: Ref<string>
+} & {
```
limemloh, line 82: `hash` is no longer optional
```suggestion
	hash: Ref<string>
```

frontend/src/components/BlockDetails/BlockDetailsContainer.vue
```diff
@@ -36,15 +36,12 @@ const paginationVars = {
 }
 
 type Props = {
```
limemloh, line 38: Somewhere we must be calling `<BlockDetailsContainer :id=... >` which should also change and we need to make sure that the hash is used then.

DOBEN: Alternatively, we can replace `id` with `blockHeight` and use that as identity.

lassemand: Why would we want to do that? I do not understand why we would prefer to be using two different identifiers across the same frontend.

DOBEN: The `id` and `blockHeight` contain the same value. Just replaceing the name should cause the least change to be done. As a second step, I agree we should either only use the `blockHeight` or only the `blockHash` to uniquely identity a block at the front-end as much as possible.

limemloh: `id` is not the same as block height for the `.NET` backend, hence the bug we are seing.
Plus we should not rely on any internal details of `id` for the rust backend, since the contract between the frontend and API for `id` is 'just some identifier' and not a particular identifier.

lassemand: I will remove it all the way through, however just to warn it is a rather substantial change.

lassemand: But imo also a lot more cleaner

frontend/src/components/molecules/BlockLink.vue
```diff
@@ -1,5 +1,5 @@
 ﻿<template>
-	<div v-if="props.hash || props.id" class="inline-block whitespace-nowrap">
+	<div v-if="props.hash" class="inline-block whitespace-nowrap">
```
limemloh, line 2: I guess this if statement is always true now:
```suggestion
	<div class="inline-block whitespace-nowrap">
```

frontend/src/components/molecules/BlockLink.vue
```diff
@@ -43,7 +42,7 @@ const emitBlur = (newTarget: FocusEvent) => {
 	emit('blur', newTarget)
 }
 
-const handleOnClick = (hash?: string, id?: string) => {
-	if (hash || id) drawer.push({ entityTypeName: 'block', hash: hash || '', id })
+const handleOnClick = (hash: string) => {
+	if (hash) drawer.push({ entityTypeName: 'block', hash: hash })
```
limemloh, line 46: If statement should always be true now:
```suggestion
	drawer.push({ entityTypeName: 'block', hash: hash })
```

frontend/src/composables/useDrawer.ts
```diff
@@ -10,7 +10,7 @@ import { useState } from '#app'
 
 type BlockDrawerItem = {
 	entityTypeName: 'block'
-} & ({ id: string; hash?: string } | { hash: string; id?: string })
+} & { hash: string }
```
limemloh, line 13:
```suggestion
    hash: string
}
```

frontend/src/composables/useDrawer.ts
```diff
@@ -38,7 +38,7 @@ type ModuleDrawerItem = {
  */
 type BakerDrawerItem = {
 	entityTypeName: 'baker'
-	id: number
+	hash: string
 }
```
limemloh, line 42: This is looks like a mistake, I would expect a baker to be associated with an id

frontend/src/composables/useDrawer.ts
```diff
@@ -94,14 +94,18 @@ export const isItemOnTop = (
...
 	)
 		return !!(
 			(item.hash && item.hash === currentTopItem.value.hash) ||
 			(item.id && item.id === currentTopItem.value.id)
 		)
+	if (
+		item.entityTypeName === 'block' &&
+		item.entityTypeName === currentTopItem.value.entityTypeName
+	)
+		return !!(item.hash && item.hash === currentTopItem.value.hash)
```
limemloh, line 108: Since `hash` is always defined:
```suggestion
		return !!(item.hash === currentTopItem.value.hash)
```

frontend/src/queries/useBlockQuery.ts
```diff
@@ -97,25 +73,18 @@ const BlockQueryByHash = gql<BlockByBlockHashResponse>`
...
-			hash?: Ref<string>
-	  }
-	| {
-			hash: Ref<string>
-			id?: Ref<string>
-	  }
-) & {
+type QueryParams = {
+	hash: Ref<string>
+} & {
```
limemloh, line 78:
```suggestion
```

frontend/CHANGELOG.md
```diff
@@ -4,10 +4,16 @@ All notable changes to this project will be documented in this file.
...
 
+## [1.7.4] - 2025-02-13
+
 ### Added
 
 - Add `blockHeight` at block details page.
 
+### Changed
+
+- Remove id from blocks
```
limemloh, line 15:
```suggestion
- Always query block by block hash and never by ID.
```

limemloh (approved):

frontend/src/composables/useDrawer.ts
lassemand: This is not true. Item is a generic item working on all types of events.

DOBEN (approved):

PR #529, opened 2025-02-13
Handle move of delegators as pool get closed or removed.

limemloh:
## Purpose

Fix missing handling of moving delegators to passive pool as their target pool got closed or removed.
Fix invalid bakers in table, due to `DelegationEvent::BakerRemoved` not being handled.


backend-rust/src/indexer.rs
```diff
@@ -1917,12 +1917,87 @@ impl PreparedAccountDelegationEvent {
...
+            remove_baker:    RemoveBaker::prepare(baker_id)?,
+            move_delegators: MovePoolDelegatorsToPassivePool::prepare(baker_id)?,
+        })
+    }
+
+    async fn save(
+        &self,
+        tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+    ) -> anyhow::Result<()> {
+        self.remove_baker.save(tx).await?;
```
lassemand, line 1946: Not that it matters since it is a transaction, but would probably be more natural to reverse the order of these two.

backend-rust/src/indexer.rs
```diff
@@ -2038,10 +2111,19 @@ impl PreparedBakerEvent {
             BakerEvent::BakerSetOpenStatus {
                 baker_id,
                 open_status,
-            } => PreparedBakerEvent::SetOpenStatus {
-                baker_id:    baker_id.id.index.try_into()?,
-                open_status: open_status.to_owned().into(),
-            },
+            } => {
+                let open_status = open_status.to_owned().into();
+                let move_delegators = if matches!(open_status, BakerPoolOpenStatus::ClosedForAll) {
```
lassemand, line 2116: Why are we not doing this when setting the BakerPoolOpenStatus to ClosedForAll instead?

DOBEN (approved):
LGTM after already mentioned comments form Lasse.

Issue #520, opened 2025-02-13
Last total micro ccd unlocked


lassemand:
Implement last total micro ccd unlocked

PR #532, opened 2025-02-14
Remove locked ccd metrics from fungy backend

lassemand:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/520

Dependent on https://github.com/Concordium/concordium-scan/pull/531 to be rolled out to all environments first.

### DIFF
--- a/backend-rust/src/graphql_api/block_metrics.rs
+++ b/backend-rust/src/graphql_api/block_metrics.rs
@@ -31,10 +31,6 @@ struct BlockMetrics {
     /// The total CCD Released. This is total CCD supply not counting the
     /// balances of non circulating accounts.
     last_total_micro_ccd_released: Amount,
-    /// The current total CCD released according to the Concordium promise
-    /// published on deck.concordium.com. Will be null for blocks with slot
-    /// time before the published release schedule.
-    last_total_micro_ccd_unlocked: Option<Amount>,
     /// The current total amount of CCD staked.
     last_total_micro_ccd_staked: Amount,
     buckets: BlockMetricsBuckets,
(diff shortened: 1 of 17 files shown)


backend-rust/src/graphql_api/block_metrics.rs
```diff
@@ -31,10 +31,6 @@ struct BlockMetrics {
     /// The total CCD Released. This is total CCD supply not counting the
     /// balances of non circulating accounts.
     last_total_micro_ccd_released: Amount,
-    /// The current total CCD released according to the Concordium promise
-    /// published on deck.concordium.com. Will be null for blocks with slot
-    /// time before the published release schedule.
-    last_total_micro_ccd_unlocked: Option<Amount>,
```
DOBEN, line 37: please also consider the `last_total_micro_ccd_unlocked` in `block_metrics`.

DOBEN (approved):

backend-rust/src/graphql_api/block_metrics.rs
```diff
@@ -236,8 +232,6 @@ LEFT JOIN LATERAL (
                 .total_amount_released
                 .unwrap_or(0)
                 .try_into()?,
-            last_total_micro_ccd_unlocked: None, // TODO implement unlocking schedule
```
limemloh, line 239: Does the frontend still query for this field? because then we need to fix that as well otherwise it will break

DOBEN: The dependency was removed in this PR `https://github.com/Concordium/concordium-scan/pull/531`. This means that we need to roll out the changes to the frontend first and then update api services.

lassemand: I have rolled out the frontend

PR #535, opened 2025-02-14
Improve performance get source module

### LINKED ISSUE #534: `get_module_source` query on the most recent block instead of the current block being indexed
**Task description**

The `get_module_source` query on old blocks are currently not performing well in the node (issue is investigated in the node and will be addressed). An easy workaround to speed up the indexer for mainnet that we consider is to use a caught-up node and query the value at the most recent block.

`.get_module_source(&module_reference, BlockIdentifier::AbsoluteHeight(latest_block_height))`

https://github.com/Concordium/concordium-scan/blob/03258762e6d5071cb5ec15adb5cd6bbbe190fd5e/backend-rust/src/indexer.rs#L2271

DOBEN:
## Purpose

Closes #534

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -2265,8 +2265,8 @@ impl PreparedModuleDeployed {
         node_client: &mut v2::Client,
         module_reference: sdk_types::hashes::ModuleReference,
     ) -> anyhow::Result<Self> {
-        // The `get_module_source`` query on old blocks are currently not performing
-        // well in the node, We query on the `lastFinal` block here as a result (https://github.com/Concordium/concordium-scan/issues/534).
+        // The `get_module_source` query on old blocks are currently not performing
+        // well in the node. We query on the `lastFinal` block here as a result (https://github.com/Concordium/concordium-scan/issues/534).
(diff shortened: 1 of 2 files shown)


lassemand (approved):

backend-rust/src/indexer.rs
```diff
@@ -2263,13 +2263,14 @@ struct PreparedModuleDeployed {
...
         node_client: &mut v2::Client,
-        data: &BlockData,
         module_reference: sdk_types::hashes::ModuleReference,
     ) -> anyhow::Result<Self> {
-        let block_height = data.finalized_block_info.height;
-
+        // The `get_module_source` query on old blocks are currently not performing
+        // well in the node. We query on the `lastFinal` block here as a result (https://github.com/Concordium/concordium-scan/issues/534).
+        // The connected node should be caught-up to the top of the chain to make use of
+        // the performance gain.
```
td202, line 2271: I think whether the node is caught up is not really relevant. As long as the module was in some finalized block, then looking up in the last finalized block is going to be the fastest way to get it. (I don't think we are concerned with non-finalized blocks here at all.)

td202 (approved):

PR #538, opened 2025-02-17
Implement SearchResult Transactions

lassemand:
## Purpose

Move search results into separate file
Finish search result transactions


lassemand:
Changelog and indexing

backend-rust/src/migrations/m0005.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX transactions_hash_idx ON transactions USING gin (hash text_pattern_ops);
```
DOBEN, line 1: Please add a comment why we introduce this index:

-- Used to efficiently perform partial string matching on the `hash` column,
-- allowing fast lookups when searching for transactions by their hash prefix using `LIKE`.

DOBEN (approved):

backend-rust/src/graphql_api/search_result.rs
```diff
@@ -0,0 +1,370 @@
...
+                    type_credential_deployment as "type_credential_deployment: CredentialDeploymentTransactionType",
+                    type_update as "type_update: UpdateTransactionType",
+                    success,
+                    events as "events: sqlx::types::Json<Vec<Event>>",
+                    reject as "reject: sqlx::types::Json<TransactionRejectReason>"
+                FROM transactions
+                WHERE
+                    starts_with(hash, $5)
+                    AND $2 < index
+                    AND index < $1
```
DOBEN, line 192: Not sure if searching by `transaction_index` is something that makes sense (or is used by people). The old backend does not search by index so would keep that behavior.

limemloh (approved):

backend-rust/src/graphql_api/search_result.rs
lassemand: This is not entirely true. It does not provide the cursors, however it is indeed true that I highly doubt that this is a useable feature. To be consistent I would like to keep it as is.

PR #545, opened 2025-02-18
Migrate to be storing base address of an account

lassemand:
## Purpose

When using an aliass we do not detect the account of origin.
This is the first step towards achieving this, namely to store the canonical address of an account. 
I will make a follow up PR on this task which uses this new column on the required indexing tasks.


backend-rust/src/migrations/m0007-canonical-address-and-transaction-search-index.sql
```diff
@@ -0,0 +1,9 @@
+-- Initial setup of canonical address with placeholder data
+ALTER TABLE accounts
+ADD COLUMN canonical_address BYTEA DEFAULT E'\\x00000000000000000000000000000000000000';
```
limemloh, line 3: I don't think it makes sense for the column to have a default, it would prevent errors if forgotten in an insert query. I would prefer for it to be nullable during the migration and then NOT NULL at the end.

backend-rust/src/migrations.rs
```diff
@@ -187,14 +188,16 @@ pub enum SchemaVersion {
...
     FixStakedAmounts,
+    #[display("0007:AccountBaseAddress")]
+    AccountBaseAddress,
 }
 impl SchemaVersion {
     /// The minimum supported database schema version for the API.
     /// Fails at startup if any breaking database schema versions have been
     /// introduced since this version.
-    pub const API_SUPPORTED_SCHEMA_VERSION: SchemaVersion = SchemaVersion::PayDayLotteryPowers;
+    pub const API_SUPPORTED_SCHEMA_VERSION: SchemaVersion = SchemaVersion::FixDanglingDelegators;
```
limemloh, line 198: Don't think the API depends on anything in `FixDanglingDelegators`, so it shold still be `PayDayLotteryPowers`.
```suggestion
    pub const API_SUPPORTED_SCHEMA_VERSION: SchemaVersion = SchemaVersion::PayDayLotteryPowers;
```

backend-rust/src/indexer.rs
```diff
@@ -3507,7 +3512,7 @@ async fn process_cis2_token_event(
 }
 
 struct PreparedScheduledReleases {
-    account_address: Arc<String>,
+    canonical_address: Vec<u8>,
```
limemloh, line 3515: It would be better to use the `CanonicalAddress` type everywhere instead of a `Vec<u8>`, since this can be stack allocated.
Then when using it in a query `.0.as_slice()`

backend-rust/src/migrations.rs
```diff
@@ -252,12 +256,16 @@ impl SchemaVersion {
                 SchemaVersion::PayDayLotteryPowers
             }
             SchemaVersion::PayDayLotteryPowers => {
-                m0005_fix_dangling_delegators::run(&mut tx, endpoints).await?
+                m0005_fix_dangling_delegators::run(&mut tx, endpoints).await?;
+                SchemaVersion::FixDanglingDelegators
```
limemloh, line 260:
```suggestion
                m0005_fix_dangling_delegators::run(&mut tx, endpoints).await?
```

backend-rust/src/indexer.rs
```diff
@@ -797,16 +797,18 @@ async fn save_genesis_data(endpoint: v2::Endpoint, pool: &PgPool) -> anyhow::Res
...
         // because the genesis accounts do not have a creation transaction.
         sqlx::query!(
-            "INSERT INTO accounts (index, address, amount, num_txs)
-            VALUES ($1, $2, $3, 0)",
+            "INSERT INTO accounts (index, address, amount, canonical_address, num_txs)
+            VALUES ($1, $2, $3, $4, 0)",
             index,
             account_address,
             amount,
+            canonical_address.0.to_vec()
```
limemloh, line 811:
```suggestion
            canonical_address.0.as_slice()

File: backend-rust/src/indexer.rs (hunk 3 above)
Comment (lassemand) [reply backend-rust/src/indexer.rs:3515]: Good point

File: backend-rust/src/migrations/m0007-canonical-address-and-transaction-search-index.sql (hunk 1 above)
Comment (lassemand) [reply backend-rust/src/migrations/m0007-canonical-address-and-transaction-search-index.sql:3]: So you would alter it in the migration? Sure.

File: backend-rust/CHANGELOG.md [hunk 6]
```diff
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 Database schema version: 6
 
+### Changed
+
+- Store canonical address as a column on the account
+
 ### Fixed
 
 - Add database migration 6 fixing invalid baker and delegator stake due to missing handling of restake earnings.
```
DOBEN, line 15: The submodule update points to a PR commit that has meanwhile been closed. Please update the connection to an open PR that can be merged long-term.
https://github.com/Concordium/concordium-rust-sdk/pull/230

backend-rust/src/migrations/m0007-canonical-address-and-transaction-search-index.sql
DOBEN: Since canonical addresses have a fixed length (29 bytes), that would be good to enforce:

ADD CONSTRAINT check_canonical_address_length CHECK (LENGTH(canonical_address) = 29);

lassemand: Also I should use UNIQUE and remove the index.

backend-rust/src/indexer.rs
```diff
@@ -3654,8 +3732,12 @@ impl PreparedScheduledReleases {
             block_height,
             AccountStatementEntryType::TransferOut,
         )?;
+        let account_address =
+            concordium_rust_sdk::base::contracts_common::AccountAddress::from_str(
+                target_address.as_str(),
+            )?;
```
limemloh, line 3738: Since we are not storing a string for the account_adress anymore in the various struct, all of current `Arc<String>` types should probably be `CanonicalAccountAddress` and we avoid parsing the string here

limemloh (approved):

Issue #230, opened 2024-09-27
Implement BakerPool::delegators


limemloh:
Implement `BakerPool::delegators` in the Rust GraphQL API and remaining delegation fields in `BakerPool`.

For reference, the [C# GraphQL schema can be found here](https://github.com/Concordium/concordium-scan/blob/main/backend/Tests/Api/GraphQL/\__snapshots_\_/committed-schema.verified.graphql).

PR #547, opened 2025-02-18
Add delegators query

DOBEN:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/230

## Changes

- Add delegators query to `bakerPool` query.


backend-rust/src/graphql_api/baker.rs
```diff
@@ -472,27 +475,135 @@ struct BakerPool<'a> {
...
+}
+
+#[Object]
+impl DelegationSummary {
+    async fn account_address(&self) -> &AccountAddress { &self.account_address }
+
+    async fn staked_amount(&self) -> ApiResult<Amount> {
+        self.staked_amount.try_into().map_err(|_| {
+            ApiError::InternalError(
+                "Staked amount in database should be a valid UnsignedLong".to_string(),
```
lassemand, line 596: amount on accounts is a BitInt which is a signed long. So you can actually cause this error to trigger.

lassemand: A bit of a thought of issue, since not even enough CCD exists to cause this to occur.

lassemand (approved):

backend-rust/src/graphql_api/baker.rs
```diff
@@ -100,11 +102,11 @@ impl Baker {
...
                 baking_commission,
                 finalization_commission,
-                payday_transaction_commission,
-                payday_baking_commission,
-                payday_finalization_commission,
-                payday_lottery_power as lottery_power
-            FROM bakers 
+                payday_transaction_commission as "payday_transaction_commission?",
+                payday_baking_commission as "payday_baking_commission?",
+                payday_finalization_commission as "payday_finalization_commission?",
```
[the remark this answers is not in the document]
DOBEN: They are nullable because of the `LEFT JOIN` condition and sqlx didn't infer them correctly.

```
{
  "data": null,
  "errors": [
    {
      "message": "Internal error (FailedDatabaseQuery): error occurred while decoding column 8: unexpected null; try decoding as an `Option`",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "bakerByBakerId"
      ]
    }
  ]
}
```

Solution: `force nullable`
https://docs.rs/sqlx/latest/sqlx/macro.query.html#force-nullable

PR #548, opened 2025-02-19
Add validation of affected rows for queries

limemloh:
## Purpose

Add validation of affected rows for most mutating queries in the indexer.


backend-rust/src/indexer.rs
```diff
@@ -2367,7 +2394,8 @@ impl PreparedBakerEvent {
                     delegator_id
                 )
                 .execute(tx.as_mut())
-                .await?;
+                .await?
+                .ensure_affected_rows_in_range(0..=1)?;
```
DOBEN, line 2398: Suggestion to add a comment at these `ensure_affected_rows_in_range` when to expect what number of row updates. E.g.
If `delegated_target_baker_id` is NULL (passive pool) we expect 0 row updates and otherwise 1.

DOBEN (approved):

backend-rust/src/indexer/ensure_affected_rows.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Validate number of affected rows are exactly 1 for a query result.
+    fn ensure_affected_one_row(self) -> Result<Self, EnsureAffectedRowsExactError> {
+        self.ensure_affected_rows(1)
+    }
+}
+
+impl EnsureAffectedRows for sqlx::postgres::PgQueryResult {
+    fn ensure_affected_rows(self, expected: u64) -> Result<Self, EnsureAffectedRowsExactError> {
+        let affected = self.rows_affected();
```
lassemand, line 31: You could just call ensure_affected_rows_in_range here

lassemand:
I am torn on this. Since we are not using accounts right now, we are currently not producing any incorrectness.

When this is merged, we will stop indexing in the individual environment as soon as we use alias. So can we enable this checks by a simple boolean?

I want to have this check on the new indexer, but not on the old for example.

backend-rust/src/indexer/ensure_affected_rows.rs
limemloh: I considered that as well, but the error message becomes less nice

lassemand:
Merging this into `lma/CCD-195/account-address-as-base`

Issue #367, opened 2025-01-09
Implement BakerPool


limemloh:
Implement the simple fields of `BakerPool`:

Already done:
```
totalStakePercentage
delegatorCount
totalStake
delegatedStake
lotteryPower
metadataUrl
paydayCommissionRates
delegatedStakeCap
rankingByTotalStake
```

Note that `apy`, `delegators` and `poolRewards` are not part of this task.

For reference, the [C# GraphQL schema can be found here](https://github.com/Concordium/concordium-scan/blob/main/backend/Tests/Api/GraphQL/\__snapshots_\_/committed-schema.verified.graphql).

PR #550, opened 2025-02-20
Add leverage_bound and capital_bound

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-scan/issues/367

## Changes

- Add database migration 10 to store the `leverage_bound` and the `capital_bound` values of the chain.
- Add query `BakerPool::delegatedStakeCap` that considers the leverage and capital bounds to report the delegate stake cap for baker pools.

```
query ($bakerId: Long!) {
  bakerByBakerId(bakerId: $bakerId) {
    state {
      __typename
      ... on ActiveBakerState {
        pool {
          delegatedStakeCap
        }
      }
    }
  }
}

```


backend-rust/src/graphql_api/baker.rs
```diff
@@ -350,6 +352,79 @@ impl Baker {
...
+            let capital_bound_cap_for_pool_denominator: f64 = 1f64 - capital_bound;
+
+            let capital_bound_cap_for_pool: f64 =
+                capital_bound_cap_for_pool_numerator / capital_bound_cap_for_pool_denominator;
+
+            capital_bound_cap_for_pool
+                .floor()
+                .try_into()
+                .map_err(|e: anyhow::Error| ApiError::InternalError(e.to_string()))?
+        };
```
td202, line 424: I would also suggest doing this calculation precisely in `u128`.

Bₚ = floor( (κ * (T - Dₚ) - Cₚ) / (1 - κ) )

If you just multiply the top and bottom by 100,000 then the computation can't overflow a `u128`:

```
  (current_chain_parameters.capital_bound as u128 * (total_stake - delegated_stake_of_pool) as u128 - 100000 * (self.staked as u128))
  / (100000 - current_chain_parameters.capital_bound as u128)
```

backend-rust/src/graphql_api/baker.rs
```diff
@@ -806,3 +884,44 @@ struct CommissionRates {
...
+    /// will no longer be considered as effective stake.
+    ///
+    /// The value is 1 or greater (1 <= leverage_bound).
+    /// The value's numerator and denominator is stored here.
+    ///
+    /// The `leverage bound` ensures that each baker has skin in the game with
+    /// respect to its delegators by providing some of the CCD staked from
+    /// its own funds.
+    leverage_bound_numerator:   i64,
+    leverage_bound_denominator: i64,
```
td202, line 907: Is there a reason for these to be `i64` instead of `u64`?

backend-rust/src/graphql_api/baker.rs
```diff
@@ -806,3 +884,44 @@ struct CommissionRates {
...
+    /// be considered as effective stake.
+    ///
+    /// The capital bound is always greater than 0 (capital_bound > 0). The
+    /// value is stored as a fraction with precision of `1/100_000`. For
+    /// example, a capital bound of 0.05 is stored as 50000.
+    ///
+    /// The `capital_bound` helps maintain network
+    /// decentralization by preventing a single baker from gaining excessive
+    /// power in the consensus protocol.
+    capital_bound:              i64,
```
td202, line 926: Could this also be unsigned?

backend-rust/src/indexer.rs
```diff
@@ -728,24 +728,35 @@ async fn save_genesis_data(endpoint: v2::Endpoint, pool: &PgPool) -> anyhow::Res
...
+                    chain_parameters_v2.pool_parameters.leverage_bound,
+                ),
+                ChainParameters::V1(chain_parameters_v1) => (
+                    chain_parameters_v1.time_parameters.reward_period_length,
+                    chain_parameters_v1.pool_parameters.capital_bound,
+                    chain_parameters_v1.pool_parameters.leverage_bound,
+                ),
+                ChainParameters::V0(_) => unimplemented!(
+                    "Expect the node to have caught up enought for the `reward_period_length`, \
+                     `capital_bound` and `leverage_bound` values being available."
```
td202, line 754:
```suggestion
                    "Expect the node to have caught up enough for the `reward_period_length`, \
                     `capital_bound` and `leverage_bound` values to be available."
```

backend-rust/src/migrations.rs
```diff
@@ -192,14 +193,16 @@ pub enum SchemaVersion {
     AddAccumulatedPoolState,
     #[display("0008:AccountBaseAddress")]
     AccountBaseAddress,
+    #[display("0007:Add delegated stake cap")]
```
td202, line 196: ?
```suggestion
    #[display("0009:Add delegated stake cap")]
```

DOBEN: Rebased on top of main and updated the migration to version `10`.

backend-rust/src/graphql_api/baker.rs
DOBEN: The macro `query_as!` has only the trait implemented to convert a database column type `BIGINT` to a `i64` rust type (the trait to `u64` is not unavailable).

E.g. we use it as follows:
```
      let current_chain_parameters = sqlx::query_as!(
            DelegatedStakeBounds,
            "
            SELECT 
                leverage_bound_numerator,
                leverage_bound_denominator,
                capital_bound 
            FROM current_chain_parameters 
            WHERE id = true
            "
        )
        .fetch_one(pool)
        .await?;
```

But I get your point that you want to ensure/check that these values are constraint to be positive as reported by the node. I added following check just in case:

```
        // The `leverage_bound` and `capital_bound` are Concordium chain parameters that were set adhering
        // to the below constraints to ensure the consensus algorithm works. 
        // We check these constraints here to ensure the values have been saved in this format to the database as expected.
        if current_chain_parameters.leverage_bound_numerator < 0 {
            return Err(ApiError::InternalError(
                "`leverage_bound_numerator` negative in database".to_string(),
            ));
        }
        if current_chain_parameters.leverage_bound_denominator < 0 {
            return Err(ApiError::InternalError(
                "`leverage_bound_denominator` negative in database".to_string(),
            ));
        }
        if current_chain_parameters.leverage_bound_numerator
            / current_chain_parameters.leverage_bound_denominator
            < 1
        {
            return Err(ApiError::InternalError("`leverage_bound` is smaller than 1".to_string()));
        }
        if current_chain_parameters.capital_bound <= 0 {
            return Err(ApiError::InternalError("`capital_bound` is not greater than 0".to_string()));
        }
```

backend-rust/src/migrations/m00010_fill_capital_bound_and_leverage_bound.rs
```diff
@@ -0,0 +1,90 @@
...
+    v2::{self, BlockIdentifier, ChainParameters},
+};
+use sqlx::Executor;
+
+/// Run database migration to fill the new columns `capital_bound` and
+/// `leverage_bound`.
+pub async fn run(
+    tx: &mut Transaction,
+    endpoints: &[v2::Endpoint],
+    next_schema_version: SchemaVersion,
```
limemloh, line 17: We only ever call it with `SchemaVersion::DelegatedStakeCap` so maybe make it a `const` in here

backend-rust/src/scalar_types.rs
```diff
@@ -69,6 +73,30 @@ impl From<concordium_rust_sdk::common::types::Amount> for UnsignedLong {
...
+
+    fn try_from(value: f64) -> Result<Self, Self::Error> {
+        if value < 0.0 {
+            return Err(anyhow::anyhow!(
+                "Failed to convert f64 to `UnsignedLong` scalar type. Actual value is negative: \
+                 {:?}.",
+                value
+            ));
+        }
+
```
limemloh, line 87: I hope we can avoid having this conversion by using `u128` as suggested by @td202, since it loose precision, but in case there is really no other way then we should probably round before truncating the decimals from the casting.
```suggestion
    let value = value.round();
```

backend-rust/src/graphql_api/baker.rs
limemloh: Using `i64` instead of `u64` is mostly because Postgres does not support `u64` directly, and we would need to use the `NUMERIC` data type instead.
We would have an issue if these values are above `i64::MAX` and since these are user input (governance but still) I guess it is possible.
For most columns like balances, stake, baker ids, contract index and so on, it seems okay though as these numbers are unlikely to be above `i64::MAX`.

td202: Most columns are probably OK, but I do feel a little bit uncomfortable with balance and stake (though realistically they are fine for quite a while). I think it could be reasonable with at least some values to have them represented in the DB as signed and cast to unsigned - such as with fractions like this.

One particular case where this will be an issue is with the microCCD:Euro rate. On testnet, it's currently 15899143920943677440 / 52859724503, so the numerator is bigger than `i64::MAX`.

backend-rust/src/graphql_api/baker.rs
```diff
@@ -350,6 +352,79 @@ impl Baker {
...
+        .fetch_one(pool)
+        .await?;
+
+        let leverage_bound: f64 = current_chain_parameters.leverage_bound_numerator as f64
+            / current_chain_parameters.leverage_bound_denominator as f64;
+        let leverage_bound_cap_for_pool: Amount = ((leverage_bound - 1f64) * self.staked as f64)
+            .try_into()
+            .map_err(|e: anyhow::Error| ApiError::InternalError(e.to_string()))?;
+
+        // The value is stored as a fraction with precision of `1/100_000` in the
```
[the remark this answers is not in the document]
DOBEN: The value can be queried from the node via the rust sdk as follows already:

get_pool_info (https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/v2/struct.Client.html#method.get_pool_info) → BakerPoolStatus → ActiveBakerPoolStatus (https://docs.rs/concordium-rust-sdk/latest/concordium_rust_sdk/types/struct.ActiveBakerPoolStatus.html) → delegated_capital_cap

backend-rust/src/migrations/m00010_fill_capital_bound_and_leverage_bound.rs
DOBEN: My idea was to have all the schema transition visible in the `migration_to_next()` function, whereas if we hardcode constants into the migration fails, it is hard to keep track what schema outcome a particular migration has.

backend-rust/src/graphql_api/baker.rs
DOBEN: `microCCD:Euro` rate is not tracked in CCDScan yet but good to keep this in mind.

backend-rust/src/indexer.rs
```diff
@@ -786,11 +797,28 @@ async fn save_genesis_data(endpoint: v2::Endpoint, pool: &PgPool) -> anyhow::Res
...
+                leverage_bound_numerator, 
+                leverage_bound_denominator
+            ) VALUES (true, $1, $2, $3, $4, $5)
+            ON CONFLICT (id) 
+            DO UPDATE SET 
+                epoch_duration = EXCLUDED.epoch_duration,
+                reward_period_length = EXCLUDED.reward_period_length,
+                capital_bound = EXCLUDED.capital_bound,
+                leverage_bound_numerator = EXCLUDED.leverage_bound_numerator,
+                leverage_bound_denominator = EXCLUDED.leverage_bound_denominator;
```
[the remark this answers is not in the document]
DOBEN: Removed the logic out of the `save_genesis_data()` function since it is now handled in the migration file.

backend-rust/src/migrations/m00010_fill_capital_bound_and_leverage_bound.rs
limemloh: okay, but it is a bit inconsistent now.
Anyway, this is not something blocking.

backend-rust/src/migrations.rs
```diff
@@ -194,14 +195,16 @@ pub enum SchemaVersion {
     AccountBaseAddress,
     #[display("0009:StakedPoolSizeConstraint")]
     StakedPoolSizeConstraint,
+    #[display("00010:Add delegated stake cap")]
```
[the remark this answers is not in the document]
limemloh: Seems to still be relevant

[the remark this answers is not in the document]
DOBEN: fixed

backend-rust/src/graphql_api/baker.rs
```diff
@@ -350,6 +352,124 @@ impl Baker {
...
+        let leverage_bound_cap_for_pool: Amount = leverage_bound_cap_for_pool.into();
+
+        // Calculating the `capital_bound_cap`
+
+        let capital_bound: u128 = current_chain_parameters.capital_bound as u128;
+
+        let capital_bound_cap_for_pool: Amount = if capital_bound == 100_000u128 {
+            // To avoid dividing by 0 in the `capital bound cap` formula,
+            // we only apply the `leverage_bound_cap_for_pool` in that case.
+            leverage_bound_cap_for_pool
```
td202, line 450: I would be inclined to use `u64::MAX.into()` instead, because the capital cap should be independent of the leverage cap.

backend-rust/src/scalar_types.rs
```diff
@@ -69,6 +73,30 @@ impl From<concordium_rust_sdk::common::types::Amount> for UnsignedLong {
...
+        } else {
+            Err(anyhow::anyhow!(
+                "Failed to convert f64 to `UnsignedLong` scalar type. Actual value `{:?}` is too \
+                 large.",
+                value
+            ))
+        }
+    }
+}
+
```
td202, line 99: I guess this isn't needed any more.

td202 (approved):

backend-rust/src/graphql_api/baker.rs
```diff
@@ -350,6 +352,124 @@ impl Baker {
...
+            ));
+        }
+
+        // Calculating the `leverage_bound_cap`
+
+        // To reduce loss of precision, the value is computed in u128.
+        let leverage_bound_cap_for_pool_numerator: u128 =
+            (current_chain_parameters.leverage_bound_numerator
+                - current_chain_parameters.leverage_bound_denominator) as u128
+                * self.staked as u128;
```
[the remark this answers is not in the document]
limemloh: Well the name makes it sound more like 
```suggestion
        let leverage_bound_cap_for_pool_numerator: u128 =
            current_chain_parameters.leverage_bound_numerator as u128
                * self.staked as u128;
```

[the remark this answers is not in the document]
DOBEN: The transformation is as follows:

```
        // Transformation applied to the `leverage_bound_cap_for_pool` formula:
        //
        // `leverage_bound_cap_for_pool`
        // = (λ – 1) * Cₚ
        // = (leverage_bound_numerator / leverage_bound_denominator – 1) * Cₚ
        // = (leverage_bound_numerator / leverage_bound_denominator – (leverage_bound_denominator / leverage_bound_denominator)) * Cₚ
        // = (leverage_bound_numerator – leverage_bound_denominator) / leverage_bound_denominator) * Cₚ
        // = (leverage_bound_numerator – leverage_bound_denominator) * Cₚ / leverage_bound_denominator
        //
        // WHERE
        // λ is the leverage bound
        // Cₚ is the equity capital (staked by the pool owner excluding delegated stake
        // to the pool) of pool p
        // `leverage_bound_numerator` is the value as stored in the database
        // `leverage_bound_denominator` is the value as stored in the database

```

PR #551, opened 2025-02-24
Fix constraint on poolsize and removal of duplication on affected accounts

lassemand:
## Purpose

Fix constraint on poolsize and removal of duplication on affected accounts


backend-rust/src/migrations/m0009_pool_info_constraint.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE bakers
```
limemloh, line 1: It would be good to add a comment with some context, so we don't reintroduce this again in the future.
```suggestion
-- Migration loosening the constraint enforcing `pool_total_staked > 0` to `pool_total_staked >= 0`, since this is the case for the genesis validator with id 8 on Concordium Mainnet.
ALTER TABLE bakers
```

limemloh (approved):

PR #552, opened 2025-02-24
Ranking bakers by lottery power

DOBEN:
## Purpose

Closes  https://github.com/Concordium/concordium-scan/issues/367

## Changes

```
query ($bakerId: Long!) {
  bakerByBakerId(bakerId: $bakerId) {
    state {
      __typename
      ... on ActiveBakerState {
        pool {
           rankingByTotalStake{
            rank
            total
          }

        }
      }
    }
  }
}

```

After discussion: we decided to use the `lottery_power` to rank bakers. The API name will be updated  to `rankingByLotteryPower` at a later PR since it introduces API breaking changes.


backend-rust/src/migrations/m00010-ranking-by-total-stake.sql
```diff
@@ -0,0 +1,11 @@
+-- Rename table and colum since the table is re-created with new data for each block and not only for each payday block now.
+ALTER TABLE bakers_payday_lottery_powers 
+RENAME TO bakers_lottery_powers;
+ALTER TABLE bakers_lottery_powers
+RENAME COLUMN payday_lottery_power TO lottery_power;
+
+-- Ranking all bakers by lottery powers staring with rank 1 for the baker with the highest lottery power.
```
lassemand, line 7:
```suggestion
-- Ranking all bakers by lottery powers
```
We can order both ASC/DESC, so talking about the ordering is a redundant/misleading.

lassemand: Tbh I would prefer just removing the comment, since the name of the column is self explanatory.

backend-rust/src/indexer.rs
```diff
@@ -918,6 +923,11 @@ impl PreparedBlock {
         )
         .await?;
         let baker_unmark_suspended = PreparedUnmarkPrimedForSuspension::prepare(data)?;
+
+        let election_info =
+            node_client.get_election_info(data.finalized_block_info.height).await?.response;
```
lassemand, line 928: While I understand the need for doing this on the block level, I do not think we should do this for every block. This is a heavy operation.

My suggestion would be that we do this not a all while catching up and while caught up then every say 30 block or so (~1m).

backend-rust/src/indexer.rs
```diff
@@ -4555,20 +4561,29 @@ impl PreparedPaydayCommissionRates {
...
     fn prepare(bakers: Vec<BirkBaker>) -> anyhow::Result<Self> {
         let capacity = bakers.len();
         let mut baker_ids: Vec<i64> = Vec::with_capacity(capacity);
         let mut bakers_lottery_powers: Vec<BigDecimal> = Vec::with_capacity(capacity);
+        let mut ranks: Vec<i64> = Vec::with_capacity(capacity);
+
+        // Sort bakers by lottery power. The baker with the highest lottery power comes
+        // first in the vector and gets rank 1.
+        let mut sorted_bakers: Vec<&BirkBaker> = bakers.iter().collect();
+        sorted_bakers.sort_by(|self_baker, other_baker| {
```
lassemand, line 4582: What is the issue with during this on the query side instead as we do everywhere else?

backend-rust/src/indexer.rs
limemloh, line 928: This query will return the same information for every block within the same reward period, so querying it at each payday block would be enough

backend-rust/src/indexer.rs
limemloh, line 4582: Since we take ownership of the `bakers` vec, we can just sort that one directly
```suggestion
    fn prepare(mut bakers: Vec<BirkBaker>) -> anyhow::Result<Self> {
        let capacity = bakers.len();
        let mut baker_ids: Vec<i64> = Vec::with_capacity(capacity);
        let mut bakers_lottery_powers: Vec<BigDecimal> = Vec::with_capacity(capacity);
        let mut ranks: Vec<i64> = Vec::with_capacity(capacity);

        // Sort bakers by lottery power. The baker with the highest lottery power comes
        // first in the vector and gets rank 1.
        bakers.sort_by(|self_baker, other_baker| {
```

backend-rust/src/indexer.rs
```diff
@@ -4555,20 +4561,29 @@ impl PreparedPaydayCommissionRates {
...
+            self_baker.baker_lottery_power.total_cmp(&other_baker.baker_lottery_power)
```
limemloh, line 4583: Did you ensure this is descending order? I expect it to be ascending but maybe I am missing something

backend-rust/src/indexer.rs
```diff
@@ -4578,11 +4593,13 @@ impl PreparedPaydayLotteryPowers {
                     )
                     .map_err(RPCError::ParseError)?,
             );
+            ranks.push((capacity - rank) as i64);
```
limemloh, line 4596: It seems to expect the bakers in ascending order here, so maybe the comment above the `sorted_by` is off

backend-rust/src/migrations/m00010-ranking-by-total-stake.sql
```diff
@@ -0,0 +1,11 @@
+-- Rename table and colum since the table is re-created with new data for each block and not only for each payday block now.
+ALTER TABLE bakers_payday_lottery_powers 
+RENAME TO bakers_lottery_powers;
+ALTER TABLE bakers_lottery_powers
+RENAME COLUMN payday_lottery_power TO lottery_power;
```
limemloh, line 5: Unless we are okay with downtime, this renaming needs to be split into two migrations and two releases, one introducing the new table another removing the old (which is marked as destructive).
But since this information only changes at payday I think the naming is fine.

backend-rust/src/migrations.rs
```diff
@@ -275,7 +278,15 @@ impl SchemaVersion {
             SchemaVersion::AddAccumulatedPoolState => {
                 m0008_canonical_address_and_transaction_search_index::run(&mut tx).await?
             }
-            SchemaVersion::AccountBaseAddress => unimplemented!(
+            SchemaVersion::AccountBaseAddress => {
+                tx.as_mut()
+                    .execute(sqlx::raw_sql(include_str!(
+                        "./migrations/m00010-ranking-by-total-stake.sql"
```
limemloh, line 284:
```suggestion
                        "./migrations/m0010-ranking-by-total-stake.sql"
```

backend-rust/src/migrations.rs
```diff
@@ -192,14 +192,16 @@ pub enum SchemaVersion {
     AddAccumulatedPoolState,
     #[display("0008:AccountBaseAddress")]
     AccountBaseAddress,
+    #[display("00010:RankingByTotalStake")]
```
limemloh, line 195:
```suggestion
    #[display("0010:RankingByTotalStake")]
```

backend-rust/src/indexer.rs
DOBEN: The comment was off (corrected).

backend-rust/src/indexer.rs
DOBEN: Good point, I initial thought the `lottery_power` is computed by the current stake at the queried `block_height` of the baker in the `get_election_info(block_height)`, but it seems to return the historical data from the last payday_block:

https://github.com/Concordium/concordium-node/blob/5dcc2b4ca777d65f119871c2735c01d410bc1f09/concordium-consensus/src/Concordium/Queries.hs#L1092

DOBEN: Changed it to only do the calculation on a payday block (once every day).

backend-rust/src/migrations/m00010-ranking-by-total-stake.sql
DOBEN: hmm, I don't think the order is self-explanatory. A ranking can go from highest to lowest in terms of lottery power or from lowest to highest.

backend-rust/src/indexer.rs
DOBEN: Honestly, there are different pros and cons for both sides (both can be reasonable ways to represent the data to the users).

- Moving the logic to the graph_QL side makes sense if we want to focus on up-to-date data and re-compute the values every time a query is sent. 
- Keeping the logic in the indexer focuses on presenting a ranking on historical 'average' data over a longer time period (one day).

- Disadvantage of moving to the graph_QL side: The baker queries are already very complex and we need to re-compute the ranking in the graphQL service on potentially many bakers (bad query performance) at every 'bakerRanking' request. The old CCDScan, also computes the ranking historically on the payday block and as such the ranking does not update that often. People are less inclined to move around stake in short intervals to get their bakers jump up the rankings. 

- Advantage: If we move the calculation to the graphQL side we can compute the ranking on the most up-to-date stake of the bake. Immediate effects are visible if the stake of a baker increases.

lassemand (approved):

PR #556, opened 2025-02-27
Baker validation

lassemand:
## Purpose

Same as https://github.com/Concordium/concordium-scan/pull/553 but for additional queries.

### DIFF
--- a/backend-rust/src/indexer.rs
+++ b/backend-rust/src/indexer.rs
@@ -1670,10 +1670,12 @@ struct EventMetadata {
     protocol_version: ProtocolVersion,
 }
 
-/// Metadata is present because prior to protocol version 7, removing a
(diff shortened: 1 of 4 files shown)


backend-rust/src/indexer.rs
```diff
@@ -4229,7 +4276,13 @@ impl RestakeEarnings {
     async fn save(
         &self,
         tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+        protocol_version: ProtocolVersion
```
lassemand, line 4279: I cannot send the range along, as I am unaware what range the individual validations are using already.

Also a time may come where additional metadata may be needed.

I could send the metadata along, I am fine with this.

limemloh (approved):

DOBEN (approved):

PR #557, opened 2025-02-27
Don't update delegation target for removed bakers

limemloh:
## Purpose

Currently, the new indexer fails on testnet with:
```
    insert or update on table "accounts" violates foreign key constraint "fk_delegated_target_baker_id"
```
This is due to delegators updating their target pool after a baker was removed (possible prior to P7) since we remove the baker immediately even for these blocks.

This PR skips the update when the pool is already removed.


DOBEN (approved):

backend-rust/src/indexer.rs
```diff
@@ -1989,18 +1989,27 @@ impl PreparedAccountDelegationEvent {
...
                 }
                 // Set the new target on the delegator.
+                // Prior to Protocol version 7, removing a baker was not immediate, but after
+                // some cooldown period, allowing delegators to still target the pool after
+                // removal. Since we remove the baker immediate even for older blocks there
+                // might not be a baker to target, so we check for existence as part of the
+                // query.
                 sqlx::query!(
-                    "UPDATE accounts SET delegated_target_baker_id = $1 WHERE index = $2",
+                    "UPDATE accounts
```
DOBEN, line 2002: This means the `old_delegation_target` stays for the delgators? Would that not also cause long-term problems?

limemloh: The old ones are cleaned up in migration 5 and the constraint ensures this stays true, which is also why it fails at the moment.
We are just missing handling of delegators targeting bakers after they are set to be removed.

PR #564, opened 2025-03-03
Add more context

limemloh:
## Purpose

When the indexer processing fails very little context is provided, this PR adds context to a lot of the processing and derives debug several types.

Besides better error messages, no behavior changes are intended from this PR.


backend-rust/src/indexer.rs
```diff
@@ -3639,7 +3733,10 @@ async fn process_cis2_token_event(
             )
             .execute(tx.as_mut())
             .await?
-            .ensure_affected_one_row()?;
+            .ensure_affected_one_row()
+            .with_context(|| {
+                format!("Failed inserting the token transfer event: {:?}", cis2_transfer_event)
+            })?;
```
DOBEN, line 3739: The `context` should be on the `await` error since this is the action that inserts.

```
  .execute(tx.as_mut())
                .await
                .context("Failed inserting the token transfer event: {:?}", cis2_transfer_event")?
                .ensure_affected_rows_in_range(0..=1)?;
```

backend-rust/src/indexer.rs
```diff
@@ -3880,7 +3980,13 @@ impl PreparedUpdateAccountBalance {
...
         .execute(tx.as_mut())
         .await?
-        .ensure_affected_one_row()?;
+        .ensure_affected_one_row()
+        .with_context(|| {
+            format!(
+                "Failed processing update to account balance, change: {}, canonical address: {:?}",
+                self.change, self.canonical_address
+            )
+        })?;
```
DOBEN, line 3989: See above.

backend-rust/src/indexer.rs
```diff
@@ -2849,13 +2912,16 @@ impl PreparedContractUpdates {
         tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
         transaction_index: i64,
     ) -> anyhow::Result<()> {
-        for elm in &self.trace_elements {
-            elm.save(tx, transaction_index).await?;
+        for (i, elm) in self.trace_elements.iter().enumerate() {
+            elm.save(tx, transaction_index).await.with_context(|| {
+                format!("Failed processing contract update trace element with index {}", i)
```
DOBEN, line 2917: I don't think we print the trace element in case of an error anywhere, so would be good to add.

```suggestion
                format!("Failed processing contract update trace element {:?} with index {}", elm, i)
```

DOBEN (approved):

backend-rust/src/indexer.rs
limemloh: I'll add it to both then, the current is on the validation of the affected rows so also related to the action that inserts.
Ideally, we would only have this context once on a database operation struct so it covers both cases.

Issue #562, opened 2025-02-28
Capital bound cap formula underflows under some network conditions


DOBEN:
**Task description**

The `capital_bound_cap` formula underflows under some network conditions with the error:
`attempt to subtract with overflow` (e.g. on stagnet or when beginning indexing testnet)
https://github.com/Concordium/concordium-scan/blob/ccdscan-backend/0.1.31/backend-rust/src/graphql_api/baker.rs#L493

e.g.
capital_bound: 5000, total_stake: 5841391183506362, delegated_stake_of_pool: 0, staked:1000000000000000

Early protocol versions do not have the concept of `capital_bound_cap`/`leverage_bound_cap` and an error might be acceptable/expected in that case. The task is to investigate when the error is to be expected and how to adjust the formula if needed.

DOBEN:
Closing the issue because negative values are cast to 0 in the node and CCDScan backend was updated to mirrors this behavior such that the capital bound does not underflow anymore.

https://github.com/Concordium/concordium-node/blob/3cb759e4607f20a9df94ace017f0e3c775d4cdb3/concordium-consensus/src/Concordium/Kontrol/Bakers.hs#L65

PR #567, opened 2025-03-03
Use saturated sub at capital bound formula

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-scan/issues/562

This helps to use the baker query again. Investigation into the formula still needs to be done (the issue stays open and is not addressed).

### DIFF
--- a/backend-rust/src/graphql_api/baker.rs
+++ b/backend-rust/src/graphql_api/baker.rs
@@ -499,7 +499,7 @@ impl Baker {
             // u128.
             let capital_bound_cap_for_pool_numerator = capital_bound
                 * ((total_stake - delegated_stake_of_pool) as u128)
-                - (100_000 * (self.staked as u128));
+                    .saturating_sub(100_000 * (self.staked as u128));
 
             // Denominator is not zero since we checked that `capital_bound != 100_000`.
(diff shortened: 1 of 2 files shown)


backend-rust/src/graphql_api/baker.rs
```diff
@@ -498,7 +498,7 @@ impl Baker {
             // u128.
             let capital_bound_cap_for_pool_numerator = capital_bound
                 * ((total_stake - delegated_stake_of_pool) as u128)
-                - (100_000 * (self.staked as u128));
+                    .saturating_sub(100_000 * (self.staked as u128));
```
lassemand, line 501: From what I could see, staked should not be multiplied by 100_000 based on the state the nodes which I queried was in.

This could be related to the genesis loading.

Please add a todo comment

lassemand (approved):

backend-rust/src/graphql_api/baker.rs
DOBEN: For lower protocol version the formula is off since `leverage_bound/capital_bound` does not exist.

lassemand: It exists using the current database in stagenet querying baker 0-4.

PR #569, opened 2025-03-03
Empty lottery power

lassemand:
## Purpose

If the baker has no lottery power total lottery power has been set.


backend-rust/src/graphql_api/baker.rs
```diff
@@ -522,7 +522,7 @@ impl Baker {
                 rank,
                 total,
             }),
-            (None, None) => None,
```
lassemand, line 525: Seems pretty odd in general we can query before total lottery power has been calculated. Should probably add this to the health endpoint

PR #571, opened 2025-03-05
Node status

lassemand:
## Purpose

Attach additional fields required as part of NodeStatus
Add node_status endpoint

## Comments

No corresponding issue has been created.
I altered the recurring node_status field, to instead only contain the data required.
There is an issue when a node is click which is no longer present (Should also be an issue on the existing UI). A task has been created: https://linear.app/concordium/issue/CCD-203/node-collector-changes

### DIFF
--- a/backend-rust/src/graphql_api/node_status.rs
+++ b/backend-rust/src/graphql_api/node_status.rs
@@ -155,7 +155,7 @@ impl Service {
                                             }
                                         });
                                         PeerReference {
-                                            node_id: Arc::new(node_id.to_string()),
+                                            node_id: node_id.to_string(),
                                             node_status: peer
                                         }
                                     }).collect();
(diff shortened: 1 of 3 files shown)


backend-rust/src/graphql_api/node_status.rs
```diff
@@ -147,45 +186,116 @@ impl Service {
...
 
-#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Clone, Default)]
 #[serde(rename_all = "camelCase")]
 #[derive(SimpleObject)]
 #[graphql(complex)]
-pub struct NodeStatus {
-    pub node_id: String,
-    pub node_name: Option<String>,
+pub struct ExternalNodeStatus {
```
lassemand, line 193: I am pretty certain most of these fields requested are not being used by the frontend.

For now lets just support them all.

limemloh (approved):

PR #574, opened 2025-03-08
Baker metrics

lassemand:
## Purpose

Implement baker metrics
Closes https://github.com/Concordium/concordium-scan/issues/226


backend-rust/src/indexer.rs
```diff
@@ -617,8 +620,9 @@ impl ProcessEvent for BlockProcessor {
         let mut new_context = self.current_context.clone();
         PreparedBlock::batch_save(batch, &mut new_context, &mut tx).await?;
         for block in batch {
+            let mut statistics = Statistics::new(block.height);
```
limemloh, line 623: We don't rely on anything that is not already available in the preprocessing step, so it would ideally be computed during this step instead of in the processing step.

limemloh (changes requested):

backend-rust/src/indexer.rs
lassemand: Well we cant. I need access to the transaction

backend-rust/src/indexer/statistics.rs
```diff
@@ -0,0 +1,98 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) enum Field {
+    Added,
+    Removed,
+}
+
+pub(crate) struct Statistics {
+    /// The counters as updated via increments.
+    current:      HashMap<Field, i64>,
```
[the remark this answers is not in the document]
lassemand: Performant is not really an argument. 

It could be using integers as well, however next step is to increase the current increments with resumed and suspended. This indeed seems more flexible.

backend-rust/src/graphql_api/baker_metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+        .await?;
+
+        let (before_added, before_removed) = before_period_row
+            .map(|r| (r.total_bakers_added, r.total_bakers_removed))
+            .unwrap_or((0, 0));
+        let (after_added, after_removed) = last_in_period_row
+            .map(|r| (r.total_bakers_added, r.total_bakers_removed))
+            .unwrap_or((0, 0));
+
+        let last_baker_count = before_added - before_removed;
```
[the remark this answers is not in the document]
lassemand:
```
                        coalesce((SELECT total_baker_count
                                  FROM metrics_bakers m2
                                  WHERE m2.time < @FromTime
                                  ORDER BY time DESC

```

Taken directly from the reference code

backend-rust/src/graphql_api/baker_metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+        for r in rows.iter() {
+            x_time.push(r.bucket_time);
+            let added_during_period: u64 = r.bucket_bakers_added.try_into()?;
+            y_bakers_added.push(added_during_period);
+            let removed_during_period: u64 = r.bucket_bakers_removed.try_into()?;
+            y_bakers_removed.push(removed_during_period);
+            y_last_baker_count.push(
+                added_during_period - removed_during_period
+                    + TryInto::<u64>::try_into(last_baker_count)?,
+            );
```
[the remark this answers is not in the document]
lassemand: it was right before I updated the last_baker_count. I will update accordingly.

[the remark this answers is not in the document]
lassemand:
```
It is including before: 

{
  "data": {
    "bakerMetrics": {
      "lastBakerCount": 37,
      "bakersAdded": 14,
      "bakersRemoved": 6,
      "buckets": {
        "bucketWidth": "P1D",
        "x_Time": [
          "2025-02-09T00:00:00.000Z",
          "2025-02-10T00:00:00.000Z",
          "2025-02-11T00:00:00.000Z",
          "2025-02-12T00:00:00.000Z",
          "2025-02-13T00:00:00.000Z",
          "2025-02-14T00:00:00.000Z",
          "2025-02-15T00:00:00.000Z",
          "2025-02-16T00:00:00.000Z",
          "2025-02-17T00:00:00.000Z",
          "2025-02-18T00:00:00.000Z",
          "2025-02-19T00:00:00.000Z",
          "2025-02-20T00:00:00.000Z",
          "2025-02-21T00:00:00.000Z",
          "2025-02-22T00:00:00.000Z",
          "2025-02-23T00:00:00.000Z",
          "2025-02-24T00:00:00.000Z",
          "2025-02-25T00:00:00.000Z",
          "2025-02-26T00:00:00.000Z",
          "2025-02-27T00:00:00.000Z",
          "2025-02-28T00:00:00.000Z",
          "2025-03-01T00:00:00.000Z",
          "2025-03-02T00:00:00.000Z",
          "2025-03-03T00:00:00.000Z",
          "2025-03-04T00:00:00.000Z",
          "2025-03-05T00:00:00.000Z",
          "2025-03-06T00:00:00.000Z",
          "2025-03-07T00:00:00.000Z",
          "2025-03-08T00:00:00.000Z",
          "2025-03-09T00:00:00.000Z",
          "2025-03-10T00:00:00.000Z",
          "2025-03-11T00:00:00.000Z"
        ],
        "y_LastBakerCount": [
          29,
          29,
          29,
          30,
          31,
          31,
          31,
          31,
          31,
          31,
          31,
          32,
          32,
          32,
          32,
          35,
          35,
          35,
          36,
          36,
          36,
          36,
          36,
          36,
          37,
          37,
          37,
          37,
          37,
          37,
          37
        ],
        "y_BakersAdded": [
          0,
          0,
          0,
          1,
          1,
          1,
          0,
          0,
          0,
          0,
          0,
          1,
          0,
          0,
          0,
          4,
          2,
          0,
          3,
          0,
          0,
          0,
          0,
          0,
          1,
          0,
          0,
          0,
          0,
          0,
          0
        ],
        "y_BakersRemoved": [
          0,
          0,
          0,
          0,
          0,
          1,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          1,
          2,
          0,
          2,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0,
          0
        ],
        "__typename": "BakerMetricsBuckets"
      },
      "__typename": "BakerMetrics"
    }
  }
}
```

backend-rust/src/indexer.rs
lassemand: The way for me to achieve this would be to alter PreparedBlock to include Statistics.

Do you find that being more clean?

backend-rust/src/migrations/m0013-baker-metrics.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE metrics_bakers
+(
+    block_height            BIGINT            PRIMARY KEY REFERENCES blocks(height),
+    total_bakers_added      BIGINT            NOT NULL,
+    total_bakers_removed    BIGINT            NOT NULL
+);
+
+-- Find BakerAdded and BakerRemoved events and fill up metrics_bakers correspondingly
+INSERT INTO metrics_bakers (block_height, total_bakers_removed, total_bakers_added)
```
lassemand, line 9: This has been running for 13 mins against testnet data. Should probably be improved performance wise.

lassemand: Took ~30 mins.

backend-rust/src/indexer/statistics.rs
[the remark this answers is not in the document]
limemloh: Performance is definitely an argument, increment a counter is one instruction on the CPU, updating something in a hash map is way way more than one instruction.
But not only that, the code becomes simpler by not using a hash map, which was the point I highlighted the most.

backend-rust/src/migrations/m0013-baker-metrics.sql
limemloh: There won't be any transactions with both added and removed bakers in the same list of events, so it would suffice to just count the transaction containing some BakerAdded event and containing BakerRemove events.
So the inner query would probably perform better using:
```sql
SELECT
    block_height,
    COUNT(*) FILTER (WHERE events @> '[{"BakerRemoved": {}}]'::JSONB) AS baker_removed_count,
    COUNT(*) FILTER (WHERE events @> '[{"BakerAdded": {}}]'::JSONB) AS baker_added_count
FROM transactions
WHERE
    type_account IN ('RemoveBaker', 'AddBaker')
    OR (
        type_account IN ('ConfigureBaker', 'ConfigureDelegation')
        AND (events @> '[{"BakerRemoved": {}}]'::JSONB
            OR events @> '[{"BakerAdded": {}}]'::JSONB
        )
    )
GROUP BY block_height
ORDER BY block_height
```

lassemand: I altered the code which seems to increase the performance a bit.

backend-rust/src/indexer/statistics.rs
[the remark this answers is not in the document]
lassemand: Altered adhere more to what you wrote. I used a boolean to reflect state changes to the baker.

backend-rust/src/graphql_api/baker_metrics.sql
```diff
@@ -0,0 +1,30 @@
...
+    COALESCE(after_bucket.total_bakers_removed, 0) - COALESCE(before_bucket.total_bakers_removed, 0) as "bucket_bakers_removed!"
+FROM
+    date_bin_series(
+        $3::interval,
+        $2,
+        $1
+    ) AS bucket_time
+LEFT JOIN LATERAL (
+    SELECT
+        total_bakers_added,
```
lassemand, line 13: This could be MIN and MAX and also where on slot_time start of period. Since this is on query time we should do that.

backend-rust/src/graphql_api/baker_metrics.rs
```diff
@@ -0,0 +1,153 @@
...
+            WHERE blocks.slot_time < $1
+            ORDER BY metrics_bakers.block_height DESC
+            LIMIT 1
+            "#,
+            before_time,
+        )
+        .fetch_optional(pool)
+        .await?;
+
+        let last_in_period_row = sqlx::query!(
```
lassemand, line 48: We could rather easily avoid this statement, since this should also be available at the last entry of the buckets.

This does however imo also compromise a bit on readability.

backend-rust/src/migrations/m0013-baker-metrics.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE metrics_bakers
```
limemloh, line 1: Some documentation would be nice.
Also the fact that this table only contain rows for blocks where these have changed and not for every block.

backend-rust/src/graphql_api/baker_metrics.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        let before_period_row = sqlx::query!(
+            r#"
+            SELECT
+                total_bakers_added,
+                total_bakers_removed
+            FROM metrics_bakers
+            LEFT JOIN blocks ON metrics_bakers.block_height = blocks.height
+            WHERE blocks.slot_time < $1
+            ORDER BY metrics_bakers.block_height DESC
```
limemloh, line 40: I am not sure that postgres have enough information to be able to do this efficiently. Maybe I am wrong but it would need to know that `slot_time` and `block_height` are both strictly increasing, which is does not know. Maybe it would be better to sort by the `slot_time`:
```suggestion
            SELECT
                total_bakers_added,
                total_bakers_removed
            FROM metrics_bakers
            LEFT JOIN blocks ON metrics_bakers.block_height = blocks.height
            WHERE blocks.slot_time < $1
            ORDER BY blocks.slot_time DESC
```

backend-rust/src/indexer/statistics.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+        self.baker_is_changed = true;
+    }
+
+    /// If any counter has been incremented, updates the latest row in
+    /// metrics_bakers by adding the increments.
+    ///
+    /// The SQL query adds the current counter values to the corresponding
+    /// columns: total_bakers_added, total_bakers_removed,
+    /// total_bakers_resumed, total_bakers_suspended.
```
DOBEN, line 41:
```suggestion
    /// The SQL query adds the current counter values to the corresponding
    /// columns: total_bakers_added, total_bakers_removed
```

backend-rust/src/indexer/statistics.rs
```diff
@@ -0,0 +1,97 @@
...
+              LIMIT 1
+            )
+            "#,
+            self.block_height,
+            self.baker_added_count,
+            self.baker_removed_count
+        )
+        .execute(tx.as_mut())
+        .await?;
+        if result.rows_affected() == 0 {
```
DOBEN, line 74: Please add a comment, when this is the case.
I assume this is to handle the `genesis block` when a indexer is started from scratch.

backend-rust/src/migrations/m0014-baker-metrics.sql
```diff
@@ -0,0 +1,11 @@
+-- Table used for showing graphs on amount of bakers
+-- Each row represents a change in any value for this particular block
+
+CREATE TABLE metrics_bakers
+(
+    block_height            BIGINT            PRIMARY KEY REFERENCES blocks(height),
+    -- total amount of added accounts at this particular block
```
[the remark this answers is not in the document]
lassemand: We can leave out the part about genesis block. It is also included

backend-rust/src/indexer/statistics.rs
lassemand: I will name the boolean expression instead

PR #580, opened 2025-03-13
`Query::bakers` sorting by block commission

### LINKED ISSUE #578: Implement Query::bakers sort by block commissions
Current rust backend does not implement the sorting options for `BlockCommissions`

limemloh:
## Purpose

Closes #578


backend-rust/src/graphql_api/baker.rs
```diff
@@ -115,8 +115,32 @@ impl QueryBaker {
             }
             BakerSort::BakerApy30DaysDesc => todo!(),
             BakerSort::DelegatorApy30DaysDesc => todo!(),
-            BakerSort::BlockCommissionsAsc => todo!(),
-            BakerSort::BlockCommissionsDesc => todo!(),
+            BakerSort::BlockCommissionsAsc => {
+                Baker::block_commission_asc_connection(
```
lassemand, line 119: Fine that we implement it for compatibility, but I do not see any use for this particular feature.

backend-rust/src/graphql_api/baker.rs
```diff
@@ -1045,6 +1076,519 @@ impl Baker {
...
+                    metadata_url,
+                    transaction_commission,
+                    baking_commission,
+                    finalization_commission,
+                    payday_transaction_commission as "payday_transaction_commission?",
+                    payday_baking_commission as "payday_baking_commission?",
+                    payday_finalization_commission as "payday_finalization_commission?",
+                    payday_lottery_power as "payday_lottery_power?",
+                    payday_ranking_by_lottery_powers as "payday_ranking_by_lottery_powers?",
+                    (SELECT MAX(payday_ranking_by_lottery_powers) FROM bakers_payday_lottery_powers) as "payday_total_ranking_by_lottery_powers?",
```
lassemand, line 1119: Does bakers_payday_lottery_powers only return a single row or why do we return MAX here?

backend-rust/src/graphql_api/baker.rs
limemloh: I kind of agree, but when it was not that hard to implement, and it is isolated in this function, so it would be easy to remove again.

backend-rust/src/graphql_api/baker.rs
lassemand, line 1119: Why make a subquery here?

backend-rust/src/graphql_api/baker.rs
```diff
@@ -1045,6 +1076,519 @@ impl Baker {
...
+                    pool_total_staked,
+                    pool_delegator_count
```
lassemand, line 1121: In general it is not good practice that we keep on spamming this select into so many different calls. This make altercations on bakers cumbersome and prone to errors. Can we define one place what a baker look like when we have this many calls?

backend-rust/src/graphql_api/baker.rs
```diff
@@ -1045,6 +1076,519 @@ impl Baker {
...
+        // below the limit, we query the second collection. The order of which
+        // collection to query first and second will depend on the whether the
+        // `last` parameter was provided in the top level query.
+        if query.is_last {
+            if include_removed_filter {
+                if let Some(removed_baker_query) = query.subquery_second() {
+                    query_removed_baker(removed_baker_query, &mut connection, pool).await?;
+                }
+            }
+            let remains_to_limit = query.limit - i64::try_from(connection.edges.len())?;
```
lassemand, line 1219: I would enjoy if we could isolate this limit behaviour and make some unit tests on it. 

Params should be something like length of the two collections, cursor direction (DESC or ASC) and then a pointer to where we start and a length to be read.

backend-rust/src/graphql_api/baker.rs
limemloh: I don't think it is prone to errors due to the compile time checked queries.

But I agree that it is cumbersome. One solution would be to not use compile time checked queries, so we split up and compose the queries, but I think this becomes prone to errors, and we would need more testing to catch mistakes, which is probably more cumbersome.

Another solution would be to only fetch `id`s in this query and then have another query, which fetches from a list of `id`s, meaning we only list every selected field in one query. Here it would mean a hit in performance instead, but I am not sure how much though and whether it would be worth it.

lassemand: We can consider it another task?

lassemand: I am writing a task on it. It could also be that sqlx defines a way in how to map from a Struct to a select statement.

lassemand: https://linear.app/concordium/issue/CCD-207/unify-selects-on-bakers

lassemand (approved):

DOBEN (approved):

backend-rust/src/graphql_api/baker.rs
limemloh: I am not sure what you want to isolate exactly.
I did briefly try to isolate some of this logic, but unfortunately abstracting async functions in rust can quickly blow up in complexity and there might be some way to isolate things into traits, but that I will leave for future iteration.

backend-rust/src/graphql_api/baker.rs
```diff
@@ -1045,6 +1076,519 @@ impl Baker {
...
+        }
+        if include_removed_filter {
+            let min_removed_baker_id =
+                sqlx::query_scalar!("SELECT MIN(id) FROM bakers_removed",).fetch_one(pool).await?;
+            connection.has_next_page = if let Some(min_removed_baker_id) = min_removed_baker_id {
+                last_item.node.get_id() != min_removed_baker_id
+            } else {
+                false
+            }
+        }
```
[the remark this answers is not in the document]
limemloh: When `include_removed_filter` is enabled, we just append the removed bakers sorted by id desc.
So the last item in the total collection, will be the removed baker with the smallest baker ID, so `has_next_page` is simply checking whether the last item is this baker id.
I think this is correct :sweat_smile:

backend-rust/src/graphql_api/baker.rs
```diff
@@ -1045,6 +1076,519 @@ impl Baker {
...
+                query.to.field,                                    // $2
+                query.from.baker_id,                               // $3
+                query.to.baker_id,                                 // $4
+                query.is_last,                                     // $5
+                query.limit,                                       // $6
+                open_status_filter as Option<BakerPoolOpenStatus>  // $7
+            )
+            .fetch(pool);
+            while let Some(row) = row_stream.try_next().await? {
+                let cursor = Cursor::First(BakerFieldDescCursor::block_commission_rate(&row));
```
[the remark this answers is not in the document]
limemloh: Right, we should not break the api, but we doc comments become public in the API schema and there we can 'break' by describing things.

PR #583, opened 2025-03-14
Baker metrics start time

lassemand:
## Purpose

Baker bucket start time was not defined accordingly to total start time.

This PR alters the calculation to be using the metrics as provided by the buckets instead of using another time interval.


limemloh (approved):
Don't forget to update the changelog

Issue #584, opened 2025-03-17
Implement remaining queries for `passive_delegation`


DOBEN:
**Task description**

**Sub-tasks**
- [x] PassiveDelegation::delegators
- [x] PassiveDelegation::delegatorCount
- [x] PassiveDelegation::commissionRates
- [x] PassiveDelegation::delegatedStake
- [x] PassiveDelegation::delegatedStakePercentage
- [x] PassiveDelegation::apy30days/apy7days

https://github.com/Concordium/concordium-scan/blob/main/backend-rust/src/graphql_api/passive_delegation.rs

PR #585, opened 2025-03-17
PassiveDelegation queries

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-scan/issues/584

## Changes

- Add database schema 16
- Add `PassiveDelegation::delegators` query
- Add `PassiveDelegation::delegatorCount` query
- Add `PassiveDelegation::commissionRates` query
- Add `PassiveDelegation::delegatedStake` query
- Add `PassiveDelegation::delegatedStakePercentage` query

```
query ($afterDelegators: String, $beforeDelegators: String, $firstDelegators: Int, $lastDelegators: Int) {
  passiveDelegation {
    delegators(
      after: $afterDelegators
      before: $beforeDelegators
      first: $firstDelegators
      last: $lastDelegators
    ) {
      nodes {
        accountAddress {
          asString
          __typename
        }
        stakedAmount
        restakeEarnings
        __typename
      }
      pageInfo {
        hasNextPage
        hasPreviousPage
        startCursor
        endCursor
        __typename
      }
      __typename
    }
    commissionRates {
      transactionCommission
      finalizationCommission
      bakingCommission
      __typename
    }
    delegatorCount
    delegatedStake
    delegatedStakePercentage
    __typename
  }
}
```


backend-rust/src/migrations/m0016-passive-delegation.sql
```diff
@@ -0,0 +1,23 @@
+CREATE INDEX passive_delegators_idx
+ON accounts (delegated_target_baker_id, delegated_stake)
+WHERE delegated_target_baker_id IS NULL AND delegated_stake != 0;
+
+CREATE TABLE passive_delegation_payday_commission_rates(
```
lassemand, line 5: I would need this table as well to calculate the reward metrics

lassemand: I actually need the history. Could you just store the history?

lassemand: After discussion it does not seem as if these are included into the rewards table. For now we are doing the same.

backend-rust/src/indexer.rs
```diff
@@ -5028,16 +5046,82 @@ impl PreparedPayDayBlock {
...
+        })
+    }
+
+    async fn save(
+        &self,
+        tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
+    ) -> anyhow::Result<()> {
+        // The fields `transaction_commission`, `baking_commission`, and
+        // `finalization_commission` are either all `Some` or all `None`.
+        if let (
```
lassemand, line 5084: Our pattern does not really fit well for terminating quickly. We should probably look into adding some common trait to the Prepared structs, such that we can provide a conditional as to when we want to save.

lassemand (approved):

backend-rust/src/indexer.rs
DOBEN: After discussions:
Moving checks up the functions will create bigger, more unclear top-level functions (that handle many cases). e.g.:
```
if let (Some(transaction_commission),
            Some(baking_commission),
            Some(finalization_commission),
        ) = (self.transaction_commission, self.baking_commission, self.finalization_commission)
        {

       self.passive_delegation_payday_commission_rates.save(tx).await?;
}
```
We keep the current behavior. Large `Prepared structs` can benefit from introducing a flag to specify whether to save the data to the database (as Lasse suggested).

PR #587, opened 2025-03-17
Reward metrics

lassemand:
Closes: https://linear.app/concordium/issue/CCD-75/implement-queryrewardmetrics-and-queryrewardmetricsforaccount

## Purpose

Implement rewards metrics


backend-rust/src/graphql_api/baker_metrics.rs
```diff
@@ -70,14 +70,14 @@ impl QueryBakerMetrics {
             y_last_baker_count.push(current_period_baker_count);
         }
 
-        let last_baker_count = y_last_baker_count.last().copied().ok_or_else(|| {
+        let last_baker_count = y_last_baker_count.iter().max().ok_or_else(|| {
```
lassemand, line 73: If for example indexer is down we still have a total value

backend-rust/src/indexer.rs
```diff
@@ -4380,6 +4388,9 @@ impl PreparedPaydaySpecialTransacionOutcomes {
         &self,
         tx: &mut sqlx::Transaction<'static, sqlx::Postgres>,
     ) -> anyhow::Result<()> {
+        if !self.has_reward_events {
```
lassemand, line 4391: Most of these are empty. Not really a reason why we would want to perform database operations on empty list

backend-rust/src/graphql_api/reward_metrics.rs
```diff
@@ -0,0 +1,102 @@
...
+        account_id.map(|x| x.try_into().map_err(ApiError::InvalidIdInt)).transpose()?;
+
+    let rows = sqlx::query!(
+        r#"
+        SELECT
+            bucket_time.bucket_start AS "bucket_time!",
+            (SELECT COALESCE(SUM(amount), 0)
+                            FROM metrics_rewards
+                            WHERE block_slot_time BETWEEN bucket_time.bucket_start AND bucket_time.bucket_end
+                            AND ($4::BIGINT IS NULL OR account_index = $4::BIGINT))::BIGINT AS "accumulated_amount!"
```
lassemand, line 55: Has been tested against stagenet which by far has the most dense dataset on rewards. Performed fine.

backend-rust/src/indexer.rs
```diff
@@ -4746,7 +4770,12 @@ impl AccountReceivedReward {
         block_height: AbsoluteBlockHeight,
         transaction_type: AccountStatementEntryType,
         protocol_version: ProtocolVersion,
+        statistics: &mut Statistics,
     ) -> anyhow::Result<Self> {
+        statistics.increment(
```
lassemand, line 4775: I could also validate whether transaction_type is of reward type

backend-rust/src/migrations/m0017-reward-metrics.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE metrics_rewards (
+    block_height        BIGINT NOT NULL,
+    block_slot_time     TIMESTAMPTZ NOT NULL,
```
limemloh, line 3: The block_height should be enough here, you can join to get the slot_time

backend-rust/src/migrations/m0017-reward-metrics.sql
```diff
@@ -0,0 +1,32 @@
...
+    account_index       BIGINT NOT NULL,
```
limemloh, line 4:
```suggestion
    block_height        BIGINT NOT NULL REFERENCES blocks,
    account_index       BIGINT NOT NULL REFERENCES accounts,
```

backend-rust/src/migrations/m0017-reward-metrics.sql
```diff
@@ -0,0 +1,32 @@
...
+    SUM(amount) AS rewards
+  FROM account_statements
+  WHERE entry_type IN (
+    'FinalizationReward',
+    'FoundationReward',
+    'BakerReward',
+    'TransactionFeeReward'
+  )
+  GROUP BY block_height, account_index
+  HAVING COUNT(*) > 0
```
limemloh, line 23: Do we need this line? I don't see when this could be 0

backend-rust/src/indexer/statistics.rs
```diff
@@ -1,52 +1,66 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) enum RewardField {
+    Account(CanonicalAccountAddress),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) enum StatisticsField {
+    Baker(BakerField),
+    Reward(RewardField),
+}
```
limemloh, line 23: Instead of building a larger and larger enum to represent every type of increment and have a single increment function, it seems simpler to just introduce separate `increment` functions for each statistic. You avoid constructing these enums everywhere and branching in the increment function.

backend-rust/src/migrations/m0017-reward-metrics.sql
```diff
@@ -0,0 +1,32 @@
...
+)
+SELECT
+  block_height,
+  block_slot_time,
+  account_index,
+  rewards
+FROM per_block;
+
+CREATE INDEX metrics_rewards_accounts_slot_time_idx ON metrics_rewards(account_index, block_slot_time);
```
limemloh, line 32: This should probably be the primary key of the table.

backend-rust/src/graphql_api/baker_metrics.rs
lassemand: Becauxe this is correct, the previous was not. I can also do it as part of another PR, but it has to be done like this.

backend-rust/src/migrations/m0017-reward-metrics.sql
lassemand: I can collect some metrics if you want, pretty sure it is significantly slower.

backend-rust/src/migrations/m0017-reward-metrics.sql
lassemand: It is rather often 0. It is empty when there are no rewards.

backend-rust/src/graphql_api/baker_metrics.rs
limemloh: But if the count is 50 in the past and goes down to 40 in the latest block, this will return 50 instead of 40, right?

backend-rust/src/indexer/statistics.rs
lassemand: I did this to avoid having annoying namings on the methods. 

Sure this is another way of doing

backend-rust/src/graphql_api/baker_metrics.rs
lassemand: Yes you are right. I was a thinking it was growing, but that is indeed a problem.

Ehh. That will probably be hard to fix for no occurring last bakers then. I should probably revert.

backend-rust/src/indexer/statistics.rs
limemloh: But this is making something abstract for no reason, we don't need to be able to call a single `increment` function where the field is abstract, because we know which to increment in every place.
We only introduce complexity if strictly needed, and so far we don't need anything to be generic.

backend-rust/src/migrations/m0017-reward-metrics.sql
lassemand: As per discussion this is indeed possible. 

Simply adding indexes and doing a join is too slow. This is related to it having to do a full join before it can derive what the date is.

So we agreed on 3 possible strategies:

1. Optimise query with join such that it only joins on the range
2. Use the fact that there is a relationship between height and time, meaning we could simply choose min and max in the range and do a between on that.
3. Use it as is 

We both agree that number one is possible, but it would require that we can tell the optimiser to only JOIN the range of within the bucket
Number two is a bit more unreadable and less elegant.
Number three uses redundant data. 

So basically since the data is available for three, and since that is the current implementation we are ok with that. In particular because this is a statistics table.

limemloh (approved):

PR #590, opened 2025-03-20
Total count

lassemand:
## Purpose

Enable total count, as is enabled on dotnet version and request by the client. https://github.com/Concordium/concordium-scan/blob/main/backend/Application/Api/GraphQL/Transactions/TransactionResult.cs

### DIFF
--- a/backend-rust/src/graphql_api/transaction.rs
+++ b/backend-rust/src/graphql_api/transaction.rs
@@ -11,7 +11,9 @@ use crate::{
         UpdateTransactionType,
     },
 };
-use async_graphql::{connection, types, Context, Object, SimpleObject, Union};
+use async_graphql::{
+    connection, connection::Connection, types, Context, Object, SimpleObject, Union,
+};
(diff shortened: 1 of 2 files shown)


backend-rust/src/graphql_api/transaction.rs
```diff
@@ -302,14 +309,26 @@ impl Success<'_> {
...
+                end == self.events.len(),
+                AdditionalFields {
+                    total_count,
+                },
+            );
+        connection.edges = edges;
         Ok(connection)
     }
+
+    async fn total_count(&self) -> u64 { self.events.len() as u64 }
```
limemloh, line 331: is this needed? The additional fields will be enough right?

limemloh (approved):

backend-rust/src/graphql_api/transaction.rs
lassemand: Yes indeed

PR #591, opened 2025-03-20
Fix total count

lassemand:
## Purpose

Change total count to be the total length of the underlying dataset

**Note** changelog was updated in another PR, and no release has been made since


DOBEN (approved):

DOBEN:
We could just update the front-end to use the length of the nodes instead of providing it via the backend so.

PR #592, opened 2025-03-21
Chain update payload

lassemand:
## Purpose

Implement chain parameter updated event


frontend/src/queries/useTransactionQuery.ts
```diff
@@ -491,6 +491,10 @@ reason {
 		bakerId
 		__typename
   }
+	... on ValidatorScoreParametersUpdate {
```
lassemand, line 494: Seems to be missing on the frontend

lassemand:
Hmm.. I also have to alter TransferredWithSchedule since it throws away the timestamp of the amounts scheduled.

I may as well finish that as part of this PR as well, such that we get both migrations working at once.

backend-rust/src/migrations/m0019_chain_update_events.rs
```diff
@@ -0,0 +1,64 @@
...
+        next_schema_version
+    ))?;
+
+    let mut client = v2::Client::new(endpoint.clone()).await?;
+
+    let rows = sqlx::query(
+        "
+            SELECT
+                block_height,
+                COUNT(*) AS update_count
```
DOBEN, line 27: it seems the `update_count` is not used.

backend-rust/src/migrations/m0019_chain_update_events.rs
```diff
@@ -0,0 +1,64 @@
...
+            .get_block_transaction_events(AbsoluteBlockHeight {
+                height: height.try_into()?,
+            })
+            .await?
+            .response;
+        while let Some(summary) = block_summary.next().await.transpose()? {
+            let BlockItemSummaryDetails::Update(update) = summary.details else {
+                continue
+            };
+            let payload: ChainUpdatePayload = update.payload.into();
```
lassemand, line 48: This should be a vec. Consider using events_from_summary.

DOBEN (approved):

Issue #560, opened 2025-02-28
Frontend transaction details


lassemand:
DoD:

* Update frontend to point to fungy CCDScan
* Ensure it aligns with the old CCDScan

PR #604, opened 2025-03-24
Amount schedule

lassemand:
Closes https://github.com/Concordium/concordium-scan/issues/599
and finishes [CCD-201/frontend-transaction-details](https://github.com/Concordium/concordium-scan/issues/560)

## Purpose

Finish up amount schedule

### DIFF
--- a/backend-rust/src/migrations/m0020_amounts_schedule.rs
+++ b/backend-rust/src/migrations/m0020_amounts_schedule.rs
@@ -3,7 +3,9 @@ use crate::transaction_event::events_from_summary;
 use anyhow::Context;
 use async_graphql::futures_util::StreamExt;
 use concordium_rust_sdk::{
-    types::{AbsoluteBlockHeight, BlockItemSummaryDetails, TransactionType},
+    types::{
+        AbsoluteBlockHeight, AccountTransactionEffects, BlockItemSummaryDetails, TransactionType,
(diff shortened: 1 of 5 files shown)


backend-rust/src/migrations/m0020_amounts_schedule.rs
```diff
@@ -0,0 +1,83 @@
...
+        next_schema_version
+    ))?;
+
+    let mut client = v2::Client::new(endpoint.clone()).await?;
+
+    let rows = sqlx::query(
+        "
+            SELECT
+                block_height,
+                COUNT(*) AS update_count
```
lassemand, line 29: This will also be used.

DOBEN (approved):

backend-rust/src/migrations/m0020_amounts_schedule.rs
DOBEN: I don't see that this is used.

lassemand: It is not, I will make the change was what I meant. I was busy with fixing something else.

lassemand: It was not as trivially calculated as before (Type and effect can differ), and will remove it to remove noise.

PR #605, opened 2025-03-25
Effective time

lassemand:
## Purpose

Alter to be using correct effective time when the value is zero

https://github.com/Concordium/concordium-scan/blob/3eadf9eb6c642ac22203e4884e37f3479d5090d7/backend/Application/Api/GraphQL/Transactions/TransactionResultEvent.cs

### DIFF
--- a/backend-rust/CHANGELOG.md
+++ b/backend-rust/CHANGELOG.md
@@ -13,7 +13,7 @@ Database schema version: 19
 ### Fixed
 
 - Change `Versions::backend_versions` to `Versions::backend_version`.
-- When effective_time is zero then it translates into `Block::slot_time`.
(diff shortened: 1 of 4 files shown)


backend-rust/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Database schema version: 19
 ### Fixed
 
 - Change `Versions::backend_versions` to `Versions::backend_version`.
+- When effective_time is zero then it translates into `Block::slot_time`.
```
DOBEN, line 16:
```suggestion
- When `effective_time` is zero then it translates into `Block::slot_time`.
```

DOBEN (approved):

PR #606, opened 2025-03-25
Support partial migrations and introduce one for migration 18

limemloh:
## Purpose

Migrations are taking a while to run, so this PR adds support for partial migrations, which allows migrations to commit changes to the database before completing.


backend-rust/src/migrations.rs
```diff
@@ -13,8 +13,8 @@ mod m0008_canonical_address_and_transaction_search_index;
 mod m0010_fill_capital_bound_and_leverage_bound;
 mod m0014_baker_metrics;
 mod m0015_pool_rewards;
-mod m0018_payday_stake_information;
-mod m0019_chain_update_events;
+mod m0019_payday_stake_information;
```
lassemand, line 16: You do not have to update the versions? Nothing is applied yet

limemloh: I had to introduce schema version in between representing the partial migration, so everything after 18 gets shifted.

backend-rust/src/migrations/m0019_payday_stake_information.rs
```diff
@@ -37,26 +41,38 @@ pub async fn run(
...
         "SELECT DISTINCT payday_block_height
          FROM bakers_payday_pool_rewards
+         WHERE payday_block_height > $1
          ORDER BY payday_block_height ASC",
     )
+    .bind(last_processed_payday.unwrap_or(i64::MIN))
     .fetch_all(tx.as_mut())
     .await?;
+    // If no more paydays to process we are done migrating.
+    if paydays.is_empty() {
```
lassemand, line 59: Mentally I find a termination condition being placed in the middle of a script being pretty annoying.

That said this does not have to be long lived or anything.

An alternative way could be to treat paydays as a queue and then make this check in the end.

DOBEN (approved):

lassemand (approved):

backend-rust/src/migrations.rs
```diff
@@ -262,6 +274,35 @@ impl SchemaVersion {
...
+            SchemaVersion::StakedPoolSizeConstraint => false,
+            SchemaVersion::DelegatedStakeCap => false,
+            SchemaVersion::RankingByLotteryPower => false,
+            SchemaVersion::TrackRemovedBakers => false,
+            SchemaVersion::FixDelegatedStakeEarnings => false,
+            SchemaVersion::BakerMetrics => false,
+            SchemaVersion::PaydayPoolRewards => false,
+            SchemaVersion::PassiveDelegation => false,
+            SchemaVersion::RewardMetrics => false,
+            SchemaVersion::PaydayPoolStakePartial => true,
```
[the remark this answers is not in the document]
lassemand: I would like the same for destructive schema.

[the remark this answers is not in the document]
limemloh: This is intensional, so you get compile errors when introducing a new schema version and make sure to update this.

backend-rust/src/migrations/m0019_payday_stake_information.rs
limemloh: we do treat it as a queue.
It needs to be placed here, to ensure we only migrate to the next schema version when we start this migration with no paydays to process.

PR #607, opened 2025-03-25
Log level only apply for crate logs

limemloh:
## Purpose

Before setting `--log-level=debug` would result in debug logs from dependency, now it only applies to logs from this crate.
The old behavior can still be triggered by setting environment variable `RUST_LOG=debug`.

### DIFF
--- a/backend-rust/src/bin/ccdscan-indexer.rs
+++ b/backend-rust/src/bin/ccdscan-indexer.rs
@@ -16,6 +16,7 @@ use std::{net::SocketAddr, path::PathBuf, time::Duration};
 use tokio::net::TcpListener;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info};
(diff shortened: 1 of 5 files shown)


backend-rust/src/bin/ccdscan-indexer.rs
```diff
@@ -83,7 +84,17 @@ async fn main() -> anyhow::Result<()> {
         let _ = dotenvy::dotenv();
     }
     let cli = Cli::parse();
-    tracing_subscriber::fmt().with_max_level(cli.log_level).init();
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::fmt::layer())
+        .with(
+            tracing_subscriber::EnvFilter::builder()
+                .with_default_directive(
+                    format!("{}={}", env!("CARGO_PKG_NAME").replace('-', "_"), cli.log_level)
```
DOBEN, line 92: can we add a comment here. People might not directly understand what we filter  out here and why?

DOBEN (approved):

Issue #601, opened 2025-03-24
Fix `NaN` values in `TopDelegators` query


DOBEN:
**Task description**

Some `Nan` values appear at the front-end when switching the `TopDelegators` query to the new backend.

Rust-backend version > 0.1.38 
Front-end version: 1.7.8

![Image](https://github.com/user-attachments/assets/beb028c7-d818-425c-b91a-05339fa3856e)

PR #608, opened 2025-03-25
Fix corrupted delegators in database

DOBEN:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/601

There are a few `passiveDelegators` on stagenet that report NAN, because their `delegated_restake_earnings` value in the database is `NULL` while it shouldn't be `NULL`.
https://github.com/Concordium/concordium-scan/blob/main/backend-rust/src/graphql_api/account.rs#L358

The value was accidentially set to `NULL` as part of this migration:
https://github.com/Concordium/concordium-scan/pull/565/files#diff-9706a6ac0147c030922b969fef8dce859ee563f49089319271883aeda70994d3R61

The migration should have been:
```
UPDATE accounts
    SET delegated_target_baker_id = NULL
FROM latest_set_delegation
```

Only passive delegators that became passive delegators because their baker pool got removed are affected by the bug.
Among those delegators, only the delegators that had the database updated as part of the  `m0013-fix-removed-delegators-restake` file were affected  by the bug.

## Changes

- Add database migration 22 fixing some passive delegators that had no `delegated_restake_earnings` value set in the database.


backend-rust/src/migrations/m0013-fix-removed-delegators-restake.sql
```diff
@@ -32,6 +32,8 @@ WHERE
 
 
 -- Migration which fixes data due to bug in indexer where delegators which set their target to a pools pending for removal did not get updated (only relevant for blocks prior to Protocol Version 7).
+-- NOTE: This migration was meant to set `delegated_target_baker_id` to NULL (and not `delegated_restake_earnings`). This migration has been fixed and being
+-- re-executed in `m0019-fix-corrupted-passive-delegators.sql`.
```
limemloh, line 36:
```suggestion
-- re-executed in `m00xx-fix-corrupted-passive-delegators.sql`.
```

limemloh (approved):

limemloh (approved):

PR #610, opened 2025-03-25
Fix transaction count value

### LINKED ISSUE #609: `TransactionCount` value
Testing the hook `useAccountQuery`  with the new rust-backend revealed a change in definition of the   `TransactionCount` variable compared to the old-backend.

Old backend: Counting transactions that the account is involved in or effected by.

New backend: Counting transactions where the account is the sender of the transaction.
https://github.com/Concordium/concordium-scan/blob/ccdscan-backend/0.1.38/backend-rust/src/graphql_api/account.rs#L399

E.g. The front-end was built with the old definition in mind, which causes a few confusion now, e.g. the value `12` is shown but the drop down has actually 18 elements. 
Image

Plan is to apply the old definition but keeping the `nonce` exposed.
```
/// Number of transactions the account is invovled in.
    async fn transaction_count<'a>(&self, ctx: &Context<'a>) -> ApiResult<i64> {
        Ok(self.num_txs)
    }    

 /// Number of transactions where this account is used as sender. This is the `nonce` of the account.
/// This value is currently not used by the front-end and will need further optimization if intended to be used.
    async fn nonce<'a>(&self, ctx: &Context<'a>) -> ApiResult<i64> {
        let count = sqlx::query_scalar!(
            "SELECT COUNT(*) FROM transactions WHERE sender_index = $1",
            self.index
        )
        .fetch_one(get_pool(ctx)?)
        .await?;
        Ok(count.unwrap_or(0))
    }

```
https://github.com/Concordium/concordium-scan/blob/ccdscan-backend/0.1.38/backend-rust/src/graphql_api/account.rs#L307

DOBEN:
## Purpose

Closes #609

### DIFF
--- a/backend-rust/src/graphql_api/account.rs
+++ b/backend-rust/src/graphql_api/account.rs
@@ -400,8 +400,15 @@ impl Account {
         Ok(slot_time)
     }
 
-    /// Number of transactions where this account is used as sender.
-    async fn transaction_count<'a>(&self, ctx: &Context<'a>) -> ApiResult<i64> {
+    /// Number of transactions the account has been involved in or
+    /// affected by.
+    async fn transaction_count<'a>(&self) -> ApiResult<i64> { Ok(self.num_txs) }
+
+    /// Number of transactions where this account is the sender of the
+    /// transaction. This is the `nonce` of the account. This value is
+    /// currently not used by the front-end and the `COUNT(*)` will need further
+    /// optimization if intended to be used.
+    async fn nonce<'a>(&self, ctx: &Context<'a>) -> ApiResult<i64> {
         let count = sqlx::query_scalar!(
             "SELECT COUNT(*) FROM transactions WHERE sender_index = $1",
             self.index
--- a/backend-rust/src/migrations.rs
+++ b/backend-rust/src/migrations.rs
@@ -236,12 +236,14 @@ pub enum SchemaVersion {
     AmountSchedule,
     #[display("0022:Fix corrupted passive delegators")]
     FixCorruptedPassiveDelegators,
+    #[display("0023:Fix transaction count")]
+    FixTransactionCount,
 }
 impl SchemaVersion {
     /// The minimum supported database schema version for the API.
     /// Fails at startup if any breaking database schema versions have been
     /// introduced since this version.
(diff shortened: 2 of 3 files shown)


backend-rust/src/migrations.rs
```diff
@@ -236,12 +236,14 @@ pub enum SchemaVersion {
     AmountSchedule,
     #[display("0022:Fix corrupted passive delegators")]
     FixCorruptedPassiveDelegators,
+    #[display("0023:Fix transaction count")]
```
limemloh, line 239: This schema version is only relevant for the scheme of the database and not the API, so there is no need to introduce this

backend-rust/src/graphql_api/account.rs
```diff
@@ -400,8 +400,15 @@ impl Account {
...
-    async fn transaction_count<'a>(&self, ctx: &Context<'a>) -> ApiResult<i64> {
+    /// Number of transactions the account has been involved in or
+    /// affected by.
+    async fn transaction_count<'a>(&self) -> ApiResult<i64> { Ok(self.num_txs) }
+
+    /// Number of transactions where this account is the sender of the
+    /// transaction. This is the `nonce` of the account. This value is
+    /// currently not used by the front-end and the `COUNT(*)` will need further
+    /// optimization if intended to be used.
+    async fn nonce<'a>(&self, ctx: &Context<'a>) -> ApiResult<i64> {
```
lassemand, line 411: Unclear to me why this is called a nonce but sure. :) 

Nonce is something that is supposed to be used for a one time purpose.

I do not see how that is the number of transactions.

lassemand (approved):

backend-rust/src/graphql_api/account.rs
DOBEN: When an account sends a transaction and the transaction is finalized on the chain, the sequence number/nonce of that account is incremented. Hence, when we count the number of transactions where the queried account is the sender of the transaction, we get the same value as the `nonce` of the account.

Yes, it is called `nonce` because a transaction is signed and as such a transaction can only be finalized one time on-chain.

https://github.com/Concordium/concordium-node/blob/fc33d661178508b8eb7bcee6715169de7535ce59/concordium-consensus/src/Concordium/KonsensusV1/TreeState/Implementation.hs#L832

PR #614, opened 2025-03-26
Adjust default log filter when using --log-level

limemloh:


backend-rust/src/bin/ccdscan-indexer.rs
```diff
@@ -15,7 +15,7 @@ use sqlx::postgres::PgPoolOptions;
 use std::{net::SocketAddr, path::PathBuf, time::Duration};
 use tokio::net::TcpListener;
 use tokio_util::sync::CancellationToken;
-use tracing::{error, info};
+use tracing::{error, info, level_filters::LevelFilter};
```
DOBEN, line 18: not used

DOBEN (approved):

Issue #303, opened 2024-11-22
`input_parameter` display of `init_transactions`


DOBEN:
**Task description**
The current CCDScan (as well as other tooling https://github.com/Concordium/concordium-client/issues/291) doesn't display the `input_parameter/message` for `init_transactions`. The reason is that the `ContractInitializedEvent` from the node doesn't expose the value so this edge case was not handled previously. https://developer.concordium.software/concordium-grpc-api/#concordium.v2.ContractInitializedEvent 

In contrast, `contract_update_transactions` have the `input_parameter` available in the event.
https://developer.concordium.software/concordium-grpc-api/#concordium.v2.InstanceUpdatedEvent

Tasks:

Expand the CCDScan backend logic to extract the `input_parameter/message` e.g. via the SDK (not from the event) for `init_transactions`. Expand the frontend to display the `input_parameter/message` for  `init_transactions` as well:

Back-end functions:

```
#[ComplexObject]
impl ContractInitialized {
    // TODO: the send message/input parameter is missing and not exposed by the event of the node currently.
    async fn message(&self) -> ApiResult<String> {
    
        // TODO: decode `self.input_parameter` with the schema.

        Ok(input_parameter_decoded)
    }

    // TODO: the send message/input parameter is missing and not exposed by the event of the node currently.
    async fn message_as_hex(&self) -> ApiResult<String> {
        Ok(hex::encode(self.input_parameter))
    }
    ...
}
```

PR #616, opened 2025-03-26
Add input parameter decoding for init functions

DOBEN:
## Purpose

Closes https://github.com/Concordium/concordium-scan/issues/303

## Changes

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/backend-rust/src/transaction_event/smart_contracts.rs
+++ b/backend-rust/src/transaction_event/smart_contracts.rs
@@ -58,19 +58,21 @@ pub struct ContractInitialized {
     pub amount:            Amount,
     pub init_name:         String,
     pub version:           ContractVersion,
-    pub input_parameter:   Vec<u8>,
+    pub input_parameter:   Option<Vec<u8>>,
     // All logged events by the smart contract during the transaction execution.
     pub contract_logs_raw: Vec<Vec<u8>>,
 }
 
 #[ComplexObject]
 impl ContractInitialized {
-    async fn message_as_hex(&self) -> ApiResult<String> { Ok(hex::encode(&self.input_parameter)) }
+    async fn message_as_hex(&self) -> ApiResult<Option<String>> {
+        Ok(self.input_parameter.as_ref().map(hex::encode))
+    }
 
     async fn message<'a>(&self, ctx: &Context<'a>) -> ApiResult<Option<String>> {
-        if self.input_parameter.is_empty() {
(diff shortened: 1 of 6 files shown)


backend-rust/src/transaction_event/smart_contracts.rs
```diff
@@ -58,12 +58,59 @@ pub struct ContractInitialized {
...
     pub init_name:         String,
     pub version:           ContractVersion,
+    pub input_parameter:   Vec<u8>,
     // All logged events by the smart contract during the transaction execution.
     pub contract_logs_raw: Vec<Vec<u8>>,
 }
 
 #[ComplexObject]
 impl ContractInitialized {
+    async fn message_as_hex(&self) -> ApiResult<String> { Ok(hex::encode(&self.input_parameter)) }
```
lassemand, line 68: Why is the return value of this field not optional as is the case in message?

backend-rust/src/transaction_event/smart_contracts.rs
```diff
@@ -364,13 +411,18 @@ struct SchemaName {
...
     InputParameterReceiveFunction,
+    InputParameterInitFunction,
 }
 
 impl SmartContractSchemaNames {
     pub const EVENT: SchemaName = SchemaName {
         type_name:  "event",
         value_name: "contract log",
     };
+    pub const INPUT_PARAMETER_INIT_FUNCTION: SchemaName = SchemaName {
```
lassemand, line 422: So this seems to only be used in the error message which seems fine to me

backend-rust/src/transaction_event/smart_contracts.rs
lassemand, line 68: Looking at the interfaces of these two functions, I would expect message_as_hex to be the simple hex upon the message. It is not why?

Cant we just make input param be optional as well, and then call message here?

lassemand (approved):

backend-rust/src/transaction_event/smart_contracts.rs
DOBEN: I made it input parameter optional as suggested. We discussed `message_as_hex` and `message` function and how they are used in a meeting.

DOBEN: Screenshot from 2025-03-27 09-12-31

PR #620, opened 2025-03-26
Migrate transaction and account

lassemand:
## Purpose

Migrate what has already been implemented on the backend.


frontend/src/queries/useAccountQuery.ts
```diff
@@ -311,6 +311,7 @@ export const useAccountQuery = ({
 	const identifier = id?.value ? { id: id.value } : { address: address?.value }
 
 	const { data, fetching, error, executeQuery } = useQuery({
+		context: { url: useRuntimeConfig().public.apiUrlRust },
```
lassemand, line 314: We cannot migrate AccountRewardMetrics without migrating this.
 
The view seems to be all implemented.

limemloh (approved):
Don't forget the changelog

PR #650, opened 2025-03-31
Pool reward metrics

lassemand:
## Purpose

Finish pool rewards


lassemand:
I do not understand why this is not true, however the values in the database on this particular field seems way off.

frontend/src/queries/useBakerPoolRewardMetrics.ts
```diff
@@ -28,6 +28,7 @@ export const useBakerPoolRewardMetrics = (
 	period: Ref<MetricsPeriod>
 ) => {
 	const { data, executeQuery, fetching } = useQuery({
+		context: { url: useRuntimeConfig().public.apiUrlRust },
```
limemloh, line 31: This should be in the changelog of the frontend

limemloh (approved):

PR #659, opened 2025-04-03
Fix pagination issue for `Query::bakers`

limemloh:
Fix pagination issue for `Query::bakers` when page bounds start and end with the same value for the selected sorting, it would include an overlap for the nested sorting (usually the validator ID).

A connection with like below:
```
outer:  a       b          c
inner:    1 2 3   1 5 8 10
bounds:             [    ]
```
Would include `b.1` even though it is outside the bounds.


backend-rust/src/graphql_api/baker.rs
```diff
@@ -2033,11 +2121,18 @@ FROM bakers
...
 WHERE
     (
-      (($2::FLOAT8 IS NULL OR delegators_apy > $2) AND ($1::FLOAT8 IS NULL OR delegators_apy < $1))
-      OR ($1 IS NOT NULL AND delegators_apy = $1    AND bakers.id < $3)
-      OR ($1 IS NULL     AND delegators_apy IS NULL AND bakers.id < $3)
-      OR ($2 IS NOT NULL AND delegators_apy = $2    AND bakers.id > $4)
-      OR ($2 IS NULL     AND delegators_apy IS NULL AND bakers.id > $4)
+      (COALESCE(delegators_apy, 0) > $2::FLOAT8 AND COALESCE(delegators_apy, 0) < $1::FLOAT8)
+      OR (
+          $1 > $2
```
DOBEN, line 2126: Shouldn't make a difference but maybe we just align these queries.
```suggestion
            $1 != $2
```

DOBEN (approved):

backend-rust/src/graphql_api/baker.rs
limemloh: yes, this was an earlier iteration, will fix

backend-rust/src/graphql_api/baker.rs
```diff
@@ -686,9 +686,28 @@ impl Baker {
...
+                           AND (
+                                -- Start inner bound for page.
+                                (pool_total_staked = $1 AND bakers.id < $3)
+                                -- End inner bound for page.
+                                 OR (pool_total_staked = $2 AND bakers.id > $4)
+                           )
+                       )
+                       -- When outer bounds are equal, use one filter for both bounds.
+                       OR (
+                           $1 = $2
```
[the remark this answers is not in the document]
limemloh: right, we can do this in another PR

Issue #600, opened 2025-03-24
Migrate remaining hooks


DOBEN:
**Task description**
Migrate hook queries at the frontend to use the new backend. Remaining hooks:

**Misc**

- [ ] useSearchQuery
- [X] usePaydayStatusQuery
- [X] useVersionsQuery

**Transactions**

- [X] useTransactionReleaseSchedule
- [X] useTransactionQuery

**Passive delegation**

- [X] usePassiveDelegationQuery
- [ ] usePassiveDelegationPoolRewardMetrics

**Bakers**

- [ ] useBakerTransactionsQuery
- [ ] useBakerRewardsQuery
- [ ] ~~useBakerPoolRewardsQuery~~  *Never used by the frontend*
- [X] useBakerPoolRewardMetrics
- [ ] useBakerDelegatorsQuery

**Accounts**

- [X] useTopDelegatorsQuery
- [ ] useAccountsMetricsQuery
- [ ] useAccountRewardMetricsQuery
- [ ] useAccountQuery
- [ ] useAccountsListQuery

**Subscriptions**

- [ ] useAccountsUpdatedSubscription
- [ ] useBlockSubscription

Issue #254, opened 2024-09-27
Implement SearchResult::contracts


limemloh:
Implement `SearchResult::contracts` in the Rust GraphQL API and relevant types.

For reference, the [C# GraphQL schema can be found here](https://github.com/Concordium/concordium-scan/blob/main/backend/Tests/Api/GraphQL/\__snapshots_\_/committed-schema.verified.graphql).

DOBEN:
address in PR
https://github.com/Concordium/concordium-scan/pull/664

Issue #251, opened 2024-09-27
Implement SearchResult::modules


limemloh:
Implement `SearchResult::modules` in the Rust GraphQL API and relevant types.

PR #665, opened 2025-04-04
Change frontend to only use rust backend

limemloh:
## Purpose

#600 

Update frontend to use the rust backend for every remaining query.

This is still blocked by #257, #251, #254, #454 and missing endpoint `GET /rest/export/statement`

### DIFF
--- a/frontend/nuxt.config.ts
+++ b/frontend/nuxt.config.ts
@@ -8,17 +8,11 @@ export default defineNuxtConfig({
 		public: {
 			version: process.env.npm_package_version,
 			// URL to use when sending GraphQL queries to the CCDscan API.
-			// (env NUXT_PUBLIC_API_URL)
-			apiUrl: 'http://localhost:5090/graphql',
-			// URL to use when using websockets in GraphQL CCDscan API.
-			// (env NUXT_PUBLIC_WS_URL)
-			wsUrl: 'ws://localhost:5090/graphql',
-			// URL to use when sending GraphQL queries to the CCDscan API.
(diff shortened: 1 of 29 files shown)


DOBEN (approved):

frontend/nuxt.config.ts
```diff
@@ -8,17 +8,11 @@ export default defineNuxtConfig({
...
-			// URL to use when using websockets in GraphQL CCDscan API.
-			// (env NUXT_PUBLIC_WS_URL)
-			wsUrl: 'ws://localhost:5090/graphql',
-			// URL to use when sending GraphQL queries to the CCDscan API.
 			// (env NUXT_PUBLIC_API_URL_RUST)
 			apiUrlRust: 'http://localhost:8000/api/graphql',
 			// URL to use when using websockets in GraphQL CCDscan API.
 			// (env NUXT_PUBLIC_WS_URL_RUST)
-			wsUrlRust: 'ws://localhost:8000/api/graphql',
+			wsUrlRust: 'ws://localhost:8000/ws/graphql',
```
DOBEN, line 15:
```suggestion
			wsUrlRust: 'ws://localhost:8000/api/graphql',
```

limemloh: I think the url for the websocket is correct https://github.com/Concordium/concordium-scan/blob/cd7a0d2a2a73c25483e7d02be2577ee120a4c31a/backend-rust/src/graphql_api.rs#L191

DOBEN: Then the env files at the front-end are wrong.

PR #667, opened 2025-04-07
Support exporting accounts statements from rust backend

limemloh:

### DIFF
--- a/backend-rust/src/rest_api.rs
+++ b/backend-rust/src/rest_api.rs
@@ -1,0 +1,144 @@
+use std::sync::Arc;
+
+use crate::graphql_api::AccountStatementEntryType;
+use axum::{
+    extract::{Query, State},
+    http::HeaderName,
+    response::{AppendHeaders, IntoResponse},
+    routing::get,
+    Router,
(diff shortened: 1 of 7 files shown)


backend-rust/src/rest_api.rs
```diff
@@ -0,0 +1,144 @@
...
+                account_statements.account_balance,
+                entry_type as "entry_type: AccountStatementEntryType"
+            FROM accounts
+                JOIN account_statements
+                    ON accounts.index = account_statements.account_index
+                JOIN blocks
+                    ON blocks.height = account_statements.block_height
+            WHERE
+                accounts.address = $1
+                AND slot_time > $2
```
DOBEN, line 65: I think we need to include the boundary in a bucket.
```suggestion
                AND slot_time >= $2
```

DOBEN (approved):

backend-rust/src/rest_api.rs
limemloh: Yes, that makes sense.
.NET actually includes both, so I'll do the same.

PR #673, opened 2025-04-10
Finish reward metrics

lassemand:
## Purpose

Alternate leading index for reward metrics


backend-rust/CHANGELOG.md
```diff
@@ -12,6 +14,9 @@ All notable changes to this project will be documented in this file.
 - Add REST API `rest/balance-statistics/latest` for querying total amount in the network, used by external parties.
 - Add `ccdscan-api` option `--statement-timeout-secs` (env `CCDSCAN_API_DATABASE_STATEMENT_TIMEOUT_SECS`) for configuring a statement timeout the database connections and abort any statement that takes more than the specified amount of time. Defaults to 30 seconds.
 
+### Fixed
+
+- Alternate leading index on reward metrics
```
DOBEN, line 19: Since there exist `alternate index` (secondary index), it is confusing.

```suggestion
- Reordered the primary key on the `reward_metrics` table from (block_slot_time, account_index) to (account_index, block_slot_time) to improve query performance.
```

DOBEN (approved):

backend-rust/src/migrations/m0028_reindex_reward_metrics.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE metrics_rewards
+DROP CONSTRAINT metrics_rewards_pkey;
+
+ALTER TABLE metrics_rewards
+ADD CONSTRAINT metrics_rewards_pkey
+PRIMARY KEY (block_slot_time, account_index);
```
limemloh, line 6: I think we need the current index for the query of rewards per account, so maybe just introduce a new index instead of removing the old one.
```sql
CREATE INDEX metrics_rewards_time_index ON metrics_rewards (block_slot_time);
```

lassemand: We do not. If it is fast enough to add all the rows in the timeslot together obviously it is also fast enough for a subset of them.

That said I have also tested the above claim on the mainnet data, and it is just fine. So no we do not need to maintain another. :)

backend-rust/src/migrations/m0028_reindex_reward_metrics.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE metrics_rewards
```
limemloh, line 1: Please document the reason for the change, otherwise we might revert it in the future without understanding why it was here in the first place.

PR #679, opened 2025-04-15
Transactions query

lassemand:
## Purpose

Fix accounts transactions. Please note that the accounts query is still slow on mainnet, but that is for another reason which will come in another PR.

Fix issue on account metrics


backend-rust/src/graphql_api/account.rs
```diff
@@ -982,14 +982,14 @@ impl Account {
                     success,
                     events as "events: sqlx::types::Json<Vec<Event>>",
                     reject as "reject: sqlx::types::Json<TransactionRejectReason>"
-                FROM transactions
+                FROM transactions t
+                LEFT JOIN (
```
lassemand, line 986: This fixes the result, allowing the optimiser to only query the list once instead

backend-rust/src/graphql_api/account_metrics.sql
```diff
@@ -23,17 +23,15 @@ FROM
         $2::interval
     ) AS bucket_time
 LEFT JOIN LATERAL (
-    -- Selects the index at or before the start of the bucket.
     SELECT accounts.index
     FROM accounts
     LEFT JOIN transactions on transaction_index = transactions.index
     LEFT JOIN blocks ON transactions.block_height = height
-    WHERE slot_time <= bucket_time
```
lassemand, line 31: This could potentially mean we are operating on the same row twice providing errors

limemloh: I am not sure whether this is correct.
I think one of the bucket edges needs to include the `slot_time` otherwise a block for the exact `slot_time` would not be included in any bucket.

lassemand: So lets take an example.

We say we have two timezone 1.0 and 2.0 respectively. The pairs (with the first being the bucket timezone) can be fetched to be of the following values:

((0.9, 4), (1.9, 6)), ((2.0, 8), (\{unlimited}, 10))

This would mean we calculates 2 in the first bucket and 2 in the last. But the expected total value is 6. This is a false claim. The correct would be to use the same point as the previous after bucket, which would mean that we instead had 

((0.9, 4), (1.9, 6)), ((1.9, 6), (\{unlimited}, 10)) which gives used buckets having value 2 and 4, totalling 6 which is the correct value.

limemloh: Right, I see now that this is the correct behavior :+1:

limemloh (approved):

backend-rust/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- Optimise `Account::transactions` performance
+- Fix potential issue on `accountMetrics`
```
limemloh, line 10: More details here would be nice

DOBEN (approved):

PR #682, opened 2025-04-22
Account metrics performance

lassemand:
## Purpose

Account metrics performance does not leverage that we know the ids are strictly increasing. This should help map a slot_time into a account index while simply making lookups on the latest known values. Tested against mainnet dataset.


backend-rust/src/graphql_api/account_metrics.sql
```diff
@@ -1,42 +1,54 @@
...
-        now() - $1::interval,
-        -- The final bucket starts 6 hours ago, since the bucket time is the start of the bucket.
-        now() - $2::interval,
-        -- Each bucket is seperated by 6 hours.
-        $2::interval
+WITH
+  buckets AS (
+    SELECT generate_series(
+      now() - $1::interval,
+      now() - $2::interval,
```
lassemand, line 10: I should reuse this date instead of redefining

backend-rust/src/graphql_api/account_metrics.sql
```diff
@@ -1,42 +1,54 @@
...
+      SELECT MAX(a.index)
+      FROM accounts a
+      WHERE a.transaction_index <= th.tx_start
+    ), 0
+  ) AS "start_index!",
+  COALESCE(
+    (
+      SELECT MAX(a.index)
+      FROM accounts a
+      WHERE a.transaction_index <= th.tx_end
```
lassemand, line 50: I do not entirely understand why this is fast enough when not having an index on transaction_index. It does seem to perform much better than dotnet version even without the index though. So lets skip it.

lassemand: I can leave a comment that this is a quick win for performance gains.

backend-rust/src/graphql_api/account_metrics.rs
```diff
@@ -90,12 +69,13 @@ impl QueryAccountMetrics {
...
-        let y_accounts_created = rows
-            .iter()
-            .map(|r| r.end_index.expect("coalesced") - r.start_index.expect("coalesced"))
-            .collect();
+        let y_last_cumulative_accounts_created: Vec<i64> =
+            rows.iter().map(|r| r.end_index).collect();
+        let y_accounts_created: Vec<i64> =
+            rows.iter().map(|r| r.end_index - r.start_index).collect();
+        let last_cumulative_accounts_created =
+            *y_last_cumulative_accounts_created.last().unwrap_or(&0);
```
lassemand, line 77: This default is a mistake

backend-rust/src/graphql_api/account_metrics.rs
```diff
@@ -90,12 +69,13 @@ impl QueryAccountMetrics {
...
 
         let x_time = rows.iter().map(|r| r.bucket_time).collect();
-        let y_last_cumulative_accounts_created =
-            rows.iter().map(|r| r.end_index.expect("coalesced")).collect();
-        let y_accounts_created = rows
-            .iter()
-            .map(|r| r.end_index.expect("coalesced") - r.start_index.expect("coalesced"))
-            .collect();
+        let y_last_cumulative_accounts_created: Vec<i64> =
+            rows.iter().map(|r| r.end_index).collect();
```
lassemand, line 73: Same goes for this. If nothing is returned this is zero which is obviously wrong. I do not know where this is used as of yet though.

backend-rust/src/graphql_api/account_metrics.sql
lassemand: I am refraining from doing it because from a user perspective performance is more than fine, so it only increases the amount of indexing which has to occur.

As long as this is fine from a user perspective it is fine and good, and we should not waste resources on maintaining something which is not needed. 

So my point being:
If you want to create such an index, in case the amount of accounts suddenly increases much, then I will just create the index instead of creating the task. 

As it sounds like you want that I will do that, as I kind of get the argument both ways.

backend-rust/src/graphql_api/account_metrics.rs
lassemand: I was wrong about this. It will fetch the latest even though it is from another bucket, and even if it is from outside the interval. Returning 0 is exactly correct.

backend-rust/src/graphql_api/account_metrics.sql
limemloh: Not doing it now, means that this might come back to us as technical dept, delaying other projects.
Since it is so low effort, and you already have the mental model build, I would prefer to go for it already.

backend-rust/src/graphql_api/account_metrics.rs
limemloh: Maybe comment above to why this is correct, it does take some reasoning to arrive at

backend-rust/src/migrations/m0029_index_account_transactions.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX accounts_transaction_index
```
limemloh, line 1: Please add a comment to when/why this index is used

limemloh (approved):

PR #687, opened 2025-04-24
Account transctions

lassemand:
## Purpose

- Fix account transactions using DESC ordering per default instead of ASC
- Fix account transaction altering index to be account index first
- Fix account transactions queries to be using WHERE instead of join


backend-rust/src/migrations/m0031_reindex_affected_accounts.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE affected_accounts
```
limemloh, line 1: Please document why this change was needed

limemloh (approved):

PR #689, opened 2025-04-24
Architectural diagrams for ccdscan

lassemand:


DOBEN (approved):
Suggestion:
- To add the diagram to the READMe.
- Connection from `GraphQL` to `Node Collector backend` describing the connection e.g. `Call backend for node information` (so it is clear that there is a request and a response).
- To signal that there are several `GRAPHQL API` services (e.g. by replicating the icon behind each other).

PR #702, opened 2025-05-07
fix: COR-1378 - added and deprecated 'TotalAmounUnlocked' for backwards compatability, 'TotalAmountUnlocked' is now preferred for the balance statistics api

squirer:
## Purpose

The balance statistics API used to accept a typo parameter `totalAmounUnlocked`, this was re-added to the API for backwards compatibility with a comment that it is deprecated and to prefer to use `totalAmountUnlocked` correctly spelled going forward.

## Changes

- Added extra parameter to the Balance enum for `TotalAmounUnlocked` with a deprecated comment
- Added extra parameter into the API itself for the parameter check

## Checklist

      hard-to-understand areas.


backend-rust/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Changed
```
limemloh, line 7:
```suggestion
### Added
```

backend-rust/src/rest_api.rs
```diff
@@ -178,6 +180,9 @@ enum Balance {
     TotalAmount,
     TotalAmountCirculating,
     TotalAmountUnlocked,
+    TotalAmounUnlocked, /* deprecated: please use 'totalAmountUnlocked' going forward over
+                         * 'totalAmounUnlocked' as 'totalAmounUnlocked' is now deprecated and
+                         * will be removed in a future release */
```
limemloh, line 185: This comment makes me worry someone might choose to remove it based on time passed, and it should instead be removed when we see no more activity for this in the API.
```suggestion
    TotalAmounUnlocked, /* deprecated: please use 'totalAmountUnlocked' going forward over
                         * 'totalAmounUnlocked' as 'totalAmounUnlocked' is now deprecated and
                         * should be removed when no more activity is registered for this field */
```

limemloh (approved):

PR #705, opened 2025-05-08
Improve performance of transaction metrics query

limemloh:
## Purpose

Closes COR-1382.

Simplify `Query::transaction_metrics` removing a query which performed badly.


backend-rust/src/graphql_api/transaction_metrics.rs
```diff
@@ -45,61 +45,38 @@ impl QueryTransactionMetrics {
...
             .as_duration()
             .try_into()
             .map_err(|e| ApiError::DurationOutOfRange(Arc::new(e)))?;
-
-        let cumulative_transaction_count_before_period = sqlx::query_scalar!(
-            "SELECT cumulative_num_txs
-            FROM blocks
-            WHERE slot_time >= (now() - $1::interval)
-            ORDER BY height ASC
-            LIMIT 1",
```
limemloh, line 66: This was the query causing issues, since it is `order by height` and filter on `slot_time` postgres could not optimize this properly.
Instead it should have been `order by slot_time`, but since the information is part of the rows below it can be skipped entirely.

DOBEN (approved):

PR #714, opened 2025-05-12
Add partial contract index search (COR-1354)

DOBEN:
## Changes

- Added partial contract index search `SearchResult::contracts` and added supporting database index.
- Fix display of `protocol-level token dashboard` which was not correctly filtered out.


backend-rust/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
```
limemloh, line 21: Could we start having the database schema version in the changelog again, it seems that we have forgotten this in the past few releases.

limemloh (approved):

PR #719, opened 2025-05-13
Reorganize indexer.rs into several files

limemloh:
## Purpose

The `backend-rust/src/indexer.rs` file has grown big and needs to be more organized, this PR splits it up, with minimal code changes.


DOBEN (approved):
Looks much more organised now.

PR #720, opened 2025-05-16
Fix performance issue in baker metrics

limemloh:
## Purpose

Close COR-1394

## Changes

- Change baker metrics query to take better advantage of slot_time index.


backend-rust/src/graphql_api/baker_metrics.sql
```diff
@@ -1,8 +1,29 @@
...
+     FROM metrics_bakers
+     WHERE block_height < after_bucket.height
+     ORDER BY block_height DESC
+     LIMIT 1), 0) as "added_after!",
+    COALESCE((SELECT
+        total_bakers_removed
+     FROM metrics_bakers
+     WHERE block_height < after_bucket.height
+     ORDER BY block_height DESC
+     LIMIT 1), 0) as "removed_after!"
```
DOBEN, line 26: Some further optimization could be to have only two sub-queries instead of the four ones (meaning one query for the `after_bucket` block and one for the `before_bucket` block. Each query would query two rows `total_bakers_removed/total_bakers_removed`.

DOBEN (approved):

backend-rust/src/graphql_api/baker_metrics.sql
limemloh: I tried but the subqueries may only return one column in this scenario.

PR #749, opened 2025-06-27
plt dashboard plt-data dynamic

Pratap2018:
## Purpose

Plt dashboard is dynamic (with limited data)

## Changes

Using nodes grpc endpoint to populate the plt dashboard data.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/src/graphql_api/stable_coin.rs
+++ b/backend/src/graphql_api/stable_coin.rs
@@ -1,13 +1,16 @@
+use crate::address::AccountAddress;
+use anyhow::Context as AnyhowContext;
 use async_graphql::{Context, Object, SimpleObject};
 use chrono::{Duration, TimeZone, Utc};
+use clap::Parser;
+use concordium_rust_sdk::v2::{self, BlockIdentifier};
+use futures::StreamExt;
 use serde::Deserialize;
 use std::{
     collections::{HashMap, HashSet},
     fs::File,
     io::BufReader,
 };
-
-use crate::address::AccountAddress;
 
 pub type DateTime = chrono::DateTime<chrono::Utc>;
 
@@ -166,6 +169,16 @@ impl StableCoin {
     }
 }
 
+#[derive(Parser)]
+struct App {
+    #[arg(
+        long = "node",
+        env = "CCDSCAN_INDEXER_GRPC_ENDPOINTS",
+        default_value = "http://localhost:20000"
+    )]
+    endpoint: v2::Endpoint,
+}
+
 #[derive(Default)]
 pub(crate) struct QueryStableCoins;
 
@@ -272,7 +285,6 @@ impl QueryStableCoins {
             Self::load_transactions(),
         );
         let coin = stablecoins.iter_mut().find(|coin: &&mut StableCoin| coin.symbol == symbol)?;
-
         coin.holdings = coin.top_holders(limit, min_quantity);
         coin.transactions = coin.last_n_transactions(last_n_transactions);
         Some(coin.clone())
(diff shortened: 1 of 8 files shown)


backend/src/graphql_api/stable_coin.rs
```diff
@@ -166,6 +169,16 @@ impl StableCoin {
     }
 }
 
+#[derive(Parser)]
+struct App {
+    #[arg(
+        long = "node",
+        env = "CCDSCAN_INDEXER_GRPC_ENDPOINTS",
```
DOBEN, line 176: One general design decision when creating the `indexer` service and the `graphQL API` service was that we decided to NOT have a node connection in the `graphQL API` service. That decision has pros/cons. 

One of the advantages was that it made it easier for us (devops) to run several instances of the GraphQL API services, without having dedicated nodes attached to them (no need to manage nodes for these services).
A disadvantage is that the data you need here would need to be collected by the indexer at every `createPLT` tx event and a table in postgres would collect all `tokenInfo` to provide it to the `grpahQL API` service.

Let me discuss that with the team and I will come back to you, but we might want to keep or break that design decision.

backend/src/graphql_api/stable_coin.rs
```diff
@@ -425,4 +434,57 @@ impl QueryStableCoins {
...
+        let mut response = client
+            .get_token_list(&BlockIdentifier::LastFinal)
+            .await
+            .context("Failed to get token list")
+            .unwrap();
+        let mut coins: Vec<StableCoin> = vec![];
+
+        while let Some(token_id) = response.response.next().await.transpose().unwrap() {
+            let token_info =
+                client.get_token_info(token_id.clone(), BlockIdentifier::LastFinal).await.unwrap();
```
limemloh, line 457: Don't unwrap, use `.context("<insert-proper-error>")` instead if we are still going with `anyhow`,
Similarly for every unwrap below

backend/src/graphql_api/stable_coin.rs
```diff
@@ -425,4 +434,57 @@ impl QueryStableCoins {
...
+
+    async fn live_stablecoins(&self, _ctx: &Context<'_>) -> anyhow::Result<Vec<StableCoin>> {
+        let endpoint_env = env::var("CCDSCAN_INDEXER_GRPC_ENDPOINTS");
+        if endpoint_env.is_err() {
+            return Err(anyhow::anyhow!(
+                "Stablecoin data is not available. Only available in Devnet"
+            ));
+        }
+        let app = App::parse();
+        let mut client = v2::Client::new(app.endpoint).await.context("Failed to create client")?;
```
limemloh, line 446: No need to do the `App::parse()` here, which is also only designed to be used from the main thread
```suggestion
        let endpoint: v2::Endpoint = endpoint_env.context("Stablecoin data is not available. Only available in Devnet")?.parse().context("Failed parsing the node endpoint from env CCDSCAN_API_GRPC_ENDPOINT")?;
        let mut client = v2::Client::new(endpoint).await.context("Failed to create client")?;
```

backend/src/graphql_api/stable_coin.rs
```diff
@@ -425,4 +434,57 @@ impl QueryStableCoins {
         );
         txn_summary
     }
+
+    async fn live_stablecoins(&self, _ctx: &Context<'_>) -> anyhow::Result<Vec<StableCoin>> {
+        let endpoint_env = env::var("CCDSCAN_INDEXER_GRPC_ENDPOINTS");
```
limemloh, line 439: We should not reuse the env variables from the indexer
```suggestion
        let endpoint_env = env::var("CCDSCAN_API_GRPC_ENDPOINTS");
```

backend/src/graphql_api/stable_coin.rs
```diff
@@ -425,4 +434,57 @@ impl QueryStableCoins {
...
+            let token_state = token_info.response.token_state;
+            let token_module_state = token_state.decode_module_state().unwrap();
+            let issuer = serde_json::to_value(&token_module_state.governance_account).unwrap()
+                ["address"]
+                .as_str()
+                .unwrap_or("")
+                .to_string();
+            let name = token_module_state.name;
+
+            let total_supply = token_state.total_supply.to_string().parse::<f64>().unwrap() as i64;
```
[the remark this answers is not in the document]
Pratap2018: We are not doing any calculation , this data is integer because we will only show graph in the ui, and this is temporary

backend/src/graphql_api/stable_coin.rs
```diff
@@ -425,4 +434,57 @@ impl QueryStableCoins {
         );
         txn_summary
     }
+
+    async fn live_stablecoins(&self, _ctx: &Context<'_>) -> anyhow::Result<Vec<StableCoin>> {
```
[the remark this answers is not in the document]
Pratap2018: yes this is temporary implementation

PR #761, opened 2025-07-04
Feature/cor 1482 investigated staked amount issue

squirer:
## Purpose

Propose changes for changing how staked amounts and pool staked amount are inserted in the DB

## Changes

Query the node to populate validator hashmap of staking information, to now be used instead of special transaction outcomes taking from the DB and adding to whatever was previously stored

## Checklist

      hard-to-understand areas.


backend/src/indexer/block.rs
```diff
@@ -265,6 +278,21 @@ impl PreparedBlock {
         }
         self.statistics.save(tx).await?;
         self.special_transaction_outcomes.save(tx).await?;
+
+        // process validator updates for staked amount and pool total staked amount
+        for (validator_id, validator_staking_information) in &self.validator_stake_updates {
```
limemloh, line 283: If you build up the information correctly, you can update these in a single sql query, which would perform better.
For inspiration look for places were we use `UNNEST`

backend/src/indexer/block_preprocessor.rs
```diff
@@ -207,7 +210,7 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
                     RewardsOverview::V1 {
                         total_staked_capital,
                         ..
-                    } => *total_staked_capital,
+                    } => (*total_staked_capital, HashMap::new()),
```
limemloh, line 213: We would want to fetch the baker stake here as well, otherwise no stake changes starting from `RewardsOverview::V1`  (P4 I think)

backend/src/indexer/block_preprocessor.rs
```diff
@@ -294,21 +298,53 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
 pub async fn compute_total_stake_capital(
     client: &mut v2::Client,
     block_height: v2::BlockIdentifier,
-) -> v2::QueryResult<Amount> {
+) -> v2::QueryResult<(Amount, HashMap<i64, ValidatorStakingInformation>)> {
```
limemloh, line 301: Since we are only inserting and then traversing and not really using the properties of a hashmap, a `Vec<(i64, ValidatorStakingInformatio)>` would mean less overhead.
If you go for the single SQL statement to insert it should probably be several `Vec`s

backend/src/indexer/block_preprocessor.rs
```diff
@@ -291,36 +299,114 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
+
+                // add validator staking structure into the vector
+                validator_staking_info_vector.push((baker_index, validator_staking_info));
+
+                // baker pool info has the all pool total capital, so we just take what we find
+                // and set it here to be returned
+                all_pool_total_capital = baker_pool_info.response.all_pool_total_capital;
+            }
+        } else {
+            // When the Pool information is not found (could be due to earlier protocol
```
squirer, line 360: @DOBEN - here the else block will now process as before - as pool info is unavailable, we will go to the accountInfo for the staked amount.

I also set the pool amount for this validator to the same value (account stake)

backend/src/indexer/block_preprocessor.rs
```diff
@@ -291,36 +299,114 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
+            })?;
+
+            let validator_staking_info: ValidatorStakingInformation = ValidatorStakingInformation {
+                stake_amount:             validator_stake.micro_ccd,
+                pool_total_staked_amount: validator_stake.micro_ccd,
+            };
+
+            validator_staking_info_vector.push((baker_index, validator_staking_info));
+
+            all_pool_total_capital += validator_stake;
```
squirer, line 391: all pool total capital as before is now adding the stake from each validator found

backend/src/indexer/block.rs
```diff
@@ -265,6 +279,40 @@ impl PreparedBlock {
...
+
+            for (validator_id, validator_staking_info) in &self.validator_stake_updates {
+                ids.push(*validator_id);
+                staked.push(validator_staking_info.stake_amount as i64);
+                pool_staked.push(validator_staking_info.pool_total_staked_amount as i64);
+            }
+
+            debug!(
+                "Updating bakers table staked and pool_total_staked now. ids: {:?} - staked: \
+                 {:?}, pool_staked: {:?}",
```
DOBEN, line 298:
```suggestion
                "Updating bakers table `staked` and `pool_total_staked` values now. ids: {:?} - staked: \
                 {:?}, pool_staked: {:?}",
```

backend/src/indexer/block.rs
```diff
@@ -265,6 +279,40 @@ impl PreparedBlock {
         }
         self.statistics.save(tx).await?;
         self.special_transaction_outcomes.save(tx).await?;
+
+        // gather vectors for the update query
+        let validator_updates_length = self.validator_stake_updates.len();
+        if validator_updates_length > 0 {
+            let mut ids = Vec::with_capacity(validator_updates_length);
+            let mut staked = Vec::with_capacity(validator_updates_length);
+            let mut pool_staked = Vec::with_capacity(validator_updates_length);
```
DOBEN, line 288:
```suggestion
            let mut pool_total_staked = Vec::with_capacity(validator_updates_length);
```

backend/src/indexer/block/special_transaction_outcomes.rs
```diff
@@ -655,44 +634,7 @@ impl RestakeEarnings {
...
-            } else {
-                0..=1
-            };
-            // Account is delegating.
-            if let (true, Some(pool)) = (restake, account_row.delegated_target_baker_id) {
-                // If restake is enabled and the target is a validator pool (not the passive
-                // pool) and we update the pool stake.
-                sqlx::query!(
-                    "UPDATE bakers
-                         SET pool_total_staked = pool_total_staked + $2
```
DOBEN, line 670: There are many other areas in the code base remaining where we update `SET pool_total_staked =` based on some events (e.g. delegation and baker events). Would we need to remove these updates as well?

backend/src/indexer/block_preprocessor.rs
```diff
@@ -194,22 +196,27 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
-                        total_staked_capital,
                         ..
-                    } => *total_staked_capital,
+                    } => {
+                        compute_validator_staking_information(
+                            &mut client3,
+                            v2::BlockIdentifier::AbsoluteHeight(fbi.height),
+                        )
+                        .await?
+                    }
```
DOBEN, line 217: since we don't do any difference between the `rewardsOverview::V0` and `rewardsOverview::V1` arms, I don't think we need the match statement anymore.

```
       compute_validator_staking_information(
                            &mut client3,
                            v2::BlockIdentifier::AbsoluteHeight(fbi.height),
                        )
                        .await?
```

DOBEN: Question is do we then need the query to `tokenomics_info` endpoint? Might still be needed.

backend/src/indexer/block_preprocessor.rs
```diff
@@ -291,36 +299,114 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
 /// Compute the total stake capital by summing all the stake of the bakers.
 /// This is only needed for older blocks, which does not provide this
```
DOBEN, line 300: it is also used by newer blocks now.

backend/src/indexer/block_preprocessor.rs
```diff
@@ -291,36 +299,114 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
-            .get_account_info(&v2::AccountIdentifier::Index(baker_id.id), block_height)
-            .await?
-            .response;
-        total_staked_capital += account_info
-            .account_stake
-            .context("Expected baker to have account stake information")
-            .map_err(v2::RPCError::ParseError)?
-            .staked_amount();
+        // Try to get the pool info for a given Baker
+        if let Ok(baker_pool_info) = client.get_pool_info(block_height, baker_id).await {
```
DOBEN, line 312: As this function `get_pool_info` only returns data starting from protocol version 4. I think we should add a comment here why we need to split into two cases and mention that the split happen at protocol version 4.

   /// Get information about a given pool at the end of a given block.
    /// If the block does not exist or is prior to protocol version 4 then
    /// [`QueryError::NotFound`] is returned.

backend/src/indexer/block_preprocessor.rs
DOBEN: please be spicific which protocol version. Maybe worth to mention that `delegation` didn't exist and as such prior to protocol version 4 and as such `stake_amount` =`pool_total_staked_amount`.

backend/src/indexer/block_preprocessor.rs
```diff
@@ -291,36 +299,114 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
+            {
+                // stake for the baker
+                let validator_stake = active_baker_pool_status.baker_equity_capital;
+
+                // delegated amount
+                let delegated_capital = active_baker_pool_status.delegated_capital;
+
+                // total stake of the pool for this baker (baker_equity_capital +
+                // delegated_capital)
+                let total_stake_for_this_baker = validator_stake + delegated_capital;
```
DOBEN, line 324:
```suggestion
                let total_stake_for_this_validator = validator_stake + delegated_capital;
```

backend/src/indexer/block_preprocessor.rs
```diff
@@ -291,36 +299,114 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
+
+                debug!(
+                    "Computing validator staking information for validator id: {} - validator \
+                     stake: {}, delegated stake: {}, total stake: {}, total pool: {}",
```
DOBEN, line 328:
```suggestion
                     stake: {}, delegated stake: {}, total stake: {}, all pools: {}",
```

backend/src/indexer/genesis_data.rs
```diff
@@ -31,14 +31,17 @@ pub async fn save_genesis_data(
...
         }
         RewardsOverview::V1 {
-            total_staked_capital,
             ..
-        } => i64::try_from(total_staked_capital.micro_ccd())?,
+        } => {
+            let (total_staked_capital, _) =
+                compute_validator_staking_information(&mut client, genesis_height).await?;
+            i64::try_from(total_staked_capital.micro_ccd())?
+        }
```
DOBEN, line 44: same

backend/src/indexer/block_preprocessor.rs
DOBEN, line 312: In case this becomes a bottleneck, we could try to `join!()` the `get_pool_info` together and execute them in parallel.

backend/src/indexer/block.rs
```diff
@@ -30,8 +37,8 @@ pub struct PreparedBlock {
     baker_id:                     Option<i64>,
     /// Total amount of CCD in existence at the time of this block.
     total_amount:                 i64,
-    /// Total staked CCD at the time of this block.
-    total_staked:                 i64,
+    /// All pools total staked CCD at the time of this block.
```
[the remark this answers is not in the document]
squirer: reverted to `total_staked` but kept updated comment as:

    /// Total staked CCD including delegation at the time of this block.
    total_staked:                 i64,

backend/src/indexer/block_preprocessor.rs
squirer: I think for now, lets see first how the initial release goes for this change - and if we need to optimise somewhere great to have the above two comments to help guide that, good to know we can parallelise and equally good to know that there is a limit to what the node can support

backend/src/indexer/block.rs
```diff
@@ -13,10 +12,18 @@ use special_transaction_outcomes::{
...
+
 pub mod block_item;
 pub mod protocol_update_migration;
 pub mod special_transaction_outcomes;
 
+#[derive(Clone)]
+pub struct ValidatorStakingInformation {
+    pub stake_amount:             u64,
+    pub pool_total_staked_amount: u64,
+}
```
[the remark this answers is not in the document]
squirer: Done!

backend/src/indexer/block_preprocessor.rs
```diff
@@ -194,22 +196,27 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
                         ..
-                    } => *total_staked_capital,
+                    } => {
+                        compute_validator_staking_information(
+                            &mut client3,
+                            v2::BlockIdentifier::AbsoluteHeight(fbi.height),
+                        )
+                        .await?
+                    }
                 };
```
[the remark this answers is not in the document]
squirer: consolidated these into one call now, no branches needed you are completely right

backend/src/indexer/block/special_transaction_outcomes.rs
squirer: I have now removed all references to this now, so when we get those stake increased events we just log that they occurred (in debug mode) as the increase and decrease will be taken from the node

PR #769, opened 2025-07-16
feat: Overview and TokenHolder chart Part of CRE-400 ,  Part of CRE-401, Part of CRE-402 , Part of  CRE-403 , CRE-404, CRE-405, CRE-406, CRE-407, CRE-408

Pratap2018:
## Purpose

Implement plt dashboard overview section and Token holders chart

## Changes

Added graphql query `pltEventMetrics`

## Checklist

      hard-to-understand areas.


backend/src/graphql_api/plt_event_metrics.sql
```diff
@@ -0,0 +1,110 @@
...
+  WHERE paa.amount > 0
+),
+
+aggregated AS (
+  SELECT
+    bt.bucket_start,
+
+    -- Event counts
+    (
+      SELECT COUNT(*) FROM events_with_block e
```
DOBEN, line 54: I don't think that `COUNT(*)` will perform well once we have lot's of data in the database (and large bucket sizes e.g. 1 year). Would be good to test this metrics against a large dataset in the database and analyze its performance to ensure we don't need to re-write it in the future.

Pratap2018:
Hi @DOBEN  I did some refactoring related to https://github.com/Concordium/concordium-scan/pull/765#discussion_r2201013491 .

backend/src/graphql_api/plt_event_metrics.sql
```diff
@@ -0,0 +1,143 @@
...
+    e.event_type,
+    CASE 
+      WHEN e.event_type = 'Transfer' THEN
+        COALESCE(((e.token_event->'amount'->>'value')::numeric / POWER(10, COALESCE(pt.decimal, 0))), 0)
+      ELSE 0
+    END AS transfer_amount
+  FROM plt_events e
+  JOIN plt_tokens pt ON pt.index = e.token_index
+  JOIN transactions tx ON e.transaction_index = tx.index
+  JOIN blocks b ON tx.block_height = b.height
```
squirer, line 28: this file concerns me.  

Even this single query on its own is going to select and carry out joins on 3 other tables to bring back information - to get plt's with transfer amounts. Not to mention all of the aggregations that are happening below. 

I am wondering, if we need to normalise some of the tables in the DB so that you have more information available - or at minimum we will need indexes to help. 

Before this PR merges, we will need an assessment on real data at scale - so for example, working with SRE to create a snapshot of the mainnet DB, Creating PLT tokens for that and also creating enough events for a timespan that makes sense so that we can check these queries and what happens.

The other thing we need to check is - the costs for running these queries, any long running queries are concerning because they hog connections and they can cause timeouts

squirer (changes requested):
I am a bit concerned about the database queries for this one - added comments inline there

backend/src/graphql_api/plt_event_metrics.sql
DOBEN: Have we done some testing on large datasets?

backend/src/graphql_api/plt_event_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    #[graphql(name = "y_TokenModuleCount")]
+    y_token_module_count: Vec<i64>,
+
+    #[graphql(name = "y_TotalEventCount")]
+    y_total_event_count: Vec<i64>,
+}
+
+#[Object]
+impl QueryPltEventMetrics {
```
squirer, line 73: Please add documentation to this function, on how it works - key search criteria or anything that would be useful for another developer taking a look

backend/src/graphql_api/plt_event_metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+            total_event_count += event_count;
+
+            y_transfer_count.push(transfer_count);
+            y_transfer_volume.push(transfer_volume);
+            y_mint_count.push(mint_count);
+            y_mint_volume.push(mint_volume);
+            y_burn_count.push(burn_count);
+            y_burn_volume.push(burn_volume);
+            y_token_module_count.push(token_module_count);
+            y_total_event_count.push(event_count);
```
squirer, line 161: Can you please add some comments here

backend/src/migrations/m0040_alter_plt_events.sql
```diff
@@ -0,0 +1,89 @@
+ALTER TABLE plt_events
+ADD COLUMN IF NOT EXISTS event_timestamp TIMESTAMPTZ;
+
+ALTER TABLE plt_events
+ADD COLUMN IF NOT EXISTS amount_value NUMERIC;
+
+ALTER TABLE plt_events
+ADD COLUMN IF NOT EXISTS amount_decimals INT;
+
+CREATE INDEX  IF NOT EXISTS idx_plt_events_token_type_time
```
squirer, line 10: i think it would be good to add a comment above indexes to explain what the index is used for - which particular part of the application does it serve for example if its search queries for dashboards or something

squirer (changes requested):
added some further comments about places to add some additional comments that will help developers

rbaconcordium:
hi @Pratap2018 , just a suggestion for forward PRs, could the PR be raised smaller and more frequent, please? This is to tackle the review complexity and hopefully it leads to speedier review and easier to spot potential bugs and recommendations, etc.

cc @squirer  @DOBEN

Pratap2018:
Closing this PR due to alternate PR raised instead of this one  #782  #783 #784

Pratap2018:
Closing this PR due to alternate PR raised instead of this one https://github.com/Concordium/concordium-scan/pull/782 https://github.com/Concordium/concordium-scan/pull/783 https://github.com/Concordium/concordium-scan/pull/784

backend/src/graphql_api/plt_event_metrics.sql
```diff
@@ -0,0 +1,54 @@
+-- Inputs:
+-- $1::interval - e.g. '90 days'
+-- $2::interval - e.g. '3 days'
+-- $3::BIGINT   - optional token filter
+
+WITH pre_agg_events AS MATERIALIZED (
```
allanbrondum, line 6: Seems this is something that will never perform well. Is the indexer even the place to store/request "analytics" data like this. I'm not sure it is the right solution to put this in the indexer/ccdscan?

PR #774, opened 2025-07-21
Introduced batches to file m0037

rbaconcordium:
## Purpose
Introducing batches to the script and adding more logging to show the progress.
The overall running of this particular script would still be quite a while, expect possibly a few days due to the large tables involved. The log should print how many rows are processed at the time of running, indicating the process is still running
and how far it has gone through.

## Changes
Added batching to the update statements, with size of 10k records. 
Added logging to show progress of the batches.

## Checklist

- [X] The code compiles without warnings.
      hard-to-understand areas.

### DIFF
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Introduced batches in migrations m0037 and logging.
+
 ## [2.0.15] - 2025-07-16
 
 ### Changed
--- a/backend/src/migrations/m0037_update_transaction_type_add_tokenupdate.rs
+++ b/backend/src/migrations/m0037_update_transaction_type_add_tokenupdate.rs
@@ -2,8 +2,12 @@
 //!   `TokenGovernance` enum values being present. These are not part of the
 //!   intended devnet account_transaction_type enum set.
 
+use tracing::info;
+
 use super::SchemaVersion;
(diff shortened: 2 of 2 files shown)


squirer:
Please add a changelog entry in the unreleased section for now

backend/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+## [2.0.16] - 2025-07-21
```
squirer, line 7: you can remove this line because this line is only added when it has been tagged for a release - we usually just keep a list of items in the unreleased section and once its tagged for release then we specify the date for all this section similar to below.

rbaconcordium: done

backend/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Introduced batches in migrations m0037 and logging.
```
DOBEN, line 7: We had one time a quite long migration file in the past and a partial migration process was developed. Might be worth to check it out in case you were not aware of it. The partial migration was storing its process in the database, so that we could re-start the migration if it fails (or SRE needed to stop/re-start the migration for some reasons). Might be worth to check it it.

https://github.com/Concordium/concordium-scan/blob/main/backend/src/migrations.rs#L361

https://github.com/Concordium/concordium-scan/blob/main/backend/src/migrations/m0019_payday_stake_information.rs

PR #776, opened 2025-07-22
Feature/cor 1642 stakedamounts performancefix

squirer:
## Purpose

Investigate and fix staked amounts by caching and not computing for every block

## Changes

Added a cache for the total staked amount, now we process and compute the new staked values for all the validators - every X number of blocks, which is configurable by environment variable: STAKE_RECOMPUTE_INTERVAL_IN_BLOCKS (default set to 500)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/src/indexer/block_preprocessor.rs
+++ b/backend/src/indexer/block_preprocessor.rs
@@ -24,7 +24,7 @@ use prometheus_client::{
     metrics::{counter::Counter, family::Family, gauge::Gauge, histogram},
     registry::Registry,
 };
-use std::{env, sync::Arc};
+use std::sync::Arc;
 use tokio::{sync::RwLock, time::Instant, try_join};
 use tracing::{debug, error, info, warn};
 
@@ -33,26 +33,30 @@ use tracing::{debug, error, info, warn};
 /// preprocessors can run in parallel, this must be `Sync`.
 pub struct BlockPreProcessor {
     /// Genesis hash, used to ensure the nodes are on the expected network.
-    genesis_hash:                 sdk_types::hashes::BlockHash,
+    genesis_hash: sdk_types::hashes::BlockHash,
     /// Metric counting the total number of connections ever established to a
     /// node.
     established_node_connections: Family<NodeMetricLabels, Counter>,
     /// Metric counting the total number of failed attempts to preprocess
     /// blocks.
-    preprocessing_failures:       Family<NodeMetricLabels, Counter>,
+    preprocessing_failures: Family<NodeMetricLabels, Counter>,
     /// Metric tracking the number of blocks currently being preprocessed.
-    blocks_being_preprocessed:    Family<NodeMetricLabels, Gauge>,
+    blocks_being_preprocessed: Family<NodeMetricLabels, Gauge>,
     /// Histogram collecting the time it takes for fetching all the block data
     /// from the node.
-    node_response_time:           Family<NodeMetricLabels, histogram::Histogram>,
+    node_response_time: Family<NodeMetricLabels, histogram::Histogram>,
     /// Max number of acceptable successive failures before shutting down the
     /// service.
-    max_successive_failures:      u64,
+    max_successive_failures: u64,
+    /// How often to recompute staked amounts for validators with the node.
+    /// Denotes the number of blocks to allow pass before updating stakes.
+    stake_recompute_interval_in_blocks: u64,
 }
 impl BlockPreProcessor {
     pub fn new(
         genesis_hash: sdk_types::hashes::BlockHash,
         max_successive_failures: u64,
+        stake_recompute_interval_in_blocks: u64,
         registry: &mut Registry,
     ) -> Self {
         let established_node_connections = Family::default();
@@ -90,6 +94,7 @@ impl BlockPreProcessor {
             blocks_being_preprocessed,
             node_response_time,
             max_successive_failures,
+            stake_recompute_interval_in_blocks,
         }
     }
 }
@@ -118,12 +123,6 @@ struct TotalStakedAmountsCache {
 /// A read write lock for the total staked amounts cache
 static TOTAL_STAKED_AMOUNTS_CACHE: Lazy<Arc<RwLock<Option<TotalStakedAmountsCache>>>> =
     Lazy::new(|| Arc::new(RwLock::new(None)));
-
-/// How often to recompute the staked amounts for validators, the default
-/// interval to recompute staked amounts is every X blocks.
-pub static STAKE_RECOMPUTE_INTERVAL_IN_BLOCKS: Lazy<u64> = Lazy::new(|| {
-    env::var("STAKE_RECOMPUTE_INTERVAL").ok().and_then(|s| s.parse::<u64>().ok()).unwrap_or(500)
-});
 
 #[tonic::async_trait]
 impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
@@ -218,6 +217,7 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
                 let computed_validator_staking_details = compute_validator_staking_information(
                     &mut client3,
                     v2::BlockIdentifier::AbsoluteHeight(fbi.height),
+                    self.stake_recompute_interval_in_blocks,
                 )
                 .await?;
                 Ok((tokenomics_info, computed_validator_staking_details))
@@ -318,6 +318,7 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
 pub async fn compute_validator_staking_information(
     client: &mut v2::Client,
     block_height: v2::BlockIdentifier,
+    stake_recompute_interval_in_blocks: u64,
 ) -> v2::QueryResult<(Amount, ValidatorStakingInformation)> {
     let current_block_height: u64 = get_block_height_from_absolute(block_height);
     let cache_read = TOTAL_STAKED_AMOUNTS_CACHE.read().await;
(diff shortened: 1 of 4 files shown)


rbaconcordium (commented):
seems ok, but will let the others review and approve

backend/CHANGELOG.md
```diff
@@ -4,9 +4,12 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+
```
DOBEN, line 7:
```suggestion
```

backend/Cargo.toml
```diff
@@ -14,14 +14,15 @@ anyhow = "1"
 async-graphql = {version = "7.0", features = ["chrono", "tracing"] }
 async-graphql-axum = "7.0"
 axum = "0.8"
-ciborium = "0.2.2"
+ciborium = "0.2.2"
```
DOBEN, line 17:
```suggestion
ciborium = "0.2.2"
```

backend/Cargo.toml
```diff
@@ -14,14 +14,15 @@ anyhow = "1"
...
-ciborium = "0.2.2"
+ciborium = "0.2.2"  
 chrono = { version = "0.4", features = ["serde"] }
 clap = { version = "4.5", features = ["derive", "env", "cargo"] }
 concordium-rust-sdk = { path = "./concordium-rust-sdk" }
 derive_more = { version = "1.0.0", features = ["display", "from", "from_str", "into"] }
 dotenvy = "0.15"
 futures = "0.3"
 hex = "0.4"
+once_cell = "1.19"
```
DOBEN, line 25: is there a reason we didn't use the newest version?

backend/src/indexer/block_preprocessor.rs
```diff
@@ -373,15 +437,41 @@ pub async fn compute_validator_staking_information(
...
+/// should always work with Absolute height for this area of the code.
+fn get_block_height_from_absolute(block_height: BlockIdentifier) -> u64 {
+    match block_height {
+        v2::BlockIdentifier::AbsoluteHeight(h) => h.height,
+        _ => {
+            debug!("Non-absolute block identifier; skipping cache.");
+            panic!("Expected absolute block height in order to work with caching effectively");
+        }
+    }
+}
```
DOBEN, line 473: If you create a function then the intention is that the function likely will be re-used at some point in the future and the function at is should be generic. As such the comment should not highlight its specific use at one specific code base point. If you only want the code to be used at that specific context then I would move this function + comment into the place where it is used (rather than creating a separate re-usable function).

backend/src/indexer/block_preprocessor.rs
```diff
@@ -106,6 +108,23 @@ impl NodeMetricLabels {
...
+
+/// A read write lock for the total staked amounts cache
+static TOTAL_STAKED_AMOUNTS_CACHE: Lazy<Arc<RwLock<Option<TotalStakedAmountsCache>>>> =
+    Lazy::new(|| Arc::new(RwLock::new(None)));
+
+/// How often to recompute the staked amounts for validators, the default
+/// interval to recompute staked amounts is every X blocks.
+pub static STAKE_RECOMPUTE_INTERVAL_IN_BLOCKS: Lazy<u64> = Lazy::new(|| {
+    env::var("STAKE_RECOMPUTE_INTERVAL").ok().and_then(|s| s.parse::<u64>().ok()).unwrap_or(500)
```
DOBEN, line 125: If you want a new environmental variable, it should be part of the cli so it is displayed under the `--help` menu to the user as a configurable option. Hardcording `500` as default value here is impossible for users to discover/see or be aware of.

https://github.com/Concordium/concordium-scan/blob/main/backend/src/bin/ccdscan-indexer.rs#L28

backend/src/indexer/block_preprocessor.rs
```diff
@@ -281,49 +300,105 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
+            let no_updates_required_for_validators = ValidatorStakingInformation {
+                ids: vec![],
+                staked_amounts: vec![],
+                pool_total_staked_amounts: vec![],
+            };
+            debug!(
+                "Returning cache total stake for height: {}, total staked: {}",
+                current_block_height, cache.total_staked
+            );
+            computed_stake_result = Ok((cache.total_staked, no_updates_required_for_validators));
```
DOBEN, line 346: if we return/leave the `compute_validator_staking_information` function here, then we don't need to specify/mark `should_update_cache` as true in all other arms of this conditional statement. In general I find it a confusing flow to have a default ` let mut should_update_cache = false;` specified at the top and then except for one arm to update its value to `true`. Would be good to get rid-off this `should_update_cache` variable.

backend/src/indexer/block_preprocessor.rs
```diff
@@ -281,49 +300,105 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
     }
 }
 
-/// compute validator staking information such as the validators stake and the
-/// total staked in their pool by getting pool info. pool info did not exist
-/// prior to protocol version 4, in this case the account info is used to get
-/// the stake for the validator and the total pool for the validator prior to
-/// protocol version 4 will be set to their account stake.
+/// Responsible for computing the latest validator staking information which
```
DOBEN, line 303: Do we have an idea how this PR affects the front-end or displays to the user? Will it cause confusion if some data is already new while other is outdated? E.g. reward/staking events are immediately visible in a transaction overview but if people click on their validator overview (some of the reward/staking events are not accounted for yet - only after 500 blocks). Would be good to involve QA (some with user flow experiences) to ensure this changes is acceptable base on the front-end experience.

If I understand correctly, the staked information will only be updated every say 500 blocks now. That means our diagrams will show a straight line on the newest 500 blocks (no new value) for example.

DOBEN (commented):
How much improvement do we get with the cache if we run the indexer locally (when everything runs on the same machine) and on a mainnet database snapshot (hope the setup as similar to how it is run it in production). Without that data it is hard to evaluate if adding complexity in the code base is worth it also I believe the changes improve the performance.

squirer:
> How much improvement do we get with the cache if we run the indexer locally (when everything runs on the same machine) and on a mainnet database snapshot (hope the setup is similar to how it is run it in production). Without that data it is hard to evaluate if adding complexity in the code base is worth in general. Looking at the code I do believe the changes improve the performance so.

With the mainnet SRE snapshot - the block processing time is much slower than mainnet, but comparible to that it dropped from averaging around 1000ms -> averaging around 800ms - so i think there is a good enough improvement there and we should get better feedback when we go to stagenet

DOBEN:
> With the mainnet SRE snapshot - the block processing time is much slower than mainnet, but comparible to that it dropped from averaging around 1000ms -> averaging around 800ms - so i think there is a good enough improvement there and we should get better feedback when we go to stagenet

I guess that data is collected with the default 500 value for `STAKE_RECOMPUTE_INTERVAL_IN_BLOCKS`.

backend/Cargo.toml
squirer: my bad, this was from a tutorial i was reading on once cell, I will update and recheck

backend/src/indexer/block_preprocessor.rs
squirer: Yes so here I was going for a tradeoff with querying the node for validator stakes and reflecting stakes as accurately as possible. My reasoning was 500 blocks will be processed quite quickly (probably less than 10mins) - so the view the user would get is stakes changing over a range of 10min intervals. 

I understand that this may have some UI repercussions and we may see slightly outdated information during this window where it has not updated, but with the window being short (like 10mins) I don't see it as a major issue. 

I agree lets check with QA on this one.

backend/src/indexer/block_preprocessor.rs
```diff
@@ -106,6 +108,23 @@ impl NodeMetricLabels {
...
+
+/// A read write lock for the total staked amounts cache
+static TOTAL_STAKED_AMOUNTS_CACHE: Lazy<Arc<RwLock<Option<TotalStakedAmountsCache>>>> =
+    Lazy::new(|| Arc::new(RwLock::new(None)));
+
+/// How often to recompute the staked amounts for validators, the default
+/// interval to recompute staked amounts is every X blocks.
+pub static STAKE_RECOMPUTE_INTERVAL_IN_BLOCKS: Lazy<u64> = Lazy::new(|| {
+    env::var("STAKE_RECOMPUTE_INTERVAL").ok().and_then(|s| s.parse::<u64>().ok()).unwrap_or(500)
+});
```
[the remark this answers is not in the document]
squirer: fixed now

backend/src/indexer/block_preprocessor.rs
```diff
@@ -373,15 +438,38 @@ pub async fn compute_validator_staking_information(
...
 }
 
+/// gets the block height as a u64 where the AbsoluteHeight identifier is
+/// passed.
+fn get_block_height_from_absolute(block_height: BlockIdentifier) -> u64 {
+    match block_height {
+        v2::BlockIdentifier::AbsoluteHeight(h) => h.height,
+        _ => {
+            debug!("Non-absolute block identifier; skipping cache.");
+            panic!("Expected absolute block height in order to work with caching effectively");
```
DOBEN, line 468: proper error reporting instead of panicing

backend/src/indexer/block_preprocessor.rs
```diff
@@ -281,49 +300,106 @@ impl concordium_rust_sdk::indexer::Indexer for BlockPreProcessor {
...
+/// amounts and their pool staked amounts. As it is intensive to go to the node
+/// through the client, the last computed total stake is stored in a cache. We
+/// only go to the retrieve the latest stakes if enough blocks have passed
+/// (defined by our env variable STAKE_RECOMPUTE_INTERVAL_IN_BLOCKS). It is
+/// important to note that in our cached scenario, we just return the last
+/// cached total stake amount - we do not return validatorStakingInformation
+/// because: if this is present when we reach the block processor it will
+/// perform database update queries for all the validator information provided,
+/// therefore it is crucial to return a struct that represents 'no updates at
+/// this time'
```
DOBEN, line 317: The `compute_validator_staking_information` is primarily executed in the `parallel` stage of the indexer. That is properly why you needed to add the `readWrite` lock because several processes tried to update the value. It seems a wrong design to now force the `parallel` stage to do sequential updates with a `ReadWrite` cache as we loose the parallel behavior. I wonder if we should move the `compute_validator_staking_information` to the save stage so that we can avoid the  `readWrite` lock.

limemloh:
I think we need to approach this differently, caching like this would make the data shown incorrect some of the time and our main objective for this application is to show the state of the chain.

PR #784, opened 2025-07-29
feat: implement plt event metrics

Pratap2018:
## Purpose

Implement PLT Event metrics 

#### Backend changes
- migration `m0040` is added.
- graphql query `plteventmetrics` is added

Explain analyze 
image

#### UI changes
image

## Checklist

      hard-to-understand areas.


backend/src/graphql_api/plt_event_metrics.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    #[graphql(name = "y_TotalEventCount")]
+    y_total_event_count: Vec<i64>,
+}
+
+// Implementation of the GraphQL query for PLT event metrics.
+// This query retrieves metrics related to PLT token events over a specified
+// period. It allows filtering by token ID and provides bucketed data for
+// analysis. The results include counts and volumes for transfers, mints, burns,
+// and token module creations.
```
squirer, line 83: i remember once before someone had mentioned to me about the Rust docs commenting style - i believe the style of the rust projects should be /// ahead of the methods and also for the inline field declarations in the structs - but may be worth double checking with someone. Not sure if the rust formatter deals well when these are the wrong way around.

@DOBEN  maybe it was you who mentioned this one to me before, or perhaps it was Emil G?

backend/src/graphql_api/plt_event_metrics.sql
```diff
@@ -0,0 +1,54 @@
+-- Inputs:
+-- $1::interval - e.g. '90 days'
+-- $2::interval - e.g. '3 days'
+-- $3::BIGINT   - optional token filter
+
+WITH pre_agg_events AS MATERIALIZED (
```
squirer, line 6: we need to be careful using Materialized - this will definitely be seen to have a performance improvement but i believe it introduces other costs. Did you explore using this query without Materialized?

Pratap2018: Without materialized.
image

With materialized
image

backend/src/graphql_api/plt_event_metrics.sql
```diff
@@ -0,0 +1,54 @@
...
+      ELSE 0
+    END AS amount
+  FROM plt_events
+  WHERE token_index IS NOT NULL
+    AND event_timestamp >= now() - $1::interval
+    AND event_timestamp <= now()
+  AND ($3::BIGINT IS NULL OR token_index = $3::BIGINT)
+),
+
+aggregated AS MATERIALIZED (
```
squirer, line 23: please remove MATERIALIZED (per my above comment) - i believe we should not materialize a table for any of these aggregated queries

backend/src/graphql_api/plt_event_metrics.sql
squirer: I think we should remove MATERIALIZED here - this will cache the table and store it in memory, which has a higher performance overhead. MATERIALIZED should be used when the data remains kind of static, which i believe is not the case for event metrics, it will be quite dynamic. So we should absorb the slight extra cost:

You can see performance considerations here: https://www.dbvis.com/thetable/view-vs-materialized-view-in-databases-differences-and-use-cases/#:~:text=Materialized%20Views%20offer%20a%20performance,doesn't%20require%20frequent%20updating.

squirer (changes requested):

backend/src/graphql_api/plt_event_metrics.sql
Pratap2018: Removed `MATERIALIZED`

squirer: i still see this Materialized here in the file source - perhaps all are not removed

backend/src/graphql_api/plt_event_metrics.rs
```diff
@@ -0,0 +1,205 @@
...
+    mint_count:      i64,
+    // Total volume of mints in the requested period.
+    mint_volume:     f64,
+
+    // Total number of burns in the requested period.
+    burn_count:  i64,
+    // Total volume of burns in the requested period.
+    burn_volume: f64,
+
+    // Total number of token modules created in the requested period.
```
DOBEN, line 37:
```suggestion
    // Total number of token module events in the requested period.
```

backend/src/graphql_api/plt_event_metrics.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    #[graphql(name = "y_TotalEventCount")]
+    y_total_event_count: Vec<i64>,
+}
+
+/// Implementation of the GraphQL query for PLT event metrics.
+/// This query retrieves metrics related to PLT token events over a specified
+/// period. It allows filtering by token ID and provides bucketed data for
+/// analysis. The results include counts and volumes for transfers, mints,
+/// burns, and token module creations.
```
DOBEN, line 83:
```suggestion
/// burns, and token creations.
```

backend/src/graphql_api/plt_event_metrics.sql
Pratap2018: Oh I missed that one

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -522,26 +525,38 @@ impl From<TokenUpdate> for PreparedTokenUpdate {
...
+                )
             }
             TokenEventDetails::Mint(e) => {
                 target = Some(e.target.address.to_string());
                 plt_amount_change =
                     BigDecimal::from(e.amount.value.parse::<f64>().unwrap_or(0.0) as u64);
-                TokenUpdateEventType::Mint
+                (
+                    TokenUpdateEventType::Mint,
+                    e.amount.value.parse::<u64>().unwrap_or(0),
```
DOBEN, line 547: I wonder if we should throw an error in all these cases where parsing fails as we should be able to parse the  value (or we have a bug in the code).

frontend/src/pages/protocol-token/index.vue
```diff
@@ -6,59 +6,81 @@
...
 							<p class="font-bold text-2xl mt-2">
 								{{
-									
 									numberFormatter(
-										pltEventMetricsDataRef?.pltEventMetrics.lastCumulativeTotalSupply
+										pltTokenDataRef.reduce(
+											(acc, coin) =>
+												acc +
+												(coin?.totalSupply ?? 0) /
+													Math.pow(10, Number(coin?.decimal)),
```
DOBEN, line 28: <PLTAmount> component or a common helper function so we don't re-write the same math formula.

backend/src/transaction_event/protocol_level_tokens.rs
Pratap2018: I think it is a good idea to throw error.

backend/src/graphql_api/plt_event_metrics.rs
```diff
@@ -0,0 +1,205 @@
...
+    // Total volume of mints in the requested period.
+    mint_volume:     f64,
+
+    // Total number of burns in the requested period.
+    burn_count:  i64,
+    // Total volume of burns in the requested period.
+    burn_volume: f64,
+
+    // Total number of token modules created in the requested period.
+    token_module_count: i64,
```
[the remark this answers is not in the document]
Pratap2018: This is token_module_type event count . (eg total number of event occured combining (pause , unpause, addallow , deny etc))

PR #789, opened 2025-08-11
feat: PLT (per plt) Token  Page Holder table dynamic (CRE-450)

Pratap2018:
## Purpose

Per Plt page Holder table dynamic

## Changes

- `idx_token_amount_desc` Index is added to  make the top Account amount holder query fast.
- Query `pltAccountsByTokenId` is added.
image

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/src/graphql_api/plt.rs
+++ b/backend/src/graphql_api/plt.rs
@@ -489,14 +489,69 @@ impl QueryPLTAccountAmount {
             .map_err(|e| ApiError::InternalServerError(InternalError::InternalError(e.to_string())))
     }
 
-    async fn plt_accounts_by_token_id(
+    async fn plt_accounts_by_token_id<'a>(
         &self,
-        ctx: &Context<'_>,
-        token_id: types::ID,
-    ) -> ApiResult<Vec<PLTAccountAmount>> {
+        ctx: &Context<'a>,
+        id: types::ID,
+        #[graphql(desc = "Returns the first _n_ elements from the list.")] first: Option<u64>,
+        #[graphql(desc = "Returns the elements in the list that come after the specified cursor.")]
+        after: Option<String>,
+        #[graphql(desc = "Returns the last _n_ elements from the list.")] last: Option<u64>,
+        #[graphql(desc = "Returns the elements in the list that come before the specified cursor.")]
+        before: Option<String>,
+    ) -> ApiResult<connection::Connection<String, PLTAccountAmount>> {
         let pool = get_pool(ctx)?;
-        let token_id: TokenId = token_id.to_string();
-        Ok(PLTAccountAmount::query_by_token_id(pool, token_id).await?)
+        let token_id = TokenId::from_str(id.as_ref()).map_err(|e| {
+            ApiError::InternalServerError(InternalError::InternalError(format!(
+                "Failed to parse token ID: {}",
+                e
+            )))
+        })?;
+        let config = get_config(ctx)?;
+        let query = ConnectionQuery::<DescendingI64>::new(
+            first,
+            after,
+            last,
+            before,
+            config.plt_token_events_collection_limit,
+        )?;
+
+        let mut row_stream = sqlx::query_as!(
+            PLTAccountAmount,
+            r#"SELECT 
+                account_index,
+                token_index,
+                amount,
+                decimal
+            FROM (
+                SELECT 
+                    account_index,
+                    token_index,
+                    amount,
+                    decimal
+                FROM plt_accounts
+                WHERE $2 < account_index AND account_index < $1 AND token_index = (
+                    SELECT index FROM plt_tokens WHERE token_id = $3 LIMIT 1
+                )
+                ORDER BY 
+                    CASE WHEN $4 THEN amount END ASC,
+                    CASE WHEN NOT $4 THEN amount END DESC
+                LIMIT $5
+            ) AS subquery
+            ORDER BY amount DESC
+           "#,
+            i64::from(query.from),
+            i64::from(query.to),
+            token_id.to_string(),
+            query.is_last,
+            query.limit,
+        )
+        .fetch(pool);
+        let mut connection = connection::Connection::new(false, false);
+        while let Some(tx) = row_stream.try_next().await? {
+            connection.edges.push(connection::Edge::new(tx.account_index.to_string(), tx));
+        }
+        Ok(connection)
     }
 
     async fn plt_unique_accounts(&self, ctx: &Context<'_>) -> ApiResult<i64> {
--- a/frontend/src/components/StableCoin/Holders.vue
+++ b/frontend/src/components/StableCoin/Holders.vue
@@ -1,6 +1,6 @@
 <template>
 	<div>
-		<div v-if="holderLoading" class="w-full h-36 text-center">
+		<div v-if="pltHolderLoading" class="w-full h-36 text-center">
 			<BWCubeLogoIcon class="w-10 h-10 animate-ping mt-8" />
 		</div>
 		<div v-else>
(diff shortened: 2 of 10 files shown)


backend/src/graphql_api/plt.rs
```diff
@@ -489,14 +489,69 @@ impl QueryPLTAccountAmount {
...
+        after: Option<String>,
+        #[graphql(desc = "Returns the last _n_ elements from the list.")] last: Option<u64>,
+        #[graphql(desc = "Returns the elements in the list that come before the specified cursor.")]
+        before: Option<String>,
+    ) -> ApiResult<connection::Connection<String, PLTAccountAmount>> {
         let pool = get_pool(ctx)?;
-        let token_id: TokenId = token_id.to_string();
-        Ok(PLTAccountAmount::query_by_token_id(pool, token_id).await?)
+        let token_id = TokenId::from_str(id.as_ref()).map_err(|e| {
+            ApiError::InternalServerError(InternalError::InternalError(format!(
```
DOBEN, line 505: I don't think that should be an `InternalServerError`, if someone queries this endpoint with an `id` that is not a `TokenId` (e.g. a string with some weird characters), then we should not raise it as an internal server error.

Internal server error messages will not be returned to the caller (as that could leak sensitive information), while in above case we would like the caller to inform that they made an mistake in calling the endpoint.

Internal server errors are there if Concordium needs to fix something (not the user/caller), e.g. Internal server errors are flagged in our internal monitoring and raise alarms and we don't want to flag a caller error there as it is nothing that we need to investigate/fix from our side (we cannot prevent outside parties calling the endpoint wrongly).

backend/src/graphql_api/plt.rs
```diff
@@ -489,14 +489,69 @@ impl QueryPLTAccountAmount {
...
+                e
+            )))
+        })?;
+        let config = get_config(ctx)?;
+        let query = ConnectionQuery::<DescendingI64>::new(
+            first,
+            after,
+            last,
+            before,
+            config.plt_token_events_collection_limit,
```
DOBEN, line 516: This query seems not related to `token_events`, so either rename that variable or use a separate variable.

backend/src/graphql_api/plt.rs
```diff
@@ -489,14 +489,69 @@ impl QueryPLTAccountAmount {
...
+                WHERE $2 < account_index AND account_index < $1 AND token_index = (
+                    SELECT index FROM plt_tokens WHERE token_id = $3 LIMIT 1
+                )
+                ORDER BY 
+                    CASE WHEN $4 THEN amount END ASC,
+                    CASE WHEN NOT $4 THEN amount END DESC
+                LIMIT $5
+            ) AS subquery
+            ORDER BY amount DESC
+           "#,
```
DOBEN, line 542:
```suggestion
            "SELECT 
                account_index,
                token_index,
                amount,
                decimal
            FROM (
                SELECT 
                    account_index,
                    token_index,
                    amount,
                    decimal
                FROM plt_accounts
                WHERE $2 < account_index AND account_index < $1 AND token_index = (
                    SELECT index FROM plt_tokens WHERE token_id = $3 LIMIT 1
                )
                ORDER BY 
                    CASE WHEN $4 THEN amount END ASC,
                    CASE WHEN NOT $4 THEN amount END DESC
                LIMIT $5
            ) AS subquery
            ORDER BY amount DESC
           ",
```

backend/src/graphql_api/plt.rs
```diff
@@ -489,14 +489,69 @@ impl QueryPLTAccountAmount {
...
+            token_id.to_string(),
+            query.is_last,
+            query.limit,
+        )
+        .fetch(pool);
+        let mut connection = connection::Connection::new(false, false);
+        while let Some(tx) = row_stream.try_next().await? {
+            connection.edges.push(connection::Edge::new(tx.account_index.to_string(), tx));
+        }
+        Ok(connection)
```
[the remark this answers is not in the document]
Pratap2018: Frontend does not rely on this pagination info, Should I add pagination?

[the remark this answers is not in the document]
Pratap2018: I have added pagination information , But i had to use nested cursor, Please review and please check if it became complex( unoptimised ).

frontend/src/utils/format.ts
```diff
@@ -259,3 +259,38 @@ export const calculateActualValue = (value: string, decimals: number) => {
 	// return the actual value as a string
 	return actualValue
 }
+// calculates percentage for BigInt values
+// value is a string representing a BigInt, total is a BigInt
+// returns a percentage as a number (0-100)
+export const calculatePercentageforBigInt = (value: string, total: bigint) => {
+	// value is bigint in string format
+	const valueBigInt = BigInt(value)
```
DOBEN, line 267: Suggestion would be to maybe just use the type directly as I anyway assume we have the value in bigint format rather than a string format. 

```suggestion
export const calculatePercentageforBigInt = (value: bigint, total: bigint) => {
```

frontend/src/utils/format.ts
```diff
@@ -259,3 +259,38 @@ export const calculateActualValue = (value: string, decimals: number) => {
...
+		nonSignificantDigits = valueStr.slice(totalLength - decimals)
+	}
+
+	const formattedInteger = Number(significantDigits)
+	const trimmedDecimals = nonSignificantDigits.slice(0, 2) // Limit to 2 decimal places
+	// as the values are in bigint, we can we represent the percentage by formatting otherwise the percentage will always integer
+	const percentageValue =
+		formattedInteger + (trimmedDecimals ? Number(`0.${trimmedDecimals}`) : 0)
+
+	return percentageValue.toString()
```
DOBEN, line 295: I am not clear about what we are doing here exactly and why there is so much type casting going on between `bigints`/`numbers`/`string` (that seems to lead to errors)?

The description of the function says that the return value is a number between 0 and 100.

I would expect:
```
const fractional_value = (valueBigInt * BigInt(100)) / total
```
or we need a better description what this function does (e.g. at which precision does it calculate the return value)

frontend/src/components/StableCoin/Holders.vue
```diff
@@ -58,38 +49,63 @@
...
 	TransactionFilterOption,
-	transactionFilterOptions,
 } from '~/types/stable-coin'
+import { ref, watch } from 'vue'
 
 // Define Props
 const props = defineProps<{
-	coinId?: string
+	coinId: string
+	totalSupply?: bigint
```
DOBEN, line 69: Can the `totalSupply` be undefined? Isn't the database always having a value as there is a `NOT NULL` constraint on the database column.

backend/src/graphql_api/plt.rs
```diff
@@ -537,14 +537,19 @@ impl QueryPltAccountAmount {
                   OR ($2 = $3 AND COALESCE(pa.amount, 0) = $2 AND pa.account_index < $5 AND \
              pa.account_index > $4)
               )
-            ORDER BY COALESCE(pa.amount, 0) DESC, pa.account_index DESC
+            ORDER BY 
+                CASE WHEN $7 THEN COALESCE(pa.amount, 0) END ASC,
+                CASE WHEN $7 THEN pa.account_index END ASC,
+                CASE WHEN NOT $7 THEN COALESCE(pa.amount, 0) END DESC,
+                CASE WHEN NOT $7 THEN pa.account_index END DESC
```
Vishwas1, line 544: I was wondering if this query can be simplified? As I understand , all we are doing here is, if $7 = true sort both these fields in asc order other wise sort them in desc order.
 
If so could we just mutiply pa.amount &  pa.account_index with 1 for asc order and -1 for dec, probably this would make the query less verbose? 

Does it make sense?

Vishwas1: Checked here https://github.com/Concordium/concordium-scan/blob/370b8fe0e58390e7b4ff8336556c931da66af206/backend/src/graphql_api/baker.rs#L1251 its is done they way you have done, so better not change anything.

backend/src/graphql_api/plt.rs
```diff
@@ -471,49 +471,146 @@ impl PLTToken {
...
+            BigDecimal::from(i64::from(query.to.outer)),   // $3 - amount lower bound
+            i64::from(query.from.inner),                   // $4 - account_index upper bound
+            i64::from(query.to.inner),                     // $5 - account_index lower bound
+            query.limit,
+            query.is_last,
+        )
+        .fetch(pool);
+        let mut connection = connection::Connection::new(false, false);
+        while let Some(tx) = row_stream.try_next().await? {
+            let amount_val = tx.amount.as_ref().and_then(|a| a.to_i64()).unwrap_or(0);
```
DOBEN, line 557: Instead of using a default value 0, suggestion would be to return an internal error so we catch that something is off. I would expect that the logic of the index stores an i64 into the database, so we should be able to parse that out.

backend/src/graphql_api/plt.rs
```diff
@@ -471,49 +471,146 @@ impl PLTToken {
...
+            let cursor: NestedCursor<DescendingI64, DescendingI64> = NestedCursor {
+                outer: amount_val.into(),
+                inner: tx.account_index.into(),
+            };
+            connection.edges.push(connection::Edge::new(cursor, tx));
+        }
+        if let (Some(page_min), Some(page_max)) =
```
DOBEN, line 564: The end cursor would need to include the `amount` and the `account_index`. It is the same logic as here:

https://github.com/Concordium/concordium-scan/blob/main/backend/src/graphql_api/passive_delegation.rs#L242

backend/src/graphql_api/plt.rs
```diff
@@ -471,49 +471,146 @@ impl PLTToken {
...
         let pool = get_pool(ctx)?;
-        let token_id: TokenId = token_id.to_string();
-        Ok(PLTAccountAmount::query_by_token_id(pool, token_id).await?)
+        let token_id = TokenId::from_str(token_id.as_ref())
+            .map_err(|e| ApiError::InvalidIdFormat(format!("Failed to parse token ID: {}", e)))?;
+        let config = get_config(ctx)?;
+        // Nested cursor: order by amount DESC, then account_index DESC for
+        // deterministic pagination This ensures accounts with same token
+        // amounts have a strict ordering
+        type Cursor = NestedCursor<DescendingI64, DescendingI64>;
```
DOBEN, line 512: https://github.com/Concordium/concordium-scan/blob/main/backend/src/graphql_api/passive_delegation.rs#L158

The best would be if you name the variables so there is no doubt which cursor is which:
```
        type AmountCursor = DescendingI64;
        type AccountIndexCursor = DescendingI64;
        type Cursor = NestedCursor<AmountCursor, AccountIndexCursor>;
```

frontend/src/utils/format.ts
```diff
@@ -259,3 +259,46 @@ export const calculateActualValue = (value: string, decimals: number) => {
...
+ *
+ * @param value - The numerator value as BigInt
+ * @param total - The denominator value as BigInt
+ * @returns Percentage as string(bigint can not represent decimals) with 2 decimal places (e.g., "12.34")
+ *
+ * @example
+ * calculatePercentageforBigInt(25n, 100n) // returns "25.00"
+ * calculatePercentageforBigInt(1n, 3n)    // returns "33.33"
+ * calculatePercentageforBigInt(1n, 7n)    // returns "14.28"
+ */
```
DOBEN, line 279: Great description.

DOBEN (approved):

PR #790, opened 2025-08-11
feat: plt-metrics (Analytics )  CRE-451

Pratap2018:
## Purpose

- Per Plt page analytics implemented 

## Changes

- Two new tables added for plt_metrics ( analytics ) `metrics_specific_plt_transfer` and `metrics_plt`
- Added query `PltTransferMetrics` to fetch metrics related to plt token transfer events.
- Added query `PltMetrics` to fetch plt metrics such as total event count, unique accounts, and transfer volume.

image

## Checklist

      hard-to-understand areas.


DOBEN (commented):
Can we break this big PR into smaller once. The description of the PR says it is about 2 tables.

 I am concern about performance as there is no performance analyses about the migration file the metrics in general.

frontend/src/components/molecules/ChartCards/StableCoinTokenTransfer.vue
```diff
@@ -56,37 +56,100 @@ const hoverColors = [
 
 // Define Props
 const props = defineProps<{
-	transferSummary?: StableCoinTokenTransferResponse
+	transferSummary?: PltTransferMetricsQueryResponse
 	isLoading?: boolean
```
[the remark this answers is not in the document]
Vishwas1: Also check if transferSummary and decimal should be optional or not. May be transferSummary need not to be optional.

frontend/src/types/stable-coin.ts
```diff
@@ -18,16 +18,17 @@ export type Transfers = {
...
 // types.ts
 export enum TransactionFilterOption {
+	Top10 = 10,
 	Top20 = 20,
 	Top50 = 50,
 	Top100 = 100,
 }
 
 // Optional: If you want labels for UI
 export const transactionFilterOptions = [
```
Vishwas1, line 28: I suggest , lets not change top to last as transactions comes based on recent time stamps. Here top signifies , recent transactions .

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -641,23 +640,58 @@ impl PreparedTokenUpdate {
...
+                    )
+                    .await?;
+
+                    self.update_metrics_plt_cumulative_transfer_amount(
+                        tx,
+                        transaction_index,
+                        self.plt_amount_change.clone()
+                            / BigDecimal::from(10u64.pow(self.amount_decimals as u32)),
+                    )
+                    .await?;
```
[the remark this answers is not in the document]
DOBEN: Personally, I would like to avoid adding too much `print/info/debug/log` statements through out the code without a clear guideline/agreement in the team what we want to log consistently the code base. Random `print/info/log` statements make it hard to see the actual code/review it. 

But I think we could discuss in a separate initiative/issue any logging concerns in the indexer and explore where we need more `logging`. One start would be to investigate the features of the `tracing` package. E.g. there are options to add tracing to the entry of a function (doesn't disrupt the main code with `print/info/log` statements) while it probably gives you the info of what is processed in each function, what are the values the function was called with:
https://github.com/Concordium/concordium-dapp-examples/blob/8e4ebfcb4f2da907312ebaf3ebad8330a333069c/sponsoredTransactionsAuction/backend/src/main.rs#L212

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -641,23 +653,58 @@ impl PreparedTokenUpdate {
...
+
         match self.event_type {
             TokenUpdateEventType::Mint => {
                 if let Some(ref target) = self.target {
+                    let account_exists = self.check_if_account_exists(tx, target).await?;
+
                     self.update_total_minted(tx).await?;
                     self.update_account_balance(tx, target, &self.plt_amount_change).await?;
+                    if !account_exists {
+                        self.update_metrics_plt_unique_account_count(tx, transaction_index).await?;
```
limemloh, line 666: We never decrement this counter again?

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -641,23 +653,58 @@ impl PreparedTokenUpdate {
...
+                    }
                 }
             }
             TokenUpdateEventType::Burn => {
                 if let Some(ref target) = self.target {
+                    let account_exists = self.check_if_account_exists(tx, target).await?;
```
limemloh, line 672: The account would need to hold PLT to burn, so `account_exists` would always be `true` right?
We can skip this query and skip updating the unique counter

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -641,23 +653,58 @@ impl PreparedTokenUpdate {
...
+                    let to_account_exists = self.check_if_account_exists(tx, to).await?;
+
                     self.update_account_balance(tx, from, &(-&self.plt_amount_change)).await?;
                     self.update_account_balance(tx, to, &self.plt_amount_change).await?;
+                    if !from_account_exists {
+                        self.update_metrics_plt_unique_account_count(tx, transaction_index).await?;
+                    }
+                    if !to_account_exists {
+                        self.update_metrics_plt_unique_account_count(tx, transaction_index).await?;
+                    }
```
limemloh, line 693: Sending queries to the database results in a lot of overhead, so we want to do as much as possible in each query.
I think the above can results in 8 queries, but I imagine this could have been 2 instead.

backend/src/transaction_event/protocol_level_tokens.rs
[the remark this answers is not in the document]
squirer: Yeah I agree we definitely need to settle on an approach - but for this PR to merge and if a release is created from it before there is the capacity to check logging standards and practices, I feel that the lack of having debug statements for now would mean that it would be exponentially harder to figure what has gone wrong for one of the events.

frontend/src/components/StableCoin/Analytics.vue
```diff
@@ -9,67 +9,74 @@
...
+							{ label: 'Top 10', value: TransactionFilterOption.Top10 },
+							{ label: 'Top 20', value: TransactionFilterOption.Top20 },
 						]"
-					/>
-					<StableCoinTokenDistributionByHolder
-						:token-transfer-data="dataPerStablecoin"
-						:is-loading="holderLoading"
-					/>
+					/> -->
+					<!-- <StableCoinTokenDistributionByHolder
```
squirer, line 33: please remove any commented out code here that is no longer needed

backend/src/migrations/m0040_alter_plt_events.sql
```diff
@@ -95,3 +95,268 @@ DROP INDEX IF EXISTS idx_temp_blocks_height;
 CREATE INDEX IF NOT EXISTS idx_token_amount_desc
 ON plt_accounts (token_index, amount DESC);
 
+
+
+-- Create table for transfer metrics PLT with cumulative data
+CREATE TABLE metrics_specific_plt_transfer (
```
allanbrondum, line 101: The name `metrics_plt_transfer` sounds a bit more straight forwards?

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -666,6 +713,229 @@ impl PreparedTokenUpdate {
...
+                unique_account_count
+            )
+            SELECT 
+                $1,
+                lm.prev_event_count + 1,
+                lm.prev_transfer_amount,
+                lm.prev_unique_count
+            FROM latest_metrics lm
+            ON CONFLICT (event_timestamp) DO UPDATE SET
+                cumulative_event_count = GREATEST(metrics_plt.cumulative_event_count, \
```
allanbrondum, line 907: Since we only have one writer, there cannot be a conflict. Can we just fail on conflict, or whatever the default behaviour is?

Pratap2018: Hi @allanbrondum  there could me multiple event in single transaction , those multiple event shares same single timestamp so there will be conflict

backend/src/transaction_event/protocol_level_tokens.rs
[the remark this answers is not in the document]
Pratap2018: I added some tracing::debug  , If you want me to remove i will remove it.

frontend/src/components/StableCoin/Analytics.vue
Pratap2018: I removed the comments

backend/src/transaction_event/protocol_level_tokens.rs
DOBEN: The balance of PLTs can become 0 again. PLTs are like smart contract tokens in that behavior. The transaction fee is paid in CCD (not a PLT token).

Your described behavior applies to CCD. It is in general hard to get a balance of 0 CCD on an account that had a positive balance before since you pay transaction fees in CCD. A small dust amount is usually left on an account since it is difficult to construct a transaction that transfers all your CCD away while having enough to pay for the tx fee.

backend/CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Database schema version: 40
 - TokenModuleRejectReason enum to proper serialization and deserialization of the `TokenModuleReject` event.
 - Query `PltUniqueAccounts` to fetch unique accounts holding PLT tokens.
 - Added query `PltEventMetrics` to fetch metrics related to plt token events.
+- Added query `PltTransferMetrics` to fetch metrics related to plt token transfer events.
+- Added query `PltMetrics` to fetch plt metrics such as total event count, unique accounts, and transfer volume.
```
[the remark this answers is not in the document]
Pratap2018: image

backend/src/graphql_api/plt_transfer_metrics.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Default)]
+
+pub(crate) struct QueryPltMetrics;
+
+#[derive(SimpleObject)]
+struct PltMetrics {
+    transaction_count: i64,
+    transfer_volume:   f64,
+    unique_accounts:   i64,
+}
```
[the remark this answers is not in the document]
Pratap2018: I completely agree with you, It doesn't make any sense as to display transfer_volume. will add docs.

backend/src/migrations/m0040_alter_plt_events.sql
```diff
@@ -95,3 +95,268 @@ DROP INDEX IF EXISTS idx_temp_blocks_height;
 CREATE INDEX IF NOT EXISTS idx_token_amount_desc
 ON plt_accounts (token_index, amount DESC);
 
+
```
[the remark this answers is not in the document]
Pratap2018: If We agree on re-indexing, I will remove the back-filling.

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -625,39 +639,217 @@ impl PreparedTokenUpdate {
...
+
+                    // METRICS UPDATE 1: Global PLT transfer aggregation
+                    //
+                    // Updates metrics_plt table with  transfer data:
+                    // - cumulative_transfer_amount: Running sum of normalized transfer volumes
+                    // - unique_account_count: Count of distinct accounts across all PLT tokens
+                    //
+                    // Query execution:
+                    // - CTE fetches previous metric values from latest row
+                    // - INSERT with ON CONFLICT (event_timestamp) for idempotent updates
```
DOBEN, line 789: I think a better explanation is that you insert max one row per block in the `metrics_plt` table.
Meaning, a new row in the `metrics_plt` table is inserted for the first PLT event in the block. For any further PLT event/update in the same block, the row is updated instead.

What do you mean with the `idempotent` exactly as the row will be changed/updated (you probably want to say that no new row will be generated in the `metrics_plt` table if several PLT events exist per block/event_timestamp/block_timestamp.

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -666,6 +858,98 @@ impl PreparedTokenUpdate {
...
+    // METRICS UPDATE: PLT event counter increment
+    //
+    // Increments cumulative_event_count in metrics_plt table for any PLT operation.
+    // Called for all event types: Transfer, Mint, Burn, TokenModule.
+    //
+    // Query execution:
+    // - CTE fetches current metric values from latest row
+    // - INSERT with ON CONFLICT (event_timestamp) handles concurrent events at same
+    //   timestamp
+    // - GREATEST() ensures monotonic increase under concurrent access
```
DOBEN, line 870: ?

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -666,6 +858,98 @@ impl PreparedTokenUpdate {
...
     }
 
+    // METRICS UPDATE: PLT event counter increment
+    //
+    // Increments cumulative_event_count in metrics_plt table for any PLT operation.
+    // Called for all event types: Transfer, Mint, Burn, TokenModule.
+    //
+    // Query execution:
+    // - CTE fetches current metric values from latest row
+    // - INSERT with ON CONFLICT (event_timestamp) handles concurrent events at same
```
DOBEN, line 868:
```suggestion
    // - INSERT with ON CONFLICT (event_timestamp) handles several events in the same block
```

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -714,4 +998,70 @@ impl PreparedTokenUpdate {
...
+        )
+        .fetch_optional(tx.as_mut())
+        .await?
+        .flatten()
+        .unwrap_or_else(|| BigDecimal::from(0));
+
+        Ok(final_balance)
+    }
+
+    async fn update_metrics_plt_unique_account_count_decrement(
```
DOBEN, line 1028: I would keep the two functions `update_metrics_plt_unique_account_count_decrement` and `update_metrics_plt_unique_account_count_increment` together.

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -666,6 +858,98 @@ impl PreparedTokenUpdate {
...
+            ",
+            event_timestamp
+        )
+        .execute(tx.as_mut())
+        .await?;
+
+        Ok(())
+    }
+
+    async fn update_metrics_plt_unique_account_count(
```
DOBEN, line 913:
```suggestion
    async fn update_metrics_plt_unique_account_count_increment(
```

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -613,51 +615,249 @@ impl PreparedTokenUpdate {
...
 
+        self.update_metrics_plt_cumulative_event_count(tx, slot_time).await?;
+
         match self.event_type {
             TokenUpdateEventType::Mint => {
                 if let Some(ref target) = self.target {
+                    let account_exists = self.check_if_account_exists(tx, target).await?;
                     self.update_total_minted(tx).await?;
                     self.update_account_balance(tx, target, &self.plt_amount_change).await?;
+                    if !account_exists {
```
DOBEN, line 647: There are still some edge case that might not be covered correctly. I think we should rethink this `unique_account_count` logic in the indexer as currently it seems to leave too much room for errors as we still finding cases that are not covered correctly. The current `account_exists` logic in the indexer seems quite buggy.

Case 1:
If it is allowed to mint 0 plts (needs some test if possible), then we could mint to a new address (that never held any plts yet) which would increase the `plt_unique_account_count` by 1 but it shouldn't as the new address holds 0 PLTs.

Case 2:
As well if an address has currently 0 PLTs but we mint something to it, it should increase the counter, which currently it doesn't.

This logic can be applied at all 3 events (mint/burn/transfer):
So what about using a simpler logic. If you have an account A, query all its tokens balance and sum it up before and after the event. If the `total_token_balance` was 0 and went to something greater than 0, we increment the counter, if the `total_token_balance` was something greater than 0 and went to 0, we decrement the counter. In other case we don't adjust the counter.

backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -613,51 +615,249 @@ impl PreparedTokenUpdate {
...
+                        self.update_metrics_plt_unique_account_count_delta(tx, slot_time, 1)
+                            .await?;
+                    }
                 }
             }
             TokenUpdateEventType::Burn => {
                 if let Some(ref target) = self.target {
                     self.update_total_burned(tx).await?;
-                    self.update_account_balance(tx, target, &(-&self.plt_amount_change)).await?;
+                    let final_balance = self
```
DOBEN, line 656: It needs to be the `final_balance` across all PLTs, isn't it? Currently, it is the `final_balance` for the given `token_id`. We only want to increment the counter if the account has no PLTs at all.

squirer (approved):

PR #792, opened 2025-08-11
add : migration m0041 to fix the token module reject reasons by properly serializing and deserializing the TokenModuleReject event using the TokenModuleRejectReason enum. CRE-479

Pratap2018:
## Purpose

- Proper serialization of `TokenModuleReject` is introduced and this Pr aims to clean up old TokenModuleReject and replace them with proper serialization in transaction table

## Changes

- m0041 migration is added.

## Checklist

      hard-to-understand areas.


backend/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Database schema version: 40
 - Added two new tables for plt metrics:
     - `metrics_specific_plt_transfer` to store plt transfer metrics (the data is being stored as cumulative increments by timestamp).
     - `metrics_plt` to store plt metrics (the data is being stored as cumulative increments by timestamp) .
+- Migration `m0041` to fix the token module reject reasons by properly serializing and deserializing the `TokenModuleReject` event using the `TokenModuleRejectReason` enum.
```
DOBEN, line 19: Didn't we say that we want to track the `Database schema version` in the changelog as well in line 7

```
Database schema version: 41
```

backend/src/migrations.rs
```diff
@@ -266,15 +266,17 @@ pub enum SchemaVersion {
     CreatePltTokenAndEventTables,
     #[display("0040: Alter PLT events add event_timestamp and index")]
     AlterPltEventsAddEventTimestampAndIndex,
+    #[display("0041: m0041_fix_token_module_reject_reasons")]
```
DOBEN, line 269: be consistent and write a proper message for display in line with the other displays

frontend/src/components/RejectionReason/Reasons/TokenModuleReject.vue
```diff
@@ -207,36 +139,21 @@
 				<span class="px-2">
 					Requested amount :
 					<b>
-						{{
-							// Todo: fix this when we reset the devnet DB
-
-							reason.details[reason.reasonType]?.requested_amount ??
-							reason.details.requested_amount
-						}}
+						{{ reason?.details?.requested_amount }}
```
DOBEN, line 142: Can these actually be undefined. I want to be sure that we don't display here `undefined` one day. Applies to the other areas as well.

frontend/src/components/RejectionReason/Reasons/TokenModuleReject.vue
```diff
@@ -35,18 +25,8 @@
...
-								reason.details[reason.reasonType]?.requiredBalance?.decimals[
-									'$serde_json::private::Number'
-								]
-								? reason.details[reason.reasonType]?.requiredBalance?.value /
-								  10 **
-										reason.details[reason.reasonType]?.requiredBalance
-											?.decimals['$serde_json::private::Number']
-								: Number(reason.details.required_balance.value) /
-								  10 ** Number(reason.details.required_balance.decimals)
+							Number(reason.details.required_balance.value) /
```
DOBEN, line 28: Bad idea to parse the value into `Number`.

We should have functions for that so we don't repeat the same math formula over and over again and I would like to get away from too much parsing of data types when you write your front-end code.

backend/src/migrations/m0041_fix_token_module_reject_reasons.sql
```diff
@@ -0,0 +1,169 @@
+-- Migration to fix previously stored token module reject reasons
```
DOBEN, line 1: This whole migration seems rather complex and not sure if we have all edge cases. To avoid these old to new type parsing (which is hard to review).

I wonder if we can generate much easier readable/reviewable code. My idea would be to identify all the `TokenUpdateTransactionFailed` tx and then parse the value from the below column through our cbor decoding function from the SDK and then overwrite what ever we need overwrite in the tables. Hope that would avoid to even bother what the `old/wrong values` were in the tables. 
https://github.com/Concordium/concordium-scan/blob/main/backend/src/migrations/m0001-initialize.sql#L224

backend/src/migrations.rs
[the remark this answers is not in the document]
allanbrondum: @DOBEN @limemloh I'm not even sure if it is possible to do a reindexing from a specific height only. Maybe it isn't? So the migration that clean up data may be needed to avoid a reindexing from genesis.

DOBEN (approved):

PR #794, opened 2025-08-14
Feature/cor 1522 db connection pool stats

squirer:
## Purpose

Expose database monitoring metrics for prometheus to be able to see the total, active and idle DB connections easily

## Changes

Added async task to the ccd scan api so that we can see easily the database monitoring metrics related to total, active and idle db connections.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -45,6 +45,7 @@ num-traits = "0.2.19"
 
 [dev-dependencies]
 mockito = "1.4"
+mockall = "0.13.1"
 
 # Recommended by SQLx to speed up incremental builds
 [profile.dev.package.sqlx-macros]
--- a/backend/src/bin/ccdscan-api.rs
+++ b/backend/src/bin/ccdscan-api.rs
@@ -1,10 +1,10 @@
-use anyhow::Context;
+use anyhow::{Context, Ok};
 use axum::{http::StatusCode, Json};
 use clap::Parser;
 use concordium_scan::{
     graphql_api::{self, node_status::NodeInfoReceiver},
-    migrations,
-    migrations::SchemaVersion,
+    migrations::{self, SchemaVersion},
+    monitoring::database_metrics::{DatabaseMetrics, RealDatabaseStatisticsProvider},
     rest_api, router,
 };
 use prometheus_client::{
@@ -17,7 +17,7 @@ use sqlx::postgres::{PgConnectOptions, PgPoolOptions};
 use std::{net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 use tokio::net::TcpListener;
 use tokio_util::sync::CancellationToken;
-use tracing::{error, info};
+use tracing::{debug, error, info};
 use tracing_subscriber::{layer::SubscriberExt as _, util::SubscriberInitExt as _};
 
 #[derive(Parser)]
@@ -65,6 +65,9 @@ struct Cli {
     /// backend
     #[arg(long, env = "CCDSCAN_API_NODE_COLLECTOR_PULL_FREQUENCY_SECS", default_value_t = 5)]
     node_collector_backend_pull_frequency_secs: u64,
+    /// Frequency in seconds of how often to update database metrics
+    #[arg(long, env = "CCDSCAN_API_UPDATE_DATABASE_METRICS_FREQUENCY_SECS", default_value_t = 1)]
+    database_metrics_update_frequency_secs: u64,
     /// Request timeout when awaiting response from the node collector backend
     /// in seconds.
     #[arg(long, env = "CCDSCAN_API_NODE_COLLECTOR_CLIENT_TIMEOUT_SECS", default_value_t = 30)]
@@ -156,6 +159,7 @@ async fn main() -> anyhow::Result<()> {
         )
         .await
         .context("Failed constructing database connection pool")?;
+
     // Ensure the database schema is compatible with supported schema version.
     migrations::ensure_compatible_schema_version(
         &pool,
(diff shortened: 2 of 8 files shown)


backend/src/monitoring/database_metrics.rs
```diff
@@ -0,0 +1,194 @@
...
+use prometheus_client::{metrics::gauge::Gauge, registry::Registry};
+use sqlx::PgPool;
+use std::sync::Arc;
+use tonic::async_trait;
+use tracing::{debug, info};
+
+/// Represent the metric name that will be registered with prometheus
+const ACTIVE_DATABASE_CONNECTIONS_METRIC_NAME: &str = "db_connections_active_count";
+const IDLE_DATABASE_CONNECTIONS_METRIC_NAME: &str = "db_connections_idle_count";
+const TOTAL_DATABASE_CONNECTIONS_METRIC_NAME: &str = "db_connections_total_count";
```
allanbrondum, line 11: I think we just need the pool size and how many are currently used? Then the difference is idle connection (in the pool, not the database)

backend/src/monitoring/database_metrics.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        // Gte the encoded metrics from the prometheus registry
+        let mut encoded = String::new();
+        prometheus_client::encoding::text::encode(&mut encoded, &registry).unwrap();
+
+        // assert that the encoded contains the correct metrics now too
+        assert!(encoded.contains("db_connections_active_count 5"));
+        assert!(encoded.contains("db_connections_idle_count 3"));
+        assert!(encoded.contains("db_connections_total_count 8"));
+    }
```
allanbrondum, line 193: Seems like we are mostly testing the prometheus exporter and not our own code? I think we may not need a test for this code?

backend/src/monitoring/database_metrics.rs
```diff
@@ -0,0 +1,194 @@
...
+        )
+        .fetch_one(&*pool)
+        .await?;
+
+        Ok((
+            database_connection_statistics.active,
+            database_connection_statistics.idle,
+            database_connection_statistics.total,
+        ))
+    }
```
td202, line 74: I'm a bit confused by this.

First, we're getting metrics from the database, rather than say using `pool.size()` and `pool.num_idle()` to see the number of connections our connection pool has, which is what I understood as the intention from the ticket.

Second, I'm not even sure if this gets quite the stats we want. From looking at `pg_state_activity`, this will give a view of the processes, which will include internal server processes.

And if we want the stats from the database rather than the API backend, then maybe it would be better to use an existing tool [like this](https://github.com/prometheus-community/postgres_exporter) rather than add this functionality here.

allanbrondum: I'll preparing to describe some task for improving our database monitoring, postgres_exporter will be one of the suggestions. So I agree, the pg_stats_activity metrics are relevant, but I don't think it should be ccdscan that exposes them, rather the database itself (via the postgres_exporter) should.

backend/src/bin/ccdscan-api.rs
```diff
@@ -65,6 +65,9 @@ struct Cli {
     /// backend
     #[arg(long, env = "CCDSCAN_API_NODE_COLLECTOR_PULL_FREQUENCY_SECS", default_value_t = 5)]
     node_collector_backend_pull_frequency_secs: u64,
+    /// Frequency in seconds of how often to update database metrics
+    #[arg(long, env = "CCDSCAN_API_UPDATE_DATABASE_METRICS_FREQUENCY_SECS", default_value_t = 1)]
```
allanbrondum, line 69: If we do database queries for metrics, every 1 sec is likely a bit too agressive. I assume we only collect metrics every 30 seconds or in that range. So we don't need to query that often. Querying system tables (like pg_stats_activity) can also consume resources, at least in some circumstances, so we should not do it more often than actually needed.

backend/src/monitoring/database_metrics.rs
squirer: @allanbrondum I agree - i can switch to using PgPool to take the size and then to calculate the active connections i just subtract the idle.

I think i should also register the max connections for the DB with prometheus - because the size() is an initial allocated number of connections, which grows to the max. So i think its useful to have metrics that display in prometheus and Grafana for example:

Max: 10, Active: 1, Idle: 9

backend/Cargo.toml
```diff
@@ -45,6 +45,7 @@ num-traits = "0.2.19"
 
 [dev-dependencies]
 mockito = "1.4"
+mockall = "0.13.1"
```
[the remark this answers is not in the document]
squirer: just a question here on this one, if we implement the stub ourself then we lose the ability to return different values under different conditions (if we have a bunch of unit tests for example) ? 

I went the mock route because i could easily just say that it returns my object with minimal setup required. Let's chat further about this one if needed

backend/src/bin/ccdscan-api.rs
allanbrondum: Even if you get an alert 30 seconds earlier, your reaction time is at least 5-10 minutes (very best case) to change connection settings in the database, so I don't think it will matter. But anyways, using `PgPool` you don't need the timer, you can just read the values whenever the metrics are scraped.

backend/src/monitoring/database_metrics.rs
squirer: hmmm im not sure i fully agree with this one - because the test (although trivial) on 181 - 183 will confirm that the object returned from our update function? So i think its still useful to confirm in case a developer did make a change in future and they accidentally swapped some parameters, the prometheus metric would be set for the wrong one for example causing an issue...

PR #801, opened 2025-08-21
Ar/cor 1687 investigate long running queries ccd scan mainnet 3

allanbrondum:
## Purpose

Improve the performance of the `query` searches for `transactions` and `blocks`.

## Changes

Use `plan_cache_mode = force_custom_plan` for the searches.
Change hash based searches to only run if there are at least 6 characters in the query
Also fixes paging but in `transactions` that did not take search query into account.

## Checklist

      hard-to-understand areas.


limemloh (approved):
Great work

PR #806, opened 2025-08-25
feat: Add PLT configuration information to dedicated PLT page CRE-473

Pratap2018:
## Purpose

Add PLT configuration information to dedicated PLT page 

## Changes

image

- Query `PltTokens` now returns `token_creation_details`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/src/graphql_api/plt.rs
+++ b/backend/src/graphql_api/plt.rs
@@ -550,6 +550,11 @@ impl PltToken {
                 return Ok(token_creation_details);
             }
         }
+        // Log an internal error as this should never happen if the database is correct
+        tracing::error!(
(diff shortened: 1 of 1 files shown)


backend/src/graphql_api/plt.rs
```diff
@@ -520,6 +521,38 @@ impl PltToken {
...
+
+        // Find the TokenCreationDetails event
+        for event in events {
+            if let crate::transaction_event::Event::TokenCreationDetails(token_creation_details) =
+                event
+            {
+                return Ok(token_creation_details);
+            }
+        }
+
```
DOBEN, line 553: I would log an internal error is it should never happen that we cannot find the `tokenCreationDetail` event if the data in the database is correct.

DOBEN (approved):

PR #807, opened 2025-08-26
fix: CRE-509

Pratap2018:
## Purpose

```
Failed processing block item with hash 000fcfa0378ae139bae6cdb9f5f51b13da63323388e26e38a21e717495b11464 for block height 1721897

Caused by:
    0: Failed processing block item event from Account transaction
    1: Failed processing block item event with token update events
    2: error returned from database: null value in column "account_index" of relation "plt_accounts" violates not-null constraint
    3: null value in column "account_index" of relation "plt_accounts" violates not-null constraint
```

https://linear.app/concordium/issue/CRE-509/plt-transfers-are-not-displayed-on-ccdscan
    
## Changes

while updating plt account balance we are now using canonical_address to find the account index (as address could  be alias)

## Checklist

      hard-to-understand areas.


backend/src/transaction_event/protocol_level_tokens.rs
```diff
@@ -644,20 +649,73 @@ impl PreparedTokenUpdate {
...
+                    let account_address =
+                        concordium_rust_sdk::base::contracts_common::AccountAddress::from_str(
+                            target.as_str(),
+                        )
+                        .map_err(|_| {
+                            anyhow::anyhow!(
+                                "Failed to convert account address to canonical address: {}",
+                                target
+                            )
+                        })?;
```
limemloh, line 661: It would be better to do it as part of the preprocessing and not when this is stored to the database.
Also you should not need to convert it into a string and then parse the string again.

Pratap2018: We will get back here, I was trying to do that , but I was getting stuck due to some errors and seems like a lot of change , I will make a card for this in the backlog.

PR #814, opened 2025-09-03
fix: CRE-531

Pratap2018:
## Purpose

When transferring PLT tokens with a memo, the memo text is not visible on CCDScan. This behavior is inconsistent with CCD transfers, where the memo is displayed correctly.

## Changes

- A new `PltTransferMemo` component is added

image

## Checklist

      hard-to-understand areas.

### DIFF
--- a/frontend/src/components/TransactionEventList/Events/PltTransferMemo.vue
+++ b/frontend/src/components/TransactionEventList/Events/PltTransferMemo.vue
@@ -1,0 +1,23 @@
+<template>
+	<div
+		v-if="memo?.__typename === 'Memo'"
+		class="bg-gray rounded flex items-center gap-2"
+	>
+		<span>Memo:</span>
+		<span class="break-all whitespace-pre-line text-xs">{{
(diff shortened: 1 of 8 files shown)


frontend/src/components/TransactionEventList/Events/PltTransferMemo.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+	<div
+		v-if="memo?.__typename === 'Memo'"
+		class="bg-gray rounded flex items-center gap-2"
+	>
+		<span>Memo:</span>
+		<span class="break-all whitespace-pre-line text-xs">{{
+			memo?.bytes ? Buffer.from(memo.bytes, 'hex').toString('utf-8') : ''
```
DOBEN, line 8: A `memo` can be either cbor encoded data or text data. You are only handling the text data here. The backend already decoded the `memo` for `RegisterData` tx and `CCDTransfer` tx. Please use that decoding:

https://github.com/Concordium/concordium-scan/blob/fa0b4f2681418b05e22cf028772036462d55230e/backend/src/decoded_text.rs#L4

DOBEN (approved):

PR #831, opened 2025-09-22
Adjust Ordering of new tokens on PLT Dashboard CRE-572

Pratap2018:
## Purpose

Adjust Ordering of new tokens on PLT Dashboard

## Changes

Plt tokens on the overview page should be shorted by supply.
Added a trigger to utilize unused total_holder column in the plt_token table,
optimised the uniquer holder query

## Checklist

      hard-to-understand areas.

### DIFF
--- a/frontend/CHANGELOG.md
+++ b/frontend/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased 
 
+### Fixed 
+
+- Fixed plt dashboard now displays total `10` plt sorted by current supply.
+
 ## [1.7.26] - 2025-09-16
 
 ### Fixed


squirer (changes requested):

backend/src/migrations/m0044-alter-plt-tokens-current-supply.sql
```diff
@@ -0,0 +1,74 @@
+ALTER TABLE plt_tokens
+ADD COLUMN normalized_current_supply NUMERIC
+GENERATED ALWAYS AS (
+    (COALESCE(total_minted, 0) - COALESCE(total_burned, 0)) 
+    / NULLIF(POWER(10::NUMERIC, decimal), 0)
```
limemloh: **Question:**
The expression `POWER(10::NUMERIC, decimal)` can never be 0, since `10^x = 0` is impossible (at least mathematically), so do we actually need `NULLIF(..)` here?

frontend/src/queries/usePltTokenQuery.ts
```diff
@@ -28,6 +28,7 @@ const PLT_TOKEN_QUERY = gql<PltTokenQueryResponse>`
 					asString
 				}
 				totalUniqueHolders
+				normalizedCurrentSupply
```
limemloh: Update the changelog, especially since it expects a new version of the backend to be released first

squirer (approved):

PR #832, opened 2025-09-25
feat: Show all  plts CRE-573

Pratap2018:
## Purpose

Show all available plt list to a new page.

## Changes

image

## Checklist

      hard-to-understand areas.

### DIFF
--- a/frontend/src/pages/protocol-token/list/all.vue
+++ b/frontend/src/pages/protocol-token/list/all.vue
@@ -174,8 +174,8 @@ const {
 	pageInfo,
 	loading: pltTokenLoading,
 } = usePltTokensPagedQuery({
-	first, 
-	last, 
+	first,
+	last,
 	after,
 	before,
 })


frontend/src/pages/protocol-token/list/all.vue
```diff
@@ -0,0 +1,274 @@
...
+		pltTokenData.value = newData
+	},
+	{ immediate: true }
+)
+
+const truncateAddress = (address: string, length: number = 12) => {
+	if (!address) return ''
+	if (address.length <= length) return address
+	return `${address.slice(0, length / 2)}...${address.slice(-length / 2)}`
+}
```
DOBEN, line 273: Please add doc comments to a function and move this function to a general place is it can/should likely be re-used (not strictly related to PLT tokens). 

Please explain the `length` parameter in the function.

frontend/src/pages/protocol-token/list/all.vue
```diff
@@ -0,0 +1,274 @@
...
+
+watch([after, before], () => {
+	if (!after.value && !before.value) {
+		currentPage.value = 1
+	} else if (after.value) {
+		currentPage.value = currentPage.value + 1
+	} else if (before.value) {
+		currentPage.value = Math.max(1, currentPage.value - 1)
+	}
+})
```
DOBEN, line 246: If we display the `token.index` then we don't need to calculated the `currentPage` and watch for these values.

frontend/src/pages/protocol-token/list/all.vue
```diff
@@ -0,0 +1,274 @@
...
+
+							<!-- Token Rows -->
+							<TableRow
+								v-for="(token, index) in pltTokenData"
+								:key="token.tokenId"
+								class="hover:bg-theme-background-secondary transition-colors duration-200"
+							>
+								<td class="py-4 px-4">
+									<span class="text-sm font-medium text-theme-text-secondary">
+										{{ (currentPage - 1) * pageSize + index + 1 }}
```
DOBEN, line 117: To avoid calculating/number the tokens as this value seems the same as the `token.index`.
```suggestion
										{{ token.index }}
```

PR #888, opened 2026-07-17
[COR-2459] Add support for Locks

Ivan-Mahda:
## Purpose

Add support for Protocol Level Token Locks so Concordium Scan can correctly process, index, and display lock-related activity introduced with P11.

This gives users a clear view of locked token activity directly from indexed chain data: lock creation, funding, sending, returning, cancellation, current lock status, balances, and history. It also prevents new lock-related transaction types and rejection reasons from being dropped or shown as unknown.

## Changes

<table>
    <tr>
        <td>
image
</td>
        <td>
image
</td>
    </tr>
    <tr>
        <td>
image
</td>
        <td>
image
</td>
    </tr>
</table>

- Added backend support for new lock-related transaction types, reject reasons, chain updates, and token transfer lock fields.
- Added transaction detail support for lock lifecycle events, so users can see when locks are created, funded, sent from, returned from, or canceled.
- Added indexed database state for locks, including current status, balances, related accounts, and historical lifecycle events.
- Added GraphQL APIs for querying individual locks and account-related locks from indexed data only.
- Added frontend lock display support, including lock links, lock-aware token transfer text, and readable labels for new lock events and rejection reasons.
- Added a lock details drawer showing lock status, creation and expiry information, balances, configuration, and lifecycle history.
- Added account drawer support for related locks, including account-specific locked balances and expired or canceled locks.
- Regenerated backend and frontend GraphQL schemas/types and updated checks around the new lock data paths.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -4,10 +4,6 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
-Database schema version: 50
-
-Database schema version: 51
-
 Database schema version: 52
 
 ### Added
--- a/backend/CHANGELOG.md
+++ b/backend/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+Database schema version: 50
+
+Database schema version: 51
+
+Database schema version: 52
+
+### Added
+
+- Added support for Locks/P11 `MetaUpdate` transaction types, reject reasons, and lock-aware PLT transfers.
+- Added lock event indexing, GraphQL lock details, account related locks, roles, and lock search.
+- Added support for `MaxLockDurationUpdate` chain update payloads.
+
 ## [2.0.29] - 2026-05-05
 
 ### Added
--- a/backend/src/graphql_api/lock.rs
+++ b/backend/src/graphql_api/lock.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeSet;
+
 use async_graphql::{connection, types, Context, Enum, Object};
 use bigdecimal::BigDecimal;
 use chrono::Utc;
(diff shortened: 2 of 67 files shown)


backend/src/migrations.rs
```diff
@@ -298,14 +298,20 @@ pub enum SchemaVersion {
...
+    #[display("0051: Add MaxLockDuration update transaction type")]
+    AddMaxLockDurationUpdateTransactionType,
+    #[display("0052: Add PLT lock state tables")]
+    AddPltLockStateTables,
 }
 impl SchemaVersion {
     /// The minimum supported database schema version for the API.
     /// Fails at startup if any breaking (destructive) database schema versions
     /// have been introduced since this version.
     pub const API_SUPPORTED_SCHEMA_VERSION: SchemaVersion = SchemaVersion::IndexPltHolderNonZero;
```
limemloh: The API will assume the new tables are available, so this needs to be bumped as well
```suggestion
    pub const API_SUPPORTED_SCHEMA_VERSION: SchemaVersion = SchemaVersion::AddPltLockStateTables;
```

backend/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+Database schema version: 50
+
+Database schema version: 51
+
```
limemloh: The newest schema version is enough here
```suggestion
```

backend/src/migrations/m0052_plt_locks.sql
```diff
@@ -0,0 +1,88 @@
...
+    created_at TIMESTAMPTZ,
+    expiry TIMESTAMPTZ,
+    canceled_transaction_index BIGINT REFERENCES transactions(index),
+    canceled_block_height BIGINT,
+    canceled_at TIMESTAMPTZ,
+    config JSONB,
+    raw_config TEXT,
+    metadata_name TEXT,
+    metadata_description TEXT
+);
```
limemloh, line 15: Some of these timestamps are redundant since they can be derived using block height and the slot time of that block, however I can imagine this simplifies some of the queries so it might still be worth it.

backend/src/migrations/m0052_plt_locks.sql
```diff
@@ -0,0 +1,88 @@
...
+    created_block_height BIGINT,
+    created_at TIMESTAMPTZ,
+    expiry TIMESTAMPTZ,
+    canceled_transaction_index BIGINT REFERENCES transactions(index),
+    canceled_block_height BIGINT,
+    canceled_at TIMESTAMPTZ,
+    config JSONB,
+    raw_config TEXT,
+    metadata_name TEXT,
+    metadata_description TEXT
```
limemloh, line 14: Should most of these be `NOT NULL`?

backend/src/graphql_api/lock.rs
```diff
@@ -0,0 +1,659 @@
...
+    pub canceled_at: Option<DateTime>,
+    pub config: Option<Json<LockCreateConfig>>,
+    pub raw_config: Option<String>,
+    pub metadata_name: Option<String>,
+    pub metadata_description: Option<String>,
+}
+
+impl Lock {
+    pub async fn query_by_id(pool: &PgPool, lock_id: &str) -> ApiResult<Option<Self>> {
+        let lock = sqlx::query_as::<_, Lock>(
```
limemloh: Please use compile-time checked queries as we have little testing. The compile-time checked queries catches a bunch of mistakes over time during migration and means we can do less manual testing.
More concretely you would use the macro version of `query_as` which is [`sqlx::query_as!(Lock, r#"...")`](https://docs.rs/sqlx/latest/sqlx/macro.query_as.html)

limemloh (commented):
Looks good to me.

backend/src/migrations/m0052_plt_locks.sql
Ivan-Mahda: `plt_locks` intentionally keeps most columns nullable because non-create events can create placeholder lock rows through `ensure_lock_exists`; the create event later fills creation/config fields. Cancellation fields also remain nullable for active/expired locks. The other lock tables already mark required columns NOT NULL, while event-specific fields remain nullable by event type.

PR #890, opened 2026-07-24
chore(ci): push ccdscan PR images to ECR

mt-concordium:
## Purpose

 - Add CI workflow to push to ECR to allow for image testing from the branch

### DIFF
--- a/.github/workflows/pr-ecr-backend.yml
+++ b/.github/workflows/pr-ecr-backend.yml
@@ -1,0 +1,53 @@
+name: PR Backend image to ECR
+
+# On every backend PR, build the indexer and api images and push them to ECR in
+# the concordium-builds account, tagged with the PR head commit SHA.
+# Release images continue to go to Docker Hub via release-backend.yml on tags.
+on:
+  pull_request:
+    paths:
+      - backend/**
+      - .github/workflows/pr-ecr-backend.yml
+
+jobs:
+  build-push:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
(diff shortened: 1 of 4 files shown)


Ivan-Mahda (approved):

ktconcordium (approved):

.github/workflows/pr-ecr-backend.yml
```diff
@@ -0,0 +1,53 @@
+name: PR Backend image to ECR
+
+# On every backend PR, build the indexer and api images and push them to ECR in
+# the concordium-builds account, tagged with the PR head commit SHA.
+# Release images continue to go to Docker Hub via release-backend.yml on tags.
+on:
+  pull_request:
```
andynelson: It's not clear here, when the expected docker build/push should happen - `pull_request` has multiple event types... I have a feeling this will run on all pull request events (opening, pushes, comments)

Would:
```
on:
  pull_request:
    type:
      - closed
```
combined with a conditional on `github.event.pull_request.merged = true` make more sense?

(alternatively we could run it on `push` to the specific branches we want to target (e.g. `main` & `release/p11` at the moment)

mt-concordium: good point, this would definitely run on every PR push

I'd be happy to set it to run on pushes on that match `release/*` branches so it works for future cases. But we can do it depending on the dev flow that we expect to see where we want to use ECR pushed images for.

mt-concordium: this is now updated to push images to ecr only on `main` and `release/p11` branch pushes

andynelson (approved):

